PR #4, opened 2016-08-19
fixed main_megamenu.less for compiling

mip-ro:
Errors because of auto-formatting of comments that broke the less compiling


Lausselloic:
Thanks for this contribution.

I will check why the boosted syntax checker let it passed thru

PR #13, opened 2016-10-13
Improve the screen-reader only classes

ffoodd:
I also wrote [a post about this](http://www.ffoodd.fr/cache-cache-css/).

It's supposed to be much more robust and future proof than the actual version, but it needs to be tested against various configurations (browsers and ATs).

I'd be pleased to discuss this.

### DIFF
--- a/scss/mixins/_screen-reader.scss
+++ b/scss/mixins/_screen-reader.scss
@@ -44,11 +44,11 @@
 @mixin sr-only-focusable {
   &:active,
   &:focus {
-    position: static;
     width: auto;
     height: auto;
     overflow: visible;
     white-space: normal;
+    clip: auto;
     clip-path: none;
   }
 }


ffoodd:
A bug has been found with iOS + VoiceOver and Android + TalkBack, [issue opened in Bootstrap](https://github.com/twbs/bootstrap/issues/20732).

My additions don't help.

ffoodd:
For your information, bugs have been filled in almost every mobile / tactile screen readers:

- [Narrator](https://microsoftaccessibility.uservoice.com/forums/307429-microsoft-accessibility-feedback/suggestions/16717318-focusable-elements-should-fire-focus-event-recei)
- [Chromium for android](https://bugs.chromium.org/p/chromium/issues/detail?id=657157)
- [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1000082)

It appears that screen readers on tactile devices move the accessibility focus without firing the JavaScript `focus` event nor applying the CSS `:focus` pseudo-class. This is the short version.

It's meant to prevent user from accidentally activate something (like text input that opens the keyboard and so on), but causes skip links to be useless — and maybe painfull.

However VoiceOver does restrict this behavior to form inputs: links and buttons do trigger focus as accessibility focus.

ffoodd:
Hi there,

I close this because it's been merged into Bootstrap and should be shipped in Beta 1:
https://github.com/twbs/bootstrap/pull/22154

If you find any issue with this, we shall report it on Bootstrap directly :)

PR #81, opened 2018-01-25
fix node_modules path in `scss/_functions.scss`

adammockor:
If you install `boosted: 4.0.0` as npm dependency, path `../node_modules/` in `scss/_functions.scss` do not exists in npm >= 3 environments.

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -1,7 +1,7 @@
 // Bootstrap functions
 //
 // Utility mixins and functions for evalutating source code across our variables, maps, and mixins.
-@import "../node_modules/sass-math-pow/sass/math-pow";
+@import "node_modules/sass-math-pow/sass/math-pow";
 // Ascending
 // Used to evaluate Sass maps like our grid breakpoints.
 @mixin _assert-ascending($map, $map-name) {


Lausselloic:
Hello,

Thanks for this PR. I put the relative path for command line build with ruby : `sass boosted.scss` if I remove it, that doesn't compile.

With or without the relative path, node-sass works.

So don't really know if there's a need to remove it? What did it improve if relative path is removed?

adammockor:
I am not sure how ruby sass compilation works and I got errors when I try to install `gem install` so I don't test it out.

But I am in situatation like this:
```json
// package.json
{
  name: "test",
  version: "0.0.1",
  dependencies: {
    "boosted": "~4.0.0"
  }
}
```
```scss
// src/style.scss
@import 'node_modules/boosted/scss/_functions';
```
and in ends up with error:
```
/src/style.scss
Module build failed: ModuleBuildError: Module build failed:
@import "../node_modules/sass-math-pow/sass/math-pow";
^
      File to import not found or unreadable: ../node_modules/sass-math-pow/sass/math-pow.
Parent style sheet: test/node_modules/boosted/scss/_functions.scss
      in test/node_modules/boosted/scss/_functions.scss (line 4, column 1)
```

Relative import does not work since npm install everything flat in `node_modules` so path `../node_modules/sass-math-pow/sass/math-pow` does not correspond to actual path on file system when modules are installed as dependecies of project. 

So solution could be to add `node_modules` to `:load_paths:` for ruby sass with http://sass-lang.com/documentation/file.SASS_REFERENCE.html?

Lausselloic:
Oh yes good catch-up :-) I will merge it

Lausselloic:
hello @adammockor this PR is now available in Boosted v4.0.1

adammockor:
I can confirm it works as expected now.

Issue #95, opened 2019-01-16
[RTL] — Scroll up link icon is reversed


ffoodd:
RTLCSS (which is good) automagically reverses angles in the `rotate()` function. Which makes `.o-scroll-up-icon` descending :)

I tricked it by adding to `_o-rtl.css` the `transform` again, but that seems wrong.

RTLCSS doesn't seem to have a way to disable a single line / selector / whatever…

ffoodd:
Oh yeah, almost found the way, using [TLCSS StringMap](https://rtlcss.com/learn/usage-guide/string-map/).

Lausselloic:
Thanks @ffoodd i've merge it, and will include it in the next release

Lausselloic:
after testing the build i've fix like that : 
```
    /*rtl:ignore*/
    transform: rotate(90deg);
    /*rtl:ignore*/
```
see : https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/commit/51023a67038f6c7efb3df454f3a19530308508bf

ffoodd:
I didn't find any other part that should be ignored from RTLCSS, but there might be. Nice to have this, though. Thanks for merging :)

PR #96, opened 2019-01-16
Using RTLCSS ignore rule for #95

ffoodd:
There might be other things to ignore from RTLCSS…


ffoodd:
And obviously you should try compiling, since I only made an example :)

Issue #134, opened 2019-04-04
Custom checkboxes - Encoutered an issue to deselect it


Handja:
Hi,

I've encountered an issue with custom checkboxes.
When I select it and after that i want to deselect it, I need to to click on the icon inside it.
If I click on the orange background nothing happens.

So some users think it's a bug.

https://boosted.orange.com/docs/4.2/components/forms/#checkboxes-and-radios-1

Lausselloic:
Hello,
Oh you're right :-(
Just reproduce it on Chrome win7

We will try to fix it

liyokuna:
#135

Lausselloic:
@Handja fix will be available in the next release (4.3.1)

Handja:
Ok thx you :)

PR #135, opened 2019-04-04
fix(custom-checkbox): remove unnecessary code #134

liyokuna:
Enabling the click on the orange background of a custom checkbox.


Lausselloic:
Tested on FF,Chrome and IE11 works like a charm. Thanks

Issue #255, opened 2019-12-13
prefers-reduced-motion is duplicated (a lot)


ffoodd:
For now, resetting transition when user `prefers-reduced-motion` is done for each `@include transition()`, which makes a lot of code duplication.

It should be added either as a basic or in reboot, top reduce **every** animation and transition, instead of being duplicated everywhere…

PR #258, opened 2019-12-13
Feature/reduced motion

ffoodd:
Apply `prefers-reduced-motion` globally for #255.

Could / should be a proper PR for Bootstrap, I guess.


ffoodd:
Just created [an issue on Bootstrap](https://github.com/twbs/bootstrap/issues/29862) to suggest this.

GaetanBt:
:heart:

PR #270, opened 2020-01-03
Bundle focus visibility & improve wesbite performances

ffoodd:
There's also a few scripts loading improvements on the website part.

Can revert some commits if required by reviewers :100:


ffoodd:
@Lausselloic I merge this, however I won't delete the branch to ease your review.

Two things to review specifically:
1. rollup config cleanup, since Swiper and TableSorter bundling never worked (because of Rollup using tree shaking): in `v4.3.1`, they're not bundled in `dist` files → https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/v4.3.1/dist/js/boosted.bundle.js
2. focus-visible inclusion, since playing in rollup won't work beceause of tree shaking, it's imported in `index.js` directly. Inspired by ob1 with focus-within polyfil.

This branch also plays around script loading in the docs, but no need to review this :)

PR #507, opened 2020-10-01
Improved docs

Sakshi525:


ffoodd:
This is a corporate fork, this tone is way too familiar.
If you're trying things out for Kacktober Fest, please find a real issue to fix and stop spamming.

PR #561, opened 2020-11-17
Fix BrowserStack failures

ffoodd:
This reverts commit a539688f68c8f07d047b1ec4eabdb0642d60e745 to ensure Karma runs on IE11 in BrowserStack.


ffoodd:
Maybe completely drop Dependabot for v4 (added with #554) and updated deps.

PR #612, opened 2021-01-22
CI optimizations

ffoodd:
Reduce our ecological footprint, somehow?

- [Skip action when a specific label is added or removed](https://github.community/t/how-can-i-keep-github-event-pull-request-labels-up-to-date-when-re-trying-a-workflow/119386/2)
- [Skip actions when an existing label is set](https://github.community/t/skip-action-when-label-already-on-pr-vs-on-label-create-event/121037/2)

Targets:
- [x] Percy has a snapshots quota per month,
- [x] BrowserStack runs several VMs…
- [x] Netlify deploys a technical URL
- [x] Pa11y is quite heavy too.

The other ones already fit their target: either by filetype (e.g. compile, lint, test, optimize images) or event type (e.g. notify on published release). I guess we're able to add a few labels, such as `skip:percy`, `skip:browserstack`, etc. And `skip:ci` should cancel all workflows, I guess.

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -23,16 +23,23 @@ jobs:
     timeout-minutes: 30
 
     steps:
+      - name: Bail early
+        if: contains(github.event.*.labels.*.name, 'skip:ci') || contains(github.event.*.labels.*.name, 'skip:browserstack')
+        run: exit 0
+
       - name: Clone repository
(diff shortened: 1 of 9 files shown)


ffoodd:
[Docs added to the wiki](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/wiki/GitHub-Actions): let's try this!

Issue #525, opened 2020-10-19
ODS > Global Headers (Navbars)


ffoodd:
## Requirement

**For** a Developer 
**Who** is creating an Orange Website
**The** Header Integration
**is an** automated component
**that** proposes the web site's Orange Headers
**unlike** the default bootstrap framework
**Boosted** will provide this ready to use component compatible with the Orange Design and the accessibility requirements

### General Integration context
The Orange page Header is divided into 3 parts that are stacked on top of each other:
* The Supra Bar (Optional): a bar above all others that allows to select from global website context (Personal or Business / language selection) - it can also contain site action icons 
* The Global Header Bar (Mandatory): a bar that must contain the Orange or OBS Logo - it can also contain the site Title,  the Sites navigation menu, the site icons or the Search Input.
* The Stand Alone Navigation Bar (Optional): this bar only shows the navigation items

The integrating team will decide how to customize the header using these various elements to best organize the site.

The header is always sticky, i.e. it must always be displayed.

#### Supra bar menu items
* The menu items will be on a single line and will be pulled to the left of the bar and/or pulled to the right of the bar
* The menu items should reference a full site context change and should not generate sub menu items 
* The language selection can be placed in the Supra bar, in that case it should be placed on the right hand side of the bar

#### Header Action Icons
* The action icons are placed on the far right of the bar
* The action icons can include User Profile, Purchasing Cart, Search,...
* When resizing the screen the action icons should always be visible

#### Logo
* The Orange logo (aka **Master Logo**)
  * it should always be sized 50px 
  * it should never be resized
  * when the corresponding breakpoint is reached it should be switched to the small logo sized at 30px that should not be resized
* The OBS logo
  * Is the The OBS logo"Business Services" Text on 2 separate lines
  * it should stay at 50px height until the corresponding break point is reached
  * it should then stay at 30px height for the rest of the resizing
  * the OBS logo is always presented the same and should not adapt to Right to Left text positioning
* Other Logos are not taken into account in Boosted (Orange Bank, Parnasss, ...)

#### Title
* The title is left align next to the logo - Xpx margin between the two
* The Title can be set to be on 1 line or 2 lines
* The font size is 
  * 1 line: 35px
  * 2 lines: 29px
* On Resize of the screen:
  * The Title should keep the same size
  * When the text does not fit anymore the title should be removed
  * 

#### Menu items
* The menu items are listed
  * in the Global Header Bar:
[...]

PR #622, opened 2021-02-05
Improve focus accessibility when fixed header/footer overlaps

ffoodd:
Focus might be hidden behind fixed header: this is a first attempt to provide a simple and generic solution.

- [x] ~~To improve when Navbar minimizing behaviour gets implemented~~ → mentioned in #525 
- [x] ~~Does it need JavaScript? I'm afraid so…~~ No it doesn't, welcome `scroll-padding-top` on `:root` :heart:
- [x] Suggest to Bootstrap, at least for docs (to replace `:target { scroll-margin-top: … }`) → [BS #33027](https://github.com/twbs/bootstrap/pull/33027)

### DIFF
--- a/scss/_back-to-top.scss
+++ b/scss/_back-to-top.scss
@@ -1,7 +1,9 @@
 // Boosted only
 // See https://moderncss.dev/pure-css-smooth-scroll-back-to-top/
-[id="#{$back-to-top-target-id}"]:target {
-  scroll-margin-top: $back-to-top-target-offset-top;
+@if $enable-fixed-header {
+  [id="#{$back-to-top-target-id}"]:target {
+    scroll-margin-top: $back-to-top-target-offset-top;
+  }
 }
 
 .back-to-top {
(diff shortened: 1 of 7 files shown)


ffoodd:
Digging up an oldish jQuery fix for this:

```js
// Handle focus behind sticky header
var stickyHeader = $('.header').get(0);
// get the sticky element's height (can be done on mediaquery change instead of each time)
var headerHeight = stickyHeader.clientHeight;

function handleFocus(e) {
  // don't try to change the scroll if the focused element is in the sticky element
  if ($.contains(stickyHeader, e.target) || e.target.tagName === 'MAIN') {
    return;
}

$(document).keyup(function (e) {
    if (e.which === 9) {
      var focused = document.activeElement;
      // get the screen position of the focused element
      var focusTop = focused.getBoundingClientRect().top;
      // get the current scroll
      var scrollTop = document.documentElement.scrollTop || document.body.scrollTop;
      // scroll when the focused element is under the sticky element
      if (focusTop < headerHeight) {
        $('html, body').scrollTop(scrollTop - headerHeight);
      }
    }
  });
}
```

ffoodd:
`scroll-padding-top` to the rescue, it works like a charm :rocket:

PR #636, opened 2021-02-19
feat: Stepped Process: add the component from v4

### LINKED ISSUE #524: Stepped process (v4:stepbar)
1. [bs-stepper](https://github.com/Johann-S/bs-stepper) ([Bootstrap #31014](https://github.com/twbs/bootstrap/issues/31014))
2. back from #221, only technical variants :
    - since next steps could be unavailable in a tunneling process, we should at least show an example with **disabled** next steps links.
    - as a matter of fact, one could also want to use **headings** for step titles instead of a list. Should we handle this?

v4 ref:
- https://boosted.orange.com/docs/4.5/components/stepbar/

Nurovek:
Stepbar is being ported into v5 as "stepped process" component.

Everything has been adapted according to this change

---

Closes #524 

---

- [x] Should match specs (eg. either the Web UI Kit or any pattern from the WAI — or both…)
- [x] Docs added:
  - including the "Sass" part using `scss-docs` shortcode
  - in Design Guidelines probably
  - in Cheatsheet / Cheatsheet RTL examples
- [x] Check (and fix) RTL version
  - [x] Beware of clip-path in RTL, which would need some attention—and maybe the loop setting z-index too, might even be dropped?
- [x] Run linters
- [x] Run compilers
- [x] Run tests
- [x] Cross-browser test:
  - Firefox ESR
  - Latest Edge, Chrome, Firefox, Safari
  - iOS Safari
  - Chrome & Firefox on Android
- [x] Clean up the branch using `rebase -i`
- [x] Commited with `feat(…): …` message
- [x] Mention it 
  - in Ship List 
  - in Migration Guide (if `back-from-v4`): renamed variables, changes in markup requirement,  etc.

---

- Preview: https://deploy-preview-636--boosted.netlify.app/docs/5.0/components/stepped-process/
- X-ref v4: https://boosted.orange.com/docs/4.6/components/stepbar/

### DIFF
--- a/scss/_stepped-process.scss
+++ b/scss/_stepped-process.scss
@@ -1,0 +1,124 @@
+//
+// Orange stepped process
+//
+.stepped-process {
+  ol {
+    display: flex;
+    padding: 0;
+    margin: 0;
+    overflow: hidden;
+    font-size: $small-font-size;
+    font-weight: $font-weight-bold;
+    text-align: center;
+    list-style: none;
+    counter-reset: $stepped-process-counter;
+  }
+}
+
+.stepped-process-item {
+  display: flex;
+  flex: 1;
+  padding: $step-item-padding;
+  margin-right: $step-item-margin-end;
+  counter-increment: $stepped-process-counter;
+  background-color: $step-item-bg;
+
+  &.active {
+    background-color: $step-item-active-bg;
+  }
+
+  .active ~ & {
+    background-color: $step-item-next-bg;
+  }
+}
+
+.stepped-process-link {
+  flex: 1 0 $step-link-width;
+  max-width: $step-link-width;
+  margin: auto;
+  overflow: hidden;
+  line-height: 1;
+  color: $step-link-color;
+  text-decoration: if($link-decoration == none, null, none);
+  white-space: nowrap;
+  outline-offset: $spacer;
+
+  &::before {
+    content: $step-link-marker;
+  }
+
+  &:hover,
+  &:focus {
+    color: $step-link-color;
+  }
+
+  &:hover {
+    text-decoration: if($link-hover-decoration == underline, null, underline);
+  }
+
+  &:focus {
+    text-decoration: $link-decoration;
+    outline-offset: $spacer / 4;
+  }
+
+  .active &,
+  .active ~ .stepped-process-item & {
+    color: $step-link-active-color;
+  }
+}
+
+@include media-breakpoint-up(sm) {
+  .stepped-process-item {
+    position: relative;
+    padding: $step-item-padding 0;
+    margin: 0;
+    filter: $step-item-drop-shadow;
+
+    & + & {
+      padding-left: $step-item-padding-end;
+    }
+
+    &:not(:last-child)::after {
+      position: absolute;
+      top: 0;
+      bottom: 0;
+      left: subtract(100%, 1px);
+      z-index: -1;
+      width: $step-item-arrow-width;
+      content: "";
+      background-color: inherit;
+      clip-path: $step-item-arrow-shape;
+    }
+  }
+
+  @for $i from 1 through $stepped-process-max-items {
+    .stepped-process-item:nth-child(#{$i}) {
+      $index: subtract($stepped-process-max-items, $i);
+      z-index: if($index == 0, null, $index);
+    }
+  }
+
+  .active .stepped-process-link {
+    max-width: none;
+
+    &::before {
+      content: $step-link-marker-lg;
+    }
+  }
+}
+
+@include media-breakpoint-between(sm, xl) {
+  .stepped-process-item.active {
+    flex: $stepped-process-max-items - 1;
+  }
+}
+
+@include media-breakpoint-up(xl) {
+  .stepped-process-link {
+    max-width: none;
+  }
+
+  .stepped-process-link::before {
+    content: $step-link-marker-lg;
+  }
+}
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1643,4 +1643,28 @@ $back-to-top-icon:               var(--#{$boosted-variable-prefix}chevron-icon) 
 $back-to-top-icon-width:         add(.5rem, 1px) !default;
 $back-to-top-icon-height:        subtract(1rem, 1px) !default;
 // scss-docs-end back-to-top
+
+//// Stepped process
+// scss-docs-start stepped-process
+$stepped-process-max-items: 5 !default;
+$stepped-process-counter:   step !default; // Used as a counter name
+
+$step-item-margin-end:   $border-width !default;
+$step-item-padding:      .5rem !default;
+$step-item-bg:           $black !default;
+$step-item-active-bg:    $primary !default;
+$step-item-next-bg:      $gray-400 !default;
+$step-item-shadow-size:  $border-width * 1.5 !default;
+$step-item-drop-shadow:  drop-shadow($step-item-shadow-size 0 0 $white) #{"/* rtl:"} drop-shadow(-$step-item-shadow-size 0 0 $white) #{"*/"} !default;
+$step-item-padding-end:  $step-item-padding * 2 !default;
+
+$step-link-width:        1.25ch !default; // Matches width of a single number
+$step-link-color:        $white !default;
+$step-link-active-color: $black !default;
+$step-link-marker:       counter($stepped-process-counter) "\A0" !default;
+$step-link-marker-lg:    counter($stepped-process-counter) ".\A0" !default;
+
+$step-item-arrow-width:  1rem !default;
+$step-item-arrow-shape:  polygon(0% 0%, subtract(100%, $border-width) 50%, 0% 100%) #{"/* rtl:"} polygon(100% 0%, $border-width 50%, 100% 100%) #{"*/"} !default; // Used in clip-path
+// scss-docs-end stepped-process
 // End mod
--- a/site/content/docs/5.0/components/stepped-process.md
+++ b/site/content/docs/5.0/components/stepped-process.md
@@ -1,0 +1,48 @@
+---
+layout: docs
+title: Stepped process
+description: Stepped process bar used for multiple steps forms process
+group: components
+toc: true
+---
+
+## How to use
+
+Use a `nav` element with `.stepped-process` class, containing an ordered list `<ol>` with `.stepped-process-item` class on list items.
+
+Use short labels for each step, otherwise they will be cut off to preserve inline layout. More information can be carried on the `title` attribute of the `.stepped-process-link`.
+
+Add `.active` to a `.stepped-process-item` to indicate the current step, alongside `aria-current="step"` to convey the active state to assistive technologies.
+
+## Example
+
+{{< example >}}
+<nav class="stepped-process" aria-label="Checkout process">
+  <p class="float-start mt-2 me-2 fw-bold d-sm-none">Step</p>
+  <ol>
+    <li class="stepped-process-item">
+      <a class="stepped-process-link" href="#" title="1. Sign in">Sign in</a>
+    </li>
+    <li class="stepped-process-item active">
+      <a class="stepped-process-link" href="#" title="2. Review" aria-current="step">Review</a>
+    </li>
+    <li class="stepped-process-item">
+      <a class="stepped-process-link" href="#" title="3. Delivery">Delivery</a>
+    </li>
+    <li class="stepped-process-item">
+      <a class="stepped-process-link" href="#" title="4. Payment">Payment</a>
+    </li>
+    <li class="stepped-process-item">
+      <a class="stepped-process-link" href="#" title="5. Place order">Place order</a>
+    </li>
+  </ol>
+</nav>
+{{< /example >}}
+
+## Sass
+
+### Variables
+
+For more details, please have a look at the exhaustive list of available variables:
+
+{{< scss-docs name="stepped-process" file="scss/_variables.scss" >}}
--- a/site/content/docs/5.0/examples/cheatsheet-rtl/index.html
+++ b/site/content/docs/5.0/examples/cheatsheet-rtl/index.html
@@ -65,6 +65,7 @@ direction: rtl
           <li><a class="d-flex" href="#progress">شريط التقدم</a></li>
           <li><a class="d-flex" href="#scrollspy">مخطوطة</a></li>
           <li><a class="d-flex" href="#spinners">المغازل</a></li>
+          <li><a class="d-flex" href="#stepped-process">عملية متدرجة</a></li>
           <li><a class="d-flex" href="#toasts">نخب</a></li>
           <li><a class="d-flex" href="#tooltips">تلميحات الأدوات</a></li>
         </ul>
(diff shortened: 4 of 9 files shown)


site/content/docs/5.0/components/stepped-process.md
```diff
@@ -0,0 +1,40 @@
...
+layout: docs
+title: Stepped process
+description: Stepped process bar used for multiple steps forms process
+group: components
+toc: true
+---
+
+## How to use
+
+Use a `nav` element with `.stepped-process` class, containing an ordered list `<ol>`  with `.stepped-process-item` class on list items.
```
ffoodd, line 11:
```suggestion
Use a `nav` element with `.stepped-process` class, containing an ordered list `<ol>` with `.stepped-process-item` class on list items.
```

site/content/docs/5.0/components/stepped-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Be careful to choose short labels when using multiple steps process, otherwise label will be cut to preserve inline layout. More information can be carried on the `title` attribute of the `.stepped-process-link`.
```
ffoodd, line 13:
```suggestion
Use short labels for each step, otherwise they will be cut off to preserve inline layout. More information can be carried on the `title` attribute of the `.stepped-process-link`.
```

Regarding `title`, @GaetanBt shouldn't `aria-label` be used instead, in such cases?

site/content/docs/5.0/components/stepped-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Use `.current` class to define current step, alongwith with `aria-current="step"`.
```
ffoodd, line 15:
```suggestion
Add `.active` to a `.stepped-process-item` to indicate the current step, alongside `aria-current="step"` to convey the active state to assistive technologies.
```

BTW, we should move from `.current` to `.active` because `.active` is used throughout Bootstrap component as a convention, whereas `.current` only pops out in our custom components…

site/content/docs/5.0/components/stepped-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Example
+
+{{< example >}}
+<nav class="stepped-process" aria-label="Checkout process">
+  <p class="float-left mt-2 mr-2 font-weight-bold d-sm-none">Step</p>
```
ffoodd, line 21:
```suggestion
  <p class="float-end mt-2 me-2 fw-bold d-sm-none">Step</p>
```

Beware that some utilities got renamed in v5: you should have seen the difference visually (e.g. not inline on small screens, and not bold either).

scss/_stepped-process.scss
```diff
@@ -0,0 +1,139 @@
...
+}
+
+.stepped-process-item {
+  display: flex;
+  flex: 1;
+  padding: map-get($spacers, 2);
+  margin-right: $stepped-process-margin-r;
+  font-weight: $font-weight-bold;
+  text-align: center;
+  list-style: none;
```
ffoodd, line 21: I think those could move up to `ol` declaration above. Need to check, but it'd be simpler to understand.

scss/_stepped-process.scss
```diff
@@ -0,0 +1,139 @@
...
+    .stepped-process-link {
+      &,
+      &:hover,
+      &:focus,
+      &:active {
+        color: $black;
+      }
+    }
+  }
+}
```
ffoodd, line 50: I think all of this could be simplified by using custom properties: I'll give it a shot to show you.

scss/_stepped-process.scss
```diff
@@ -0,0 +1,139 @@
...
+
+  &:focus {
+    text-decoration: $link-decoration;
+    outline-offset: map-get($spacers, 1);
+  }
+}
+
+@include media-breakpoint-up(sm) {
+  .stepped-process-item {
+    max-width: 3.75rem;
```
ffoodd, line 84: Hardcoded magical values like this, and most of values based on variables, should be defined as variables in `_variables.scss` for the component. We'll do this at some point.

scss/_stepped-process.scss
```diff
@@ -0,0 +1,139 @@
...
+    padding: map-get($spacers, 2) 0 map-get($spacers, 2) $spacer;
+    margin: 0;
+
+    &::after {
+      display: inline-block;
+      width: $stepped-process-arrow-width;
+      height: $stepped-process-bar-height;
+      margin: -#{map-get($spacers, 2)} 0 -#{map-get($spacers, 2)} auto;
+      content: "";
+      background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 17 40'><path fill='#{$black}' d='M0 0l14 20L0 40z'/><path fill='#{$white}' d='M3 0H0l14 20L0 40h3l14-20z'/></svg>"));
```
ffoodd, line 94: This (and maybe the entire pseudo-element) could be dropped in favor of `mask`: I'll tackle this in another PR, too. :)

ffoodd (changes requested):

site/content/docs/5.0/components/stepped-process.md
Nurovek: Fact is I didn't see a thing, probably because of the `d-sm-none` so I didn't test all breakpoints. My bad...
Secondly I didn't look for changes on utilities, my bad twice.

(In our case, the label should be at the start, I corrected accordingly to this case)

scss/_stepped-process.scss
Nurovek: Yes I could have been more careful, because I took it from v4 as-is.

site/content/docs/5.0/components/stepped-process.md
ffoodd: Yup, that's what reviews and tests are for ;)

site/content/docs/5.0/components/stepped-process.md
Nurovek: In the meantime, can't we put both title & aria-label ?

ffoodd:
New side topic here, now that we merged `main` recently: add a "Sass# part in the docs, using the `scss-docs` shortcode to display at least variables.

scss/_variables.scss
```diff
@@ -1644,3 +1644,10 @@ $back-to-top-icon-width:         add(.5rem, 1px) !default;
 $back-to-top-icon-height:        subtract(1rem, 1px) !default;
 // scss-docs-end back-to-top
 // End mod
```
ffoodd, line 1646: This comment should stay at the end of the file, preferably (stepped process is still a Boosted mod).

ffoodd (changes requested):

scss/_variables.scss
```diff
@@ -1644,3 +1644,10 @@ $back-to-top-icon-width:         add(.5rem, 1px) !default;
...
+
+// Stepped process
+
+$stepped-process-bar-height:  $spacer * 2 !default;
+$stepped-process-arrow-width: 1.0625rem !default;
+$stepped-process-margin-r:    $border-width !default;
+$stepped-process-item-max-width: 3.75rem !default;
```
ffoodd, line 1653: Also we try to align variables in the same group (on the last one in that case, would be good).

scss/_variables.scss
```diff
@@ -1643,4 +1643,29 @@ $back-to-top-icon:               var(--#{$boosted-variable-prefix}chevron-icon)
...
+$stepped-process-counter:   step !default; // Used as a counter name
+
+$step-item-max-width:    3.75rem !default;
+$step-item-margin-end:   $border-width !default;
+$step-item-padding:      $spacer / 2 !default;
+$step-item-bg:           $black !default;
+$step-item-active-bg:    $primary !default;
+$step-item-next-bg:      $gray-400 !default;
+$step-item-shadow-size:  $border-width * 1.5 !default;
+$step-item-drop-shadow:  drop-shadow($step-item-shadow-size 0 0 $white) #{"/* rtl:"} drop-shadow(-$step-item-shadow-size 0 0 $white) #{"*/"} !default;
```
ffoodd, line 1659: Simpler version, needs to check if that works:
```suggestion
$step-item-drop-shadow:  drop-shadow(#{"/*rtl:insert:-*/"}$step-item-shadow-size 0 0 $white) !default;
```

We're using this insertion in `$font-family-sans-serif` in `_variables.scss` for example. Useful when you don't need to change the whole value but only to add something in the existing one.

site/content/docs/5.0/examples/cheatsheet-rtl/index.html
```diff
@@ -1418,6 +1418,36 @@ <h3>ترقيم الصفحات</h3>
         {{< /example >}}
       </div>
     </article>
+    <article class="my-3" id="stepped-process">
+      <div class="bd-heading sticky-xl-top align-self-start mt-5 mb-3 mt-xl-0 mb-xl-2">
+        <h3>Stepped process</h3>
```
ffoodd, line 1423: This heading isn't translated. From Google Translate:
```suggestion
        <h3>عملية متدرجة</h3>
```

site/content/docs/5.0/examples/cheatsheet-rtl/index.html
```diff
@@ -1418,6 +1418,36 @@ <h3>ترقيم الصفحات</h3>
...
+              <a class="stepped-process-link" href="#" title="2. Step two" aria-current="step">الخطوة الثانية</a>
+            </li>
+            <li class="stepped-process-item">
+              <a class="stepped-process-link" href="#" title="3. Step three">الخطوة الثالثة</a>
+            </li>
+            <li class="stepped-process-item">
+              <a class="stepped-process-link" href="#" title="4. Step four">الخطوة الرابعة</a>
+            </li>
+            <li class="stepped-process-item">
+              <a class="stepped-process-link" href="#" title="5. Step five">الخطوة الخامسة</a>
```
ffoodd, line 1444: `title` attributes need to contain the textual, visible content (this is a generic rule regarding accessibility).
In that case you'd need to replace e.g. `Step one` with `عملية متدرجة` in the first item's `title` attribute.

site/content/docs/5.0/examples/cheatsheet-rtl/index.html
```diff
@@ -1418,6 +1418,36 @@ <h3>ترقيم الصفحات</h3>
         {{< /example >}}
       </div>
     </article>
+    <article class="my-3" id="stepped-process">
```
ffoodd, line 1421: You didn't add a link in the sidebar nav, leading to this component.

site/content/docs/5.0/guidelines/navigation.md
```diff
@@ -270,7 +270,31 @@ toc: true
 </div>
 
 ## Stepped process
+[Documentation]({{< docsref "/components/stepped-process" >}})&nbsp;—&nbsp;
```
ffoodd, line 273: Missing an empty line between heading and link.

site/content/docs/5.0/guidelines/navigation.md
```diff
@@ -270,7 +270,31 @@ toc: true
...
+<span class="d-none d-sm-block d-lg-block">{{< anchor web-nav-stp-001 >}}</span>
+<span class="d-block d-sm-none d-lg-none">{{< anchor web-nav-stp-002 >}}</span>
```
ffoodd, line 275: Shoudln't we have three IDs here?

site/data/sidebar.yml
```diff
@@ -76,6 +76,7 @@
     - title: Popovers
     - title: Progress
     - title: Scrollspy
+    - title: Stepped process
```
ffoodd, line 79: I'd say it should be after "Spinners# alphabetically, shouldn't it?

ffoodd (changes requested):

site/content/docs/5.0/guidelines/navigation.md
Nurovek: In Zeplin, "Normal" and "mobile" have the same ID

ffoodd:
## Design

From my first understanding, the design published in Zeplin uses a smaller font-size (my guess: `.875rem`) and probably smaller vertical `padding` too, since the bar is only `30px` tall on Zeplin and `38px` in Boosted.

## Improvements

Also the intermediary display (between `sm` and `lg`), the last item does not have enough space. Here's my recommendation:
- [ ] drop the `&:last-child` part L80 that sets `padding-right`
- [ ] drop `max-width` on `..stepped-process-item` (L69, L96, L118)
- [ ] add an exception for the `.active` item between `sm` and `lg` breakpoints:

```scss
@include media-breakpoint-between('sm', 'lg') {
  .stepped-process-item.active {
    flex: $stepped-process-max-items - 1;
  }
}
```

This should output `flex: 4` with the current settings — and we only rely on flexbox. :tada:

## Docs

In the guidelines, please tend to re-use UI Kit / Zeplin's wording for your examples, in order to match as much as possible the UI Kit. So your items should be "Sign in", "Review", Delivery", "*Payment", "Place order".

scss/_variables.scss
Nurovek: That was a good idea, except it seems sass adds a space, breaking the rtl version. Closing :(

site/content/docs/5.0/guidelines/navigation.md
ffoodd: Well, good game GDL :rocket:

Nurovek: Only "small" has another ID

scss/_stepped-process.scss
```diff
@@ -0,0 +1,122 @@
...
+  .stepped-process-link::before {
+    content: $step-link-marker-lg;
+  }
+}
+
+@include media-breakpoint-between("sm", "lg") {
+  .stepped-process-item.active {
+    flex: $stepped-process-max-items - 1;
+  }
+}
```
ffoodd, line 122: Fixing this by myself since I made a few mistakes in my first suggestion:
1. should be `xl`, not `lg`,
2. should be used without quotes around breakpoint keys,
3. since it's meant to be between the two breakpoints, place it consistently in your source ;)

```suggestion
@include media-breakpoint-between(sm, xl) {
  .stepped-process-item.active {
    flex: $stepped-process-max-items - 1;
  }
}

@include media-breakpoint-up(xl) {
  .stepped-process-link {
    max-width: none;
  }

  .stepped-process-link::before {
    content: $step-link-marker-lg;
  }
}
```

ffoodd (changes requested):

scss/_variables.scss
```diff
@@ -1643,4 +1643,28 @@ $back-to-top-icon:               var(--#{$boosted-variable-prefix}chevron-icon)
...
 $back-to-top-icon-height:        subtract(1rem, 1px) !default;
 // scss-docs-end back-to-top
+
+//// Stepped process
+// scss-docs-start stepped-process
+$stepped-process-max-items: 5 !default;
+$stepped-process-counter:   step !default; // Used as a counter name
+
+$step-item-margin-end:   $border-width !default;
+$step-item-padding:      $spacer / 2.5 !default;
```
ffoodd, line 1653: I'd go with `.5rem` instead of this: `$spacer / 2` is a value in our `$spacers` map, `/ 2.5` is not.

ffoodd (changes requested):

scss/_variables.scss
Nurovek: I hesitated but ok, go for it (maybe should I store it into a variable too ?)

PR #638, opened 2021-02-22
Refactor stepped process

ffoodd:
| File                        | `feature/stepped-process` | current refactor |
| ------------------------ | ---------------------------------- | --------------------- |
| `boosted.css`        |                             26.54ko |             26.49ko |
| `boosted.min.css` |                             24.24ko |             24.19ko |

Small yet useful :p

- [ ] Beware of `clip-path` in RTL, which would need some attention—and maybe the loop setting `z-index` too, might even be dropped?

### DIFF
--- a/scss/_stepped-process.scss
+++ b/scss/_stepped-process.scss
@@ -8,7 +8,6 @@
     margin: 0;
     overflow: hidden;
     font-weight: $font-weight-bold;
-    line-height: 1.25;
     text-align: center;
     list-style: none;
     counter-reset: step;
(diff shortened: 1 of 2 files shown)


scss/_stepped-process.scss
```diff
@@ -81,27 +74,36 @@
...
-    padding: map-get($spacers, 2) 0 map-get($spacers, 2) $spacer;
+    padding: $spacer / 2 0 $spacer / 2 $spacer;
     margin: 0;
-
-    &::after {
-      display: inline-block;
-      width: $stepped-process-arrow-width;
-      height: $stepped-process-bar-height;
-      margin: -#{map-get($spacers, 2)} 0 -#{map-get($spacers, 2)} auto;
+    filter: drop-shadow($border-width * 1.5 0 0 $white);
```
Nurovek: If I understand correctly, it is used for border between step-items, isn't it ?

PR #640, opened 2021-02-24
New examples: Standard VS Variable font // Leaflet

ffoodd:
### Standard VS Variable font

- [x] Responsive improvement for left sidebar
- [x] Social links: make them real with sharing URL
- [x] Add the Variable font version
  - [x] Requires to drop Helvetica Neue ~~(maybe rely on `boosted-utilities.css` only?)~~
  - [x] Add Helvetica Now implementation:
    - ~~drop "instances"? Does this impact file size?~~ Don't know how to do this.
- [x] Add screenshots

#### Previews
- Helvetica Neue: https://deploy-preview-640--boosted.netlify.app/docs/5.0/examples/standard-font/
- Helvetica Now: https://deploy-preview-640--boosted.netlify.app/docs/5.0/examples/variable-font/

---

### Leaflet

- [x] Scratch a first example based on startup.orange.com existing map
- [x] Get IoT Map Component examples, markers and popup…
  - [ ] Maybe add an interface to handle `createIcon()` arguments?
  - [ ] Maybe randomize arguments for each marker, as a demonstration?
- [x] Add screenshot

#### Preview

- https://deploy-preview-640--boosted.netlify.app/docs/5.0/examples/leaflet/


ffoodd:
To get this done @Nurovek:
1. needs a rebase to deploy properly `_site` on Netlify (Pa11y is also failing because of this)
2. improve Leaflet example a bit I guess, see the two missing steps in the description

Then you can merge this, I guess. Feel free to ping me for any questions :)

ffoodd:
Sidenote while I still think about this :p 

In order to include "map" as a component in Boosted, here are a few suggestions—that's not a to-do list, improving the current example in this PR is the first step, but those suggestions might be the next ones:
- markers & clusters could be handled CSS-side, using `background-image` or `mask-image` (basically URL encode SVGs, as done in other parts of Boosted) or `clip-path` in some cases.
- popups could be handled as a component too, maybe mixing cards and modals for style consistencies.
- could also dig on `leaflet.css` side: it might be possible to drop this file entirely, and replacing it with our own (using our helpers and utilities as much as possible).

BTW I think controls (like zoom in/out buttons) could be customized and aligned with Orange brand. That'd make sense. You'll need to power up on Leaflet customization.

PR #647, opened 2021-03-02
chore(merge v4-dev): patched commit → 8ec32ee

Nurovek:


ffoodd (approved):
That was an easy one :dancing_men:

Nurovek:
> That was an easy one 👯‍♂️

Indeed, which is quite a good news I suppose, it means no breaking change was introduced in the current version :D

PR #654, opened 2021-04-22
fix(docs): some '[comment]: boosted mod' appear in the documentation

julien-deramond:
In some parts of the 4.6 documentation, we can see the following string `[comment]: #Boosted mod ` that shouldn't be there. This PR transforms those strings into comments as seen in other markdown files in the project.

Example: https://boosted.orange.com/docs/4.6/components/alerts/#additional-content


ffoodd (approved):
:+1:

Lausselloic (approved):

PR #658, opened 2021-04-27
fix dropdowns on `.bg-dark`

### LINKED ISSUE #657: Dropdown items - Dark background
- Operating system: macOS
- Browsers: Firefox and Chrome

This bug is directly viewable in https://boosted.orange.com/docs/4.6/components/dropdowns/#dark-background.
The dropdown items are written in white on a white background; and so become invisible except while hovering them.
This is maybe a duplicate but I haven't found another reference to it.

Lausselloic:
Fix issue #657

Invert color of `.dropdown-menu`, and \.dropdown-menu-item` when hovered or active on `.bg-dark`  fix #657

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -64,6 +64,12 @@
   border: $dropdown-border-width solid $dropdown-border-color;
   @include border-radius($dropdown-border-radius);
   @include box-shadow($dropdown-box-shadow);
+  // Boosted mod
+  .bg-dark & {
+    color: $dropdown-bg;
+    background-color: $dropdown-color;
+  }
+  // End mod
 }
 
 @each $breakpoint in map-keys($grid-breakpoints) {
@@ -177,6 +183,12 @@
     color: $dropdown-link-hover-color;
     text-decoration: none;
     @include gradient-bg($dropdown-link-hover-bg);
+    // Boosted mod
+    .bg-dark & {
+      color: $dropdown-link-hover-bg;
+      @include gradient-bg($dropdown-link-hover-color);
+    }
+    // End mod
   }
 
   &.active,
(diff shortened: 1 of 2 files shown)


julien-deramond:
Something is not coherent between the background colors when focused: the first dropdown button stays black.



I would rather expect something like that (the first dropdown button staying white):



I would say it is somewhat linked to the line 17 of __dropdown.scss_.

ffoodd:
Do you have access to GDL's Zeplin, @julien-deramond and @Lausselloic?

You should, it'd really hello you with such cases.

julien-deramond:
> Do you have access to GDL's Zeplin, @julien-deramond and @Lausselloic?
> 
> You should, it'd really hello you with such cases.

Yes, I do. The image below is a summary of what I found:

Frame 1

ffoodd:
I guess dropdowns states should be the same than inputs but that's somehow inconsistent.

CyriaqueBillard:
Hello,

There is an interaction model for the Forms components family which can be seen here: https://system.design.orange.com/0c1af118d/p/88ab5b-forms/b/599459
There is also an interaction model for the Buttons components family which can be seen here: https://system.design.orange.com/0c1af118d/p/04ae65-buttons/b/48012d

Closed and Open states of dropdown lists should be kept as defined in the Web UI Kit, otherwise it breaks the consistency of the family's interaction model.
Split button comes with two hover state variants so the behavior can be aligned with the behavior of the hover state on dropdown list.

scss/_dropdown.scss
```diff
@@ -14,7 +14,25 @@
   background-color: $dropdown-bg;
   border-color: $dropdown-border-color;
 
-  .bg-dark &:not(.dropdown-toggle-split),
+  .bg-dark &:not(.dropdown-toggle-split) {
+    border-color: $gray-700;
```
Lausselloic, line 18: I think that border color is managed by `.btn-secondary .btn-inverse` style

julien-deramond (approved):
IMO, the design spec [in the DSM](https://system.design.orange.com/0c1af118d/p/910b9b-dropdown/b/04c480) are not complete. So I've removed the extra parts that didn't covered all the cases and kept just the important fix for the dropdown menu on `.bg-dark`.

PR #660, opened 2021-04-27
fix accordions padding / margin inconsistency 

### LINKED ISSUE #656: Accordion component issue ?
Hi
We noticed that the component accordion seem to have a padding with negative margin but value seem to be too high (0.875rem rather than 0.75rem) so text could get out of the component.
Normally it should not be seen, but in our integration as we used a div with the property overflow:hidden, everything that get out of the component is truncated

image

Could you please provide us a feedback on it ?

Lausselloic:
add a padding to `.card-body` to keep content aligned with accordion title fix #656

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -307,7 +307,7 @@
     a {
       display: flex;
       width: add(100%, $card-accordion-spacer * 2);
-      padding: $card-spacer-y;
+      padding: $card-accordion-spacer;
       margin: 0 -#{$card-accordion-spacer};
       @include font-size($h3-font-size);
       line-height: $h3-line-height;
@@ -338,7 +338,7 @@
 
     .btn-sm {
       width: add(100%, $card-spacer-y * 2);
-      padding: map-get($spacers, 2);
+      padding: $card-spacer-y;
       margin: 0 -#{$card-spacer-y};
       @include font-size($h5-font-size);
       line-height: $h5-line-height;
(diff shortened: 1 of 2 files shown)


scss/_card.scss
```diff
@@ -364,7 +364,7 @@
   }
 
   .card-body {
-    padding: 0 1.5rem 0 0;
+    padding: 0 1.5rem 0 .125rem;
```
julien-deramond, line 367: I doesn't seem to be well left-aligned with `button, a` and `.btn-sm`. But it is better for `.btn-lg`.

site/content/docs/4.6/components/collapse.md
```diff
@@ -133,7 +133,7 @@ You may use [buttons' sizes utilities]({{< docsref "/components/buttons#sizes" >
       <div class="card">
         <div class="card-header" id="headingOne-3">
           <h2 class="mb-0">
-            <button class="btn btn-link btn-sm" type="button" data-toggle="collapse" data-target="#collapseOne-3" aria-expanded="true" aria-controls="collapseOne-3">
+            <button class="btn btn-link btn-sm text-left" type="button" data-toggle="collapse" data-target="#collapseOne-3" aria-expanded="true" aria-controls="collapseOne-3">
```
julien-deramond, line 136: Could you explain to me please why adding a `.text-left` here? Can't see the difference with or without this class.

Lausselloic: adding `.text-left` because it's in the spec off accordions and on very very very small screen display is wrong otherwise 
image

scss/_card.scss
Lausselloic: 🤔 can't remember why I've added that padding. But you're right this break the alignement. Maybe need to revert this line

julien-deramond: Yep, I would revert this line.

Lausselloic: reverted

julien-deramond (approved):

PR #662, opened 2021-05-04
chore(storybook): test branch to automatically create stories

julien-deramond:
This is a test to automatically create stories from the documentation.

The important file is `stories/automatic-builder/run.js`. The rest of the files are generated by this JS file.

Run locally with a `npm run storybook`

It works rather well with alerts, accordion and badges. From there, there is a problem with everything which looks like to `{{< buttons.inline >}}`. My question is (I haven't had time for the moment to do some tests): would it be possible to handle it automatically too?


ffoodd:
You're already checking `{{< example >}}`, so I guess parsing inline shortcodes should not be much different. You'd need to store shortcode identifier (`buttons`, in `{{< buttons.inline >}}` to ensure start and end match, and use variable for start and end instead of hardcoded `example`.

Does this make sense? Is it time to regex :smile:?

julien-deramond:
I haven't tried it yet but I was a bit afraid of this kind of things: `{{< badge.inline >}} {{- range (index $.Site.Data "theme-colors") }}` 😄 I'm going to try to regex all that 😇 

From a more general point of view, I like the dynamic version in https://deploy-preview-565--boosted.netlify.app/storybook/?path=/story/components-alerts--alerts but I'm really afraid that it will demand to much work to maintain it.
I think that this PR propose a more automatic way to maintain those stories which will be used by Zeroheight (and so the designers)... if I succeed to "parse" all the examples of the documentation.

ffoodd:
Regarding loops et all, maybe switch to another approach: there're quite a lot of examples in the Cheatsheet example, you may get them more easily from its HTML output than you'd do with markdown sources?

julien-deramond:
Yep. I need to try different approaches to find the most appropriate and maintainable.
The loops would be manageable, I think, but make it a bit more complex and less maintainable. I wasn't aware of those loops when I wrote the first lines of this JS file.

I agree and also suppose it would be easier to get the examples from the HTML.

julien-deramond:
Just tried it quickly (_quick and dirty_), but this kind of technic seems to work (by using Puppeteer):

```js
const fs = require('fs')
const puppeteer = require('puppeteer');

function createTemplate(component, index, content) {
  return "export default {\n\
    title: 'Components/" + component + "'\n\
  }\n\
  \n\
  export const " + component + "_" + index + " = () => {\n\
    return `" + content + "`\n\
  }";
}

(async() => {    
  const browser = await puppeteer.launch();
  const page = await browser.newPage();    
  await page.goto(`file://${__dirname}/../../_site/docs/5.0/components/badge/index.html`);
  
  const e = await page.evaluate(() => 
    Array.from(document.querySelectorAll('.bd-example'), 
  e => e.innerHTML));

  let index = 0;
  e.forEach(example => {
    try {
      fs.writeFileSync('./stories/badge_' + index + '.stories.js', createTemplate('badge', index, example))
    } catch (err) {
      console.error(err);
    }
    index++;
  })

  await browser.close();    
  })();
```

ffoodd:
Qucik and dirty feels fine for a POC in another POC :dancers:

julien-deramond:
After a few tests.

Everything's not working of course:

- Problems when there are comments in examples (but they could be removed easily)
- Some resources are not displayed/accessible but I could modify their URL to point to the online Boosted documentation (or try to embed them in the static Storybook site)
- Boosted JS is not executed but this is because there is a problem with the beta2 (will be fixed in beta3 for sure)
- We could generate the stories only when generating the Storybook static website

The Puppeteer (using HTML doc examples) works better than the first idea of parsing the Markdown.
There's no need to install explicitly puppeteer because always installed via `pa11y`.

This seems to be a good solution for having automatic versions of the Boosted examples. We still could generate specific stories if needed (more dynamic with interactions thanks to the Controls addon).

Can I merge this POC into the original POC?

Lausselloic (approved):
Need to see to remove stories from git repo. Only publish stories to pages

PR #663, opened 2021-05-06
fix(docs): Orange components naming harmonization

vogloblinsky:
Just a little PR to clean Orange components naming in left menu


config.yml
```diff
@@ -50,6 +50,7 @@ params:
   current_version:      "4.6.0"
   current_ruby_version: "4.6.0"
   docs_version:         "4.6"
+  docs_version_short:   "4"
```
Lausselloic, line 53: as we follow bootstrap version, maybe use `current_version` to link to the tree, and just prefix with the v in footer link?

Lausselloic (changes requested):

config.yml
vogloblinsky: The tree on Bootstrap repository is `v4-dev` and not `v4.6.0-dev`

Lausselloic: https://github.com/twbs/bootstrap/tree/v4.6.0 it's also tree for tags that are independent from branch that could be renamed

ffoodd:
I think you need to add an alias or something in orange-navbar's frontmatter, to handle redirections right.

vogloblinsky:
> I think you need to add an alias or something in orange-navbar's frontmatter, to handle redirections right.

What do you mean ?

ffoodd:
See the `nav-tabs.md` doc, which was previously `navs`: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/v5-dev/site/content/docs/5.0/components/navs-tabs.md

Since it's been moved, an alias is set to handle redirection from the old URL to the new one.

vogloblinsky:
> See the `nav-tabs.md` doc, which was previously `navs`: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/v5-dev/site/content/docs/5.0/components/navs-tabs.md
> 
> Since it's been moved, an alias is set to handle redirection from the old URL to the new one.

Done ;)

site/content/docs/4.6/components/navs.md
```diff
@@ -3,6 +3,7 @@ layout: docs
 title: Navs
 description: Documentation and examples for how to use Boosted's included navigation components.
 group: components
+aliases: "/docs/4.6/components/navbar-orange/"
```
julien-deramond, line 6: Maybe I have misunderstood something but for me you'd rather do this:

- Remove this line in this file because we don't want this redirection
- Rename `navbar-orange.md` in `orange-navbar.md`
- Insert this line in `orange-navbar.md`

site/data/sidebar.yml
```diff
@@ -52,11 +52,11 @@
     - title: Navs
     - title: Navbar
     # Boosted mod
```
julien-deramond, line 54: The elements in this `# Boosted mod`/`# end mod` don't respect the alphabetical order that seems to be in place here in 'Components'. Would it be possible to use this PR to also reorder those 5 components in the list at the right place?

site/layouts/partials/footer.html
```diff
@@ -22,7 +22,7 @@ <h2 class="sr-only">Site map & informations</h2>
         <li class="nav-item"><span class="nav-link">Currently v{{ .Site.Params.current_version }}</span></li>
         <li class="nav-item"><a class="nav-link" href="{{ .Site.Params.repo }}">GitHub</a></li>
         <li class="nav-item"><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/">Examples</a></li>
-        <li class="nav-item"><a class="nav-link" href="https://github.com/twbs/bootstrap/tree/{{ .Site.Params.docs_version }}/">Bootstrap</a></li>
+        <li class="nav-item"><a class="nav-link" href="https://github.com/twbs/bootstrap/tree/v{{ .Site.Params.current_version }}/">Bootstrap</a></li>
```
julien-deramond, line 25: The same modification can be done in `masthead.html`

julien-deramond (changes requested):

PR #673, opened 2021-05-25
fix focus-visible import

### LINKED ISSUE #672: focus-within import in boosted 4.6.0 doesn't work outside the bundle
When using `boosted.js` the included polyfill is lighter thant the one in 4.5.6 who works

Maybe linked to the import modification. Tried localy with previous import works well
https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/commit/10d2408e81e35114cf9f57e98d966c1bea1b3ac6#diff-3f689a613b62130589affae189bdfbcff8533286dcf50c288ccf61bd9b59c11cR25

Lausselloic:
fix(focus-visible): fix focus-visible import in `boosted.js` and remove `focus-visible` from peer-dependencies

fix #672


julien-deramond (changes requested):
OK for this modification.

Because we haven't reviewed it quickly, maybe rebase it and check if the bundlewatch new values are still coherent.

Just a question because I don't know much about "jspm", you don't need to remove the `focus-visible` dependency as well in the following section in _package.json_?

```json
"jspm": {
    "registry": "npm",
    "main": "js/boosted",
    "directories": {
      "lib": "dist"
    },
    "shim": {
      "js/boosted": {
        "deps": [
          "focus-visible",
          "jquery",
          "popper.js",
          "tablesorter",
          "swiper"
        ],
        "exports": "$"
      }
    },
    "dependencies": {},
    "peerDependencies": {
      "focus-visible": "^5.2.0",
      "jquery": "1.9.1 - 3",
      "popper.js": "^1.16.1",
      "tablesorter": "^2.31.3",
      "swiper": "^6.5.1"
    }
  }
```

julien-deramond:
I propose to remove as well `focus-visible` from jspm which works like npm from what I have understood. @Lausselloic if you are OK with it, I can merge this PR.

julien-deramond (approved):

PR #693, opened 2021-06-14
docs(style): fix display of nested `<ul><li>`

julien-deramond:
Coming from https://github.com/twbs/bootstrap/pull/34223

The result can be seen here: https://deploy-preview-693--boosted.netlify.app/docs/5.0/content/reboot/ and there: https://deploy-preview-693--boosted.netlify.app/docs/5.0/migration/

### DIFF
--- a/site/assets/scss/_content.scss
+++ b/site/assets/scss/_content.scss
@@ -22,12 +22,12 @@
   > ol li {
     margin-bottom: .25rem;
 
-    // stylelint-disable selector-max-type
-    > ul {
+    // stylelint-disable selector-max-type, selector-max-compound-selectors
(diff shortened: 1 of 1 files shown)


ffoodd:
No need for a PR here, you'll get it back for free when syncing with Bootstrap 👌

julien-deramond:
OK so I'm closing it! I thought the process was to also create a PR for Boosted before getting it back from Boostrap 😊

ffoodd:
You can for a bugfix you'd want to deliver before Bootstrap, but the only moment where that could happen is right after a Bootstrap release and juste before Boostés equivalent release 🙂

PR #719, opened 2021-07-05
Global Header standard with search, shopping basket and avatar icons

Nurovek:
#702


Nurovek:
@julien-deramond Missing:

- align icons on right <768px
- "account icon" (for all examples that have one)

PR #720, opened 2021-07-05
Global header minimised with search, shopping basket and avatar icons

Nurovek:
#703


Nurovek:
@julien-deramond 
Missing :
- align icons on right <768px
- small orange logo (as it should be minimized...) > 768px 
- the bar should be slimmer...

PR #721, opened 2021-07-05
Global header with title, navigation and icons

Nurovek:
#704


Nurovek:
@julien-deramond Missing:
same as #702

Issue #705, opened 2021-06-23
Global Header > Title 2 Lines


julien-deramond:
## References

- **Orange Design System**: [Modules > Global Headers](https://system.design.orange.com/0c1af118d/p/37609b-global-headers/b/366c91) — WEB-GLH-TNI-002
- **v4**: [Navbar Orange > Main navbar](https://boosted.orange.com/docs/4.5/components/navbar-orange/#main-navbar)

PR #722, opened 2021-07-05
Global header on two lines with search, shopping basket and avatar icons

Nurovek:
#705


Nurovek:
@julien-deramond Missing:

- Same as #719
- Spacing between two lines title

Issue #706, opened 2021-06-23
Global header with title and navigation bar


julien-deramond:
Design reference: WEB-GLH-TNI-003

![Capture d’écran 2021-06-21 à 12 59 46](https://user-images.githubusercontent.com/17381666/123070745-4fe87180-d414-11eb-9ee1-32494f6aeeef.png)

PR #723, opened 2021-07-05
Global header with title and navigation bar

Nurovek:
#706


Nurovek:
@julien-deramond Missing:
- Same as #719

PR #724, opened 2021-07-05
Global header with title on 2 lines and navigation bar

Nurovek:
#707


Nurovek:
@julien-deramond Missing: 
- same as #719 
- spacing between 2 lines

Issue #708, opened 2021-06-23
Header Search > Global Header Search Input Box


julien-deramond:
## References

- **Orange Design System**: [Modules > Global Headers](https://system.design.orange.com/0c1af118d/p/37609b-global-headers/b/366c91) — WEB-GLH-NSC-001
- **v4**: [Navbar Orange > Main navbar](https://boosted.orange.com/docs/4.5/components/navbar-orange/#main-navbar)

PR #725, opened 2021-07-05
Global header standard with search box

Nurovek:
#708


Nurovek:
@julien-deramond Missing :

- Black search bar

PR #726, opened 2021-07-05
OBS global header standard

Nurovek:
#709


Nurovek:
@julien-deramond Missing:
- align right for icon <768px
- Authenticated user

PR #736, opened 2021-07-16
fix(#735): separate local fonts from CDN in the starter template

### LINKED ISSUE #735: Docs - invalid preload in starter template
In starter template the preload for font-familly is not valid when using cdn 
https://boosted.orange.com/docs/5.0/getting-started/introduction/#starter-template

Don't know what to mention, let it as a comment as it's a good practice if loading it from local dist folder

image

julien-deramond:
closes #735 

This is a draft to suggest a solution to this problem. Need some technical help to check if this is the good solution.

If this solution is validated, some further development is needed:

- [x] Add `integrity` when fonts are loaded with the CDN
  - ~~Verify that `npm run release-sri` modifies those new integrity values~~
- [x] Handle `rtl.md`
- [ ] Backport this solution into v4

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,87 +1,95 @@
-languageCode:           "en"
-title:                  "Boosted"
-baseURL:                "https://boosted.orange.com"
-enableInlineShortcodes: true
+languageCode:                       "en"
+title:                              "Boosted"
+baseURL:                            "https://boosted.orange.com"
+enableInlineShortcodes:             true
 
 markup:
   goldmark:
     renderer:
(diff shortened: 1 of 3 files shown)


Lausselloic (approved):
sounds good for me Don't know if @ffoodd has a better idea?

ffoodd:
Nope, sounds good!

Regarding SRI, I think there's no need to handle them since files won't change in each release.

Until now, fonts only changed on major release.

Lausselloic (approved):

Issue #733, opened 2021-07-15
Docs: Clicking on social buttons redirects to the main page


julien-deramond:
Clicking on social buttons redirects to the main page in the [v5 documentation](https://boosted.orange.com/docs/5.0/components/buttons/#social-buttons).

This was not the case in the [v4 documentation](https://boosted.orange.com/docs/4.6/components/buttons/#social-buttons) and is not the case in the v5 documentation for other buttons in the same page.

ckkeppler:
> Clicking on social buttons redirects to the main page in the [v5 documentation](https://boosted.orange.com/docs/5.0/components/buttons/#social-buttons).
> 
> This was not the case in the [v4 documentation](https://boosted.orange.com/docs/4.6/components/buttons/#social-buttons) and is not the case in the v5 documentation for other buttons in the same page.

I would like to contribute but cannot seem to gain access to v5 code when cloning.  I am new to contributing but would like to help.  Can you tell me how to access v5 code?

julien-deramond:
Welcome @ckkeppler!
> Can you tell me how to access v5 code? 

Our default branch is still `v4-dev` (version 4). You can checkout the `v5-dev` to access to the version 5 (current version).

PR #740, opened 2021-07-19
Social button fix

ckkeppler:
fixed issue #733

PR #743, opened 2021-07-20
Social button fixed

ckkeppler:
fix(#733): prevent a redirection to the main page after clicking on social buttons


julien-deramond (approved):
Thank you for this PR.

Just few comments for your next contribution:

- When you create the PR to add some code into the version 5, you need to choose as base "v5-dev" and not "v4-dev".
- After some comments, you must keep the same branch to push your modifications and not create another PR (for example this one and https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/740).

PR #789, opened 2021-09-06
docs(version): automatically retrieve latest version in version selector

julien-deramond:
To avoid having 5.0.x instead of 5.1.x in the version selector

Capture d’écran 2021-09-06 à 16 57 51

If you think it is the correct way of doing that, I could propose this fix to Bootstrap too


ffoodd (approved):
Yep, seems appropriate to me 👌

:shipit:

PR #808, opened 2021-09-20
Fix orange navbar 1px border bottom color

isabellechanclou:
Closes #804

PR #809, opened 2021-09-21
fix(ci): 'if' is defined two times in JS GitHub CI workflow

julien-deramond:
@isabellechanclou pointed out in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/808 that there is a problem with the JS GitHub workflow: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/actions/runs/1254019727.

To make it work, also have to merge now those Bootstrap commits:
- https://github.com/twbs/bootstrap/commit/254de4cb42507f965caa337de949df33e3898d43
- https://github.com/twbs/bootstrap/commit/9cf31d9f8e54aa605d0982f372e3c830c598baec


julien-deramond:
Everything's also failing with a fake PR based on `main` branch (https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/810/checks?check_run_id=3659556111) so this is normal here.

PR #819, opened 2021-09-23
chore(deps): regenerate package-lock.json

julien-deramond:
Sometimes, Bootstrap regenerates their `package-lock.json` and we don't.

They told me how they do that:
```
$ git clean -dfx
$ rm -f package-lock.json
$ npm i
```

After that, I compared the results with https://github.com/twbs/bootstrap/blob/main/package-lock.json and most of it seems OK.

For some packages as `autoprefixer` or `tarteaucitronjs` (for example), I've relaunched `npm i <package>` to pin the dependency and avoid a big jump into new versions.

The main issue that I observed is that some packages are downgraded in Boosted `package-lock.json` and not in Bootstrap and I don't know why:
* (`linkinator` || `stylelint`) > `meow` > (`camelcase-keys`, `indent-string`, `minimist-options`)

Note: it will fix https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/818

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,12 +22,12 @@
       }
     },
     "@babel/code-frame": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
-      "integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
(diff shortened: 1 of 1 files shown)


julien-deramond:
OK got it! There's no problem with `meow`.
We don't have the same version inside our `package-lock.json` for `meow` because we use `delay-cli` which has a `meow` dependency in `^4.0.0`.

```
$ grep version node_modules/meow/package.json
"version": "4.0.1"
$ grep version node_modules/linkinator/node_modules/meow/package.json
"version": "9.0.0"
```

Nurovek:
Good for me, well explained, let's merge

PR #830, opened 2021-10-04
fix(cookies): display cookie consent banner on Firefox

julien-deramond:
On https://boosted.orange.com/ in incognito mode you won't see the cookie consent banner at the bottom of the screen.
This PR proposes to fix this issue.
In local mode, you can test it by opening http://localhost:9001 in incognito mode.


julien-deramond:
My bad, it is the normal behavior :)

PR #835, opened 2021-10-05
Chore/merge main@c81c645

julien-deramond:
* Will automatically close https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/827
* Haven't taken into account https://github.com/twbs/bootstrap/commit/78e9b402df84042920d9f2a61e86fc3bb0b4dcea because we haven't exactly the same configuration to handle Dependabot. Maybe we should create an issue to see if our approach is the good one.
  * Shouldn't we use `branches-ignore` too?
  * Bundlewatch is run in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/827 for example while there are `!dependabot/**` rules


julien-deramond:
Closing it. See https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/844

Issue #773, opened 2021-08-23
Vertical alignment of icons within buttons in Firefox


julien-deramond:
Operating system and version: macOS Big Sur 11.4

Considering this page: https://boosted.orange.com/docs/4.6/components/buttons/#examples.

The icons within buttons are slightly to high in Firefox (91.0.1 64 bits macOS).

![Capture d’écran 2021-08-23 à 11 51 33](https://user-images.githubusercontent.com/17381666/130427682-0111381c-f3e5-4aed-b70c-ffd00c692bb2.png)

This issue doesn't appear in Google Chrome (92.0.4515.159 x86_64 macOS).

![Capture d’écran 2021-08-23 à 11 52 22](https://user-images.githubusercontent.com/17381666/130427758-7f64596f-1b42-4b21-87ac-55d5ae5cab4f.png)

PR #851, opened 2021-10-11
Fix (button): vertical alignment of icons within buttons in Firefox (#773)

isabellechanclou:

In Firefox, the button icon is situated 1px too high compared to the same in Chrome or design specifications.
It occurs only in the ::before container. Otherwise, the icon is of the exact same height of 20px.
I propose a go around solution (didn't found out an other one) that seems to solve the problem. Thanks for checking if it is acceptable or not.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -50,12 +50,13 @@
   [class*="svg-"],
   [class*="icon-"] {
     display: inline-block;
-    transform: translateY(1px);
   }
 
   &:not(.btn-icon) {
     [class*="svg-"],
     [class*="icon-"] {
+      position: relative;
+      top: 1px;
       margin-right: $border-width * 3;
     }
   }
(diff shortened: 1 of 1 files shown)


scss/_buttons.scss
```diff
@@ -50,6 +50,7 @@
   [class*="svg-"],
   [class*="icon-"] {
     display: inline-block;
+    transform: translateY(1px);
```
julien-deramond, line 53: If you make this modification here, you will also move down the icon when the button is only composed of an icon. I suppose you rather need to modify the content of the following selector:

```
&:not(.btn-icon) {
    [class*="svg-"],
    [class*="icon-"] {
```

On Firefox, when I enable/disable `transform: translateY(1px);`, I can see the icon slightly also moving on the x-axis. I don't have this behavior with:

```scss
&:not(.btn-icon) {
  [class*="svg-"],
  [class*="icon-"] {
    margin-right: $border-width * 3;
    position: relative;
    top: 1px;
  }
}
```

However, I haven't tried it in all browsers

PR #857, opened 2021-10-18
Fixing modals titles and offcanvas titles

### LINKED ISSUE #786: Font size of modals titles
The `font-size` of modals titles is currently `18px` (see https://boosted.orange.com/docs/5.0/components/modal/) but should be `20px` according to the [Orange Design System](https://system.design.orange.com/0c1af118d/p/16d9f3-modals/b/774d3d/i/48901782).

The design of Offcanvas is for the moment based on the modals one. Its title must also be modified.

This is also true in v4. Do we fix this also in v4?

Take the opportunity to check the entire design for the modals to detect some other bugs 🐛.

louismaximepiton:
Fix #786 Netlify : https://deploy-preview-857--boosted.netlify.app/docs/5.2/components/modal/
### Modal titles are :
- 20px in desktop view
- 18px in mobile view

### Buttons in modal are :
- extended, spaced and reversed in mobile view

### Offcanvas titles are (shall I do variables for offcanvas and modals ?) : 
- 20px in desktop view
- 18px in mobile view

### Issue encountered
- In mobile view in Firefox, some scrolling bar have a higher z-index (between 1000000000 and 10000000000) than the modal (1055) : 

image

- In desktop view on Chrome + Mac OS, in the 1st example of "Scrolling long content", the scroll doesn't word at all:
image

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -170,14 +170,13 @@
 // Title text within header
 .modal-title {
   margin-bottom: 0;
-  font-size: $h5-font-size; // Boosted mod
   line-height: var(--#{$prefix}modal-title-line-height);
-  letter-spacing: $h5-spacing; // Boosted mod
 
   // Boosted mod
-  @include media-breakpoint-up(md) {
-    font-size: $h4-font-size;
-    letter-spacing: $h4-spacing;
+  @include media-breakpoint-down(sm) {
+    &.h4 {
+      font-size: $h5-font-size;
+    }
   }
   // End mod
 }
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -170,7 +170,16 @@
 // Title text within header
 .modal-title {
   margin-bottom: 0;
+  font-size: $h5-font-size; // Boosted mod
   line-height: var(--#{$prefix}modal-title-line-height);
+  letter-spacing: $h5-spacing; // Boosted mod
+
+  // Boosted mod
+  @include media-breakpoint-up(md) {
+    font-size: $h4-font-size;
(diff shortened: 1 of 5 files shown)


isabellechanclou (approved):

site/content/docs/5.1/components/modal.md
```diff
@@ -644,7 +644,7 @@ Our default modal without modifier class constitutes the "medium" size modal.
   <div class="modal-dialog modal-xl">
     <div class="modal-content">
       <div class="modal-header">
-        <h5 class="modal-title h4" id="exampleModalXlLabel">Extra large modal</h5>
```
julien-deramond, line 647: Wouldn't it be possible to keep the `.h4` class which [has already the right values for desktop and tablet](https://boosted.netlify.app/docs/5.1/content/typography/#headings) and just add a media breakpoint rule to only fix the mobile value (with a selector that targets `.h4`)?

scss/_modal.scss
```diff
@@ -189,18 +198,19 @@
   display: flex;
   flex-shrink: 0;
   flex-wrap: wrap;
+  gap: var(--#{$prefix}modal-footer-gap);
```
julien-deramond, line 201: Do you have more info about why this modification was postponed to v6 in Bootstrap? (see `// Todo in v6: replace with gap on parent class`)

louismaximepiton: I made this modification before the bootstrap commit arrived, so I didn't knew about that. It was in https://github.com/twbs/bootstrap/pull/36060 and I can't see any info about this.

PR #861, opened 2021-10-21
No border for focused <select> in Forms > Custom styles

### LINKED ISSUE #715: No border for focused `<select>` in Forms > Custom styles
Operating system: Ubuntu
Browser: Firefox 89.0.2 (64-bit)

There is no border when the `<select>` is focused in https://boosted.orange.com/docs/5.0/forms/validation/#custom-styles.

isabellechanclou:
Closes #715 

**Preview**: https://deploy-preview-861--boosted.netlify.app/docs/5.1/forms/validation/#custom-styles


scss/_variables.scss
```diff
@@ -1001,7 +1001,7 @@ $input-border-radius:                   $border-radius !default;
 $input-border-radius-lg:                $border-radius-lg !default;
 
 $input-focus-bg:                        $input-bg !default;
-$input-focus-border-color:              currentColor !default;
+$input-focus-border-color:              $input-color !default; // Boosted mod: for border to show in Firefox
```
julien-deramond, line 1004: From what I understand, form select has a special treatment in `_form-select.scss`:
```scss
// Remove outline from select box in FF
&:-moz-focusring {
  color: transparent;
  text-shadow: 0 0 0 $form-select-color;
}
```
So when we have here `$input-focus-border-color: currentColor`, it is transparent.
Rather than modifying `$input-focus-border-color` here (which has an impact on all form controls), wouldn't it be better to modify in this same file `$form-select-focus-border-color`?:

```scss
$form-select-focus-border-color:  $input-color !default; // Boosted mod: avoid having a transparent color in Firefox
```

isabellechanclou: Indeed :-)
I don't know if I understood well, but I've tried in `_form-select.scss` file to replace : 
```
  &:focus {
    border-color: $form-select-focus-border-color !important; // stylelint-disable-line declaration-no-important

```
by

```
&:focus {
    border-color: $input-focus-color !important; // stylelint-disable-line declaration-no-important
    outline: none -webkit-focus-ring-color; // Boosted mod: remove outline from select box in Chrome
```
Its works properly BUT as a consequence, the variable `$form-select-focus-border-color` du fichier `_variables.css ` (ligne 1115) is no more used. If we want to keep this variable, we need to change its value. Doing that leads to my first proposal...

Thanks for your opinion on that, or maybe there is something else to do from your comment that I didn't see :-)

julien-deramond (approved):

PR #872, opened 2021-10-27
fix(titlesColor): Fixed utilities' primary color

### LINKED ISSUE #815: The orange colour used for the text samples on white background should be #F16E00 instead of #FF7900
Before opening:

- [Search for duplicate or closed issues](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [Validate](https://html5.validator.nu/) any HTML to avoid common problems
- Read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

Bug reports must include:

- Operating system and version (Windows, macOS, Android, iOS)
- Browser and version (Chrome, Firefox, Safari, Microsoft Edge, Opera, Android Browser)
- A [Reduced test case](https://css-tricks.com/reduced-test-cases/) or suggested fix using [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/)

louismaximepiton:
Fix #815: Color of text utilities from #ff7900 to #f16e00 on white backgrounds

Netlify : https://deploy-preview-872--boosted.netlify.app/docs/5.2/utilities/colors/

### DIFF
--- a/site/content/docs/5.2/utilities/colors.md
+++ b/site/content/docs/5.2/utilities/colors.md
@@ -54,6 +54,8 @@ Consider our default `.text-primary` utility.
 
 We use an RGB version of our `$accessible-orange` (with the value of `241, 110, 0`) Sass variable as a CSS variable and attached a second CSS variable, `--bs-text-opacity`, for the alpha transparency (with a default value `1` thanks to a local CSS variable). That means anytime you use `.text-primary` now, your computed `color` value is `rgba(241, 110, 0, 1)`. The local CSS variable inside each `.text-*` class avoids inheritance issues so nested instances of the utilities don't automatically have a modified alpha transparency.
 
+When used in a dark variant context, `--bs-primary-text-rgb` will use the value of `$brand-orange` (with the value of `255, 121, 0`).
+
 ### Example
 
 To change that opacity, override `--bs-text-opacity` via custom styles or inline styles.


scss/_root.scss
```diff
@@ -18,7 +18,11 @@
   }
 
   @each $color, $value in $theme-colors-rgb {
-    --#{$variable-prefix}#{$color}-rgb: #{$value};
+    @if ($color == "primary") {
+      --#{$variable-prefix}#{$color}-rgb: #{to-rgb($accessible-orange)};
```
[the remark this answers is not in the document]
isabellechanclou: Maybe I'm wrong, but it seems that this issue is not relevant anymore. The concerned page doesn't exist anymore (content moved elsewhere and/or deleted). So, I suggest to close/abandon this PR.

julien-deramond:
@louismaximepiton I think that this PR must be updated.

julien-deramond:
In [Helpers > Colored links](https://deploy-preview-872--boosted.netlify.app/docs/5.2/helpers/colored-links/) we have orange on black bg. So the orange should be "Brand orange" instead of "Accessible orange" IMO.

julien-deramond:
At some point (maybe not in this PR) we could maybe say in [Utilities > Color](https://deploy-preview-872--boosted.netlify.app/docs/5.2/utilities/colors/#colors) that `.text-primary` should only be used in combination with large texts (source: https://system.design.orange.com/0c1af118d/p/7059a5-colour/b/86fdf5)

julien-deramond:
While I'm checking for `.text-primary`, maybe we can drop `.bg-transparent` from:
https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/64f0eb4b66e4d2af9a986b911ef1c05e8d1c8eb5/site/layouts/partials/home/masthead.html#L9

scss/_utilities.scss
```diff
@@ -542,6 +542,7 @@ $utilities: map-merge(
           "black-50": rgba($black, .5), // deprecated
           "white-50": rgba($white, .5), // deprecated
           "reset": inherit,
+          "primary": var(--#{$prefix}link-hover-color),
```
julien-deramond, line 545: Isn't it better to use directly `$accessible-orange`?

louismaximepiton: Not sure about that since we want it to be `$brand-orange` on dark backgrounds, right ?

julien-deramond: Ah yes I was too focused on the weird usage of "link-hover-color". So OK to use a CSS var to allow the switch of orange color but maybe we could create a specific CSS var rather than using the one for the hovered links.

louismaximepiton: I'm not sure about the name I gave, it could be discussed as well

scss/_root.scss
```diff
@@ -21,6 +21,7 @@
     --#{$prefix}#{$color}-rgb: #{$value};
   }
 
+  --#{$prefix}accessible-orange: #{$accessible-orange}; // Boosted mod
```
julien-deramond, line 24: In Bootstrap dark mode PR, `--bs-{themeColor}-text` is introduced:

```css
--bs-primary-text: #0a58ca;
--bs-secondary-text: #6c757d;
--bs-success-text: #146c43;
--bs-info-text: #087990;
--bs-warning-text: #997404;
--bs-danger-text: #b02a37;
--bs-light-text: #6c757d;
--bs-dark-text: #495057;
```

Not 100% sure yet that it's going to be merged like this but it seems likely. I have the feeling that having `--bs-primary-text` would be more appropriate here.

In light mode `--bs-primary-text` would be `$accessible-orange` and in dark mode (dark variant for Boosted), it would be `$brand-orange`. And so, everywhere in the code we use something without the word "accessible" and not linked to a specific color; we can't do that `--#{$prefix}accessible-orange: #{$brand-orange};` IMHO. It would be like saying `--bs-danger`: #($warning)`.

What do you think?

louismaximepiton: Agree on this, changed on last commit ! Forgot about checking the Bootstrap one 🤦

julien-deramond (changes requested):
I've spotted an issue with our spinners that use `text-primary` to define the color so our orange spinners now use `#f16e00` instead of `#ff7900` ([DSM guidelines reference](https://system.design.orange.com/0c1af118d/p/45093c-progress-indicators/b/2875f1/i/66611042))

julien-deramond:
* 🟢 Text switch between oranges looks good
* 🟢 Icon switch between oranges looks good based on https://system.design.orange.com/0c1af118d/p/361038-iconography/t/56c4f9
* 🔴 Spinner orange color on white background is not the same as the one for icons (see https://system.design.orange.com/0c1af118d/p/45093c-progress-indicators/b/2875f1/i/66611042) so let's wait for an explanation from the designers.

louismaximepiton:
Returns from the spec review: the spinners should be `$accessible-orange` on white and `$brand-orange` on black.

julien-deramond (approved):

julien-deramond (changes requested):
Just before merging, discovered that https://deploy-preview-872--boosted.netlify.app/docs/5.2/utilities/colors/#opacity doesn't work anymore

scss/_root.scss
```diff
@@ -21,6 +21,7 @@
     --#{$prefix}#{$color}-rgb: #{$value};
   }
 
+  --#{$prefix}primary-text: #{to-rgb($accessible-orange)}; // Boosted mod
```
julien-deramond, line 24: So it should be named `--#{$prefix}primary-text-rgb`

PR #878, opened 2021-11-02
Chore/merge main@05f1b15

julien-deramond:


site/content/docs/5.1/components/navbar.md
```diff
@@ -235,12 +235,6 @@ Immediate child elements of `.navbar` use flex layout and will default to `justi
 </nav>
 {{< /example >}}
 
-{{< callout warning >}}
```
julien-deramond, line 238: See conversation in https://github.com/twbs/bootstrap/pull/35223

scss/mixins/_container.scss
```diff
@@ -1,9 +1,11 @@
 // Container mixins
 
 @mixin make-container($gutter: $container-padding-x) {
+  --#{$variable-prefix}gutter-x: #{$gutter};
```
julien-deramond, line 4: Double check if it isn't a regression

louismaximepiton: In my opinion based on the tests I made, it isn't a regression. But as it initializes the variable, do we need to keep the second part of expression (#{$gutter}) line 7 and line 8 ?

julien-deramond: Got the same question, I would have done `var(--#{$variable-prefix}gutter-x);` instead of  `var(--#{$variable-prefix}gutter-x, #{$gutter});` but I kept the Bootstrap version just to avoid any regressions

louismaximepiton (approved):

PR #881, opened 2021-11-02
Dropdown tab works only one time in Boostwatch example

### LINKED ISSUE #822: Dropdown tab works only one time in Boostwatch example
- Operating system and version: macOS Big Sur 11.5.2
- Browser and version: Firefox 92.0 (64 bits)

Manipulation done in https://boosted.orange.com/docs/4.6/boostwatch/#tab:

The first time we choose "Action" or "Another action" within the dropdown, it works!; the tabs content is modified.
When you come back to this dropdown and click on it, both actions seem active and clicking on it does nothing.

julien-deramond:
Closes #822


MewenLeHo (approved):
OK

PR #910, opened 2021-11-15
Main kld global headers

Nurovek:
Mostly documentation.


julien-deramond (approved):
Haven't reviewed it but just checked there is new content and no regression compared to the main branch of this feature.

PR #921, opened 2021-11-19
Chore/merge main@9c01633

julien-deramond:
I'm not sure if I need to move the following from `_variables.css` to the new `_maps.scss`:

```scss
// Boosted mod
$background-colors: map-merge(
  $theme-colors,
  (
    "success": $supporting-green,
    "info":    $supporting-blue,
    "warning": $supporting-yellow,
    "danger":  $danger
  )
) !default;
// End mod
```

**Note:** Some errors can be observed while testing this part but are fixed via https://github.com/twbs/bootstrap/pull/35373


MewenLeHo:
> I'm not sure if I need to move the following from `_variables.css` to the new `_maps.scss`:
> 
> ```css-scss
> // Boosted mod
> $background-colors: map-merge(
>   $theme-colors,
>   (
>     "success": $supporting-green,
>     "info":    $supporting-blue,
>     "warning": $supporting-yellow,
>     "danger":  $danger
>   )
> ) !default;
> // End mod
> ```
> 
> **Note:** Some errors can be observed while testing this part but are fixed via [twbs/bootstrap#35373](https://github.com/twbs/bootstrap/pull/35373)


According to the doc: 'Your custom Boosted CSS builds should now look something like this with a separate maps import'.
IMO this part should be move.

julien-deramond:
At first, I thought it wasn't a problem since Orange developers won't override `$theme-colors`. But maybe OB1 does something like this. In this case, yes, it would be better to move it into the new file and maybe `$alert-colors: map-remove($theme-colors, "primary", "secondary", "light", "dark") !default;` as well since it depends on `$theme-colors`.

julien-deramond:
@MewenLeHo Fixed via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/921/commits/77dd05232cb441f3dcd5820286032d14dea43519 and https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/921/commits/9c691995c06ae56a8d33f2b62518aa5a5250aa3b.

MewenLeHo (approved):
👍

PR #933, opened 2021-11-25
Beta - Display danger callouts to highlight non Brand compliant components

julien-deramond:
This PR is a discussion paper regarding the way we handle what's in the framework (inherited from Bootstrap) and what's forbidden by the Orange Brand.

In v4, we removed some components, some versions of a component, some features, etc. but still, there were some elements still not allowed by the Orange Brand.

In this PR, I tried to add a message in the top of two components instead of removing partly or entirely their content or features.

IMO, those technical information must be kept in the documentation even if they're against the design specifications:
* If a developer wants to break some rules, it's still possible with CSS or by reintegrating some code
* Third-party apps or tools like OB1 could use some features not allowed by Orange Brand and we're only breaking them if we remove some part of our library

What's in this PR
* [Danger callout for badges that are normally totally forbidden](https://deploy-preview-933--boosted.netlify.app/docs/5.1/components/badge/)
* [Danger callout for Navbar](https://deploy-preview-933--boosted.netlify.app/docs/5.1/components/navbar/) that contains a lot of technical examples (I've reintegrated all examples from Bootstrap) that are not compliant and redirect to the future Orange Navbar

### DIFF
--- a/site/content/docs/5.1/components/badge.md
+++ b/site/content/docs/5.1/components/badge.md
@@ -6,6 +6,17 @@ group: components
 toc: true
 ---
 
+<!--Boosted mod-->
+{{< callout danger >}}
+### Not compliant with Orange brand
+
+Badges come from Bootstrap and are not compliant with Orange brand except the [Positioned](#positioned) version used in headers.
+
+They are kept for compatibility with Bootstrap and third-party libraries linked to Bootstrap.
+
+{{< /callout >}}
+<!--End mod-->
+
 ## Examples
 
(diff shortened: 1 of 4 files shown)


Lausselloic:
I agree with the approach. From my experience what's difficult is to keep warning up-to-date in the time, when new components are produced in Bootstrap need to check with brand what's the status.

Now there's a design system, maybe it could be usefull to link to it when saying "don't use this default component => check out the brand guidelines"

By the way it's better to keep all default component from bootstrap, at least to allow using boosted with bootstrap templates availbale with major CMS

PR #934, opened 2021-11-25
Layouts: add a top margin to the doc footer

julien-deramond:
Add a top margin to the doc footer to avoid this rendering when the side menu is non collapsed and we scroll:

Screenshot from 2021-11-25 08-06-11

New rendering:
Screenshot from 2021-11-25 08-10-46

### DIFF
--- a/site/layouts/_default/docs.html
+++ b/site/layouts/_default/docs.html
@@ -2,7 +2,7 @@
   {{ partial "docs-subnav" . }}
 
   <div class="container-xxl bd-layout">
-    <aside class="bd-sidebar border-end border-light border-1">
+    <aside class="bd-sidebar">
       {{ partial "docs-sidebar" . }}
     </aside>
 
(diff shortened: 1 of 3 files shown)


MewenLeHo (commented):
It's good but now the right border of the aside stops before the top of the footer and it looks a bit "strange". Could we try instead a margin-bottom on #bd-docs-nav?

julien-deramond:
Good point yeah. Maybe me could remove this border. I think it's inherited from Bootstrap 4; Bootstrap 5 doesn't have this border anymore.

MewenLeHo:
> Good point yeah. Maybe me could remove this border. I think it's inherited from Bootstrap 4; Bootstrap 5 doesn't have this border anymore.

Good idea, it will also look more modern :+1:

MewenLeHo (approved):
👌

PR #727, opened 2021-07-05
Global Headers (all)

Nurovek:
#702 #703 #704 #705 #706 #707 #708 #709 #710 

This pull request contains every global headers (Global Headers, OBS headers)

Preview:
* [Docs > Orange Navbar](https://deploy-preview-727--boosted.netlify.app/docs/5.1/components/orange-navbar/)
* [Design guidelines > Global Headers](https://deploy-preview-727--boosted.netlify.app/docs/5.1/guidelines/global-headers/)
* [Example > Navbars](https://deploy-preview-727--boosted.netlify.app/docs/5.1/examples/navbars/)

---

### Development

- [ ] Should match specs (eg. either the Web UI Kit or any pattern from the WAI — or both…)
- [ ] Docs added:
  - including the "Sass" part using `scss-docs` shortcode
  - in /about/overview/#custom-components if it is a new Orange custom component
  - in /getting-started/introduction/#components if it is a new Orange custom component that requires JavaScript (and Popper)
  - in /customize/overview#csps-and-embedded-svgs if it is a new Orange custom component that includes embedded SVGs in our CSS
  - in /forms/validation/?#supported-elements if it is a new Orange custom component that is a form control
  - in /forms/overview/ if it is a new Orange custom component that is a form control
- [ ] Check (and fix) RTL version
- [ ] Run linters
- [ ] Run compilers
- [ ] Tests added for JS-side
- [ ] Run tests
- [ ] Manually run BrowserStack test
- [ ] Manually run Percy test
- [ ] Cross-browser test:
  - Firefox ESR
  - Latest Edge, Chrome, Firefox, Safari
  - iOS Safari
  - Chrome & Firefox on Android
- [ ] Clean up the branch using `rebase -i`
- [ ] Commited with `feat(…): …` message
- [ ] Mention it in Migration Guide (if `back-from-v4`): renamed variables, changes in markup requirement,  etc.

### Reviews

- [ ] Code review
- [ ] A11y review
- [ ] Design review

PR #936, opened 2021-11-26
Global headers, some solutions

louismaximepiton:
Each commit will represent a proposal for a fix

Some can be replaced by utilities in html


Nurovek:
I manually brought what was changed into #727, closing this one

PR #939, opened 2021-11-29
Chore/merge main@5290080

julien-deramond:


louismaximepiton (approved):
I found several bugs, but I also found them on Bootstrap's versions. 

- Tooltip : (Example on [copy button](https://deploy-preview-32692--twbs-bootstrap.netlify.app/docs/5.1/components/tooltips/#example-enable-tooltips-everywhere)) When the copy button is clicked, the tooltip miss some space (also its arrow) and miss its animation. image
- Popover : (Example on [Popover](https://deploy-preview-34519--twbs-bootstrap.netlify.app/docs/5.1/components/popovers/#example)) When click on any button with popover, looks like the arrow is missplaced on the popover image
- Not related to those commits but rtl modals close button is not well aligned (Maybe due to `.modal-header .btn-close {margin}` l.125 in modals.scss).

julien-deramond:
Good catches for the 🐛s you've spotted!
For the first 2, you could:

- Check on Bootstrap main branch if they are still here (there have been some new commits since your review—don't seem related to that but it's better to check anyway)
- Create an issue on the Bootstrap repo for each different bug with the commits references. I've only checked the one regarding the popovers and I confirm it's from https://github.com/twbs/bootstrap/commit/94a596fbcb1011ba990da2078ba7e20b39dba2d9—maybe this change but not 100% sure:

```diff
element: `.${this.constructor.NAME}-arrow`
element: SELECTOR_TOOLTIP_ARROW
```
- If you want to take some time to try to fix them, go ahead!

For the third one, if it is directly in Bootstrap, you can do the same or create it in Boosted if it concerns only us.

PR #958, opened 2021-12-01
Chore/merge main@2fe90d8

julien-deramond:


js/tests/browsers.js
```diff
@@ -51,13 +51,6 @@ const browsers = {
     browser: 'Firefox',
     browser_version: 'latest'
   },
-  EsrWin10: {
```
julien-deramond, line 54: This one was only in Boosted. I chose to remove it to have the same file than Bootstrap.

Lausselloic: But ESR is the one used inside Orange 🤔

louismaximepiton (approved):
Many little corrections to add such as `js/tests/tab` line 117 -> `</button>` tag instead of `</a>`, I'll provide patches to Bootstrap directly.

js/tests/browsers.js
julien-deramond: OK: verified in the history of both Boosted and Bootstrap. Thanks for the info, I've put it back in this PR surrounded by a `Boosted mod` comment to highlight it via 45207e3c0b0bae8c76689c911efdff836e5eb8dc

PR #960, opened 2021-12-06
feat(guidelines): drop design guidelines content and replace it with a simpler page

### LINKED ISSUE #888: Drop "Design guidelines" content from documentation
Drop the content of ["Design guidelines"](https://boosted.orange.com/docs/5.1/guidelines) which has become a poor clone of the [Orange Design System](https://system.design.orange.com).

Replace it with a description and a link to [Orange Design System](https://system.design.orange.com) — a mockup on Zeplin has been proposed by @CyriaqueBillard 

Elements that are only in this part of the documentation, if they are relevant, should be moved / copied somewhere else in the documentation; inside the components documentation or in a specific example page.

julien-deramond:
Closes #888 

Previews:
- [Main page](https://deploy-preview-960--boosted.netlify.app)
- [New Orange Design System for web page](https://deploy-preview-960--boosted.netlify.app/docs/5.1/guidelines/)


julien-deramond:
@isabellechanclou @louismaximepiton Some thoughts about the cards (not to do here but with a dedicated issue to create):
- Create specific classes for the developers in order to use the 3 authorized sizes: Large (460px), Medium (300px) and Small (220px). Something like `.card-sm` and `.card-lg` (medium size being the default size like the buttons).
- Let in the documentation all the parts regarding the ways of dealing with the sizes but precise that the recommended sizes are the larger, medium and small
- Display the 3 sizes in the examples
- Only use the medium size in the examples for all the examples
- Check the rendering and compare it to [the design specs](https://system.design.orange.com/0c1af118d/p/82eaf1-cards/b/69655e/i/66611055). I've spotted some inconsistencies: height of the area when there is only a title for example

julien-deramond:
We should check _.github/workflows.percy.yml_ and how it works. I can see `flags: "--snapshot-files ./docs/5.1/guidelines/**/*.html,./docs/5.1/examples/cheatsheet/index.html,./docs/5.1/examples/cheatsheet-rtl/index.html"`. Since we won't have the components anymore in "/guidelines" and that the 2 cheatsheet examples are hidden.

PR #972, opened 2021-12-14
Kld global header fixes

Nurovek:


Nurovek:
Added a orange-navbar.js file, where there will be js for orange navbars 
Next : 
- add the js file to the index (import/...)
- add documentation for `.scroll-minimized`

PR #987, opened 2021-12-16
Handle `.container-fluid` to follow brand

louismaximepiton:
Previews: 
- https://deploy-preview-987--boosted.netlify.app/docs/5.1/examples/grid-tests/
- https://deploy-preview-987--boosted.netlify.app/docs/5.1/examples/grid/

Before merge : 
- [x] Need to remove grid-tests related changes


louismaximepiton:
Another solution could be to set max-width to 100% and to set margin to what the design want (directly in `.container-fluid`). Do you have an idea of what is best here ?

isabellechanclou:
Code review OK for me.
To do to finalise and before merging:
-[] approval of the changes by someone else
-[] modification of bundlewatch values to make it succeed

site/content/docs/5.1/layout/containers.md
```diff
@@ -151,22 +154,19 @@ Use `.container-fluid` for a full width container, spanning the entire width of
...
-```scss
-// Source mixin
-@mixin make-container($padding-x: $container-padding-x) {
-  width: 100%;
-  padding-right: $padding-x;
-  padding-left: $padding-x;
-  margin-right: auto;
-  margin-left: auto;
-}
+{{< scss-docs name="container-mixins" file="scss/mixins/_container.scss" >}}
```
julien-deramond, line 167: If their `@mixin make-container` example is supposed to be the same the code than in `scss/mixins/_container.scss`, then you could propose your modification to Bootstrap. 

Because what's displayed here is different from https://github.com/twbs/bootstrap/blob/main/scss/mixins/_container.scss.

julien-deramond (approved):
From what I've seen, we can try this solution in order to be able to finish the Global Headers, Footers and Local Headers, and have the correct responsive behavior asked by the designers.
The use of `.container-xxl` everywhere seems to be a good solution for Orange websites to easily have this behavior.
Let's merge and see in few commits if it works while we provide examples for the developers in Boosted!

PR #989, opened 2021-12-20
chore/merge-main@4216254

julien-deramond:


julien-deramond:
In order to apply https://github.com/twbs/bootstrap/commit/cb46ad633c3ca0597e7205ce1957d8dbc6484db6 in Boosted, I've installed [svg/svgo](https://github.com/svg/svgo) and run for each SVG defined in Sass files:
```
$ svgo -s "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 14 7'><path d='M7 7L0 0h14L7 7z'/></svg>" | sed "s/\"/'/g"
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 14 7'><path d='M7 7 0 0h14L7 7z'/></svg>
```
It probably should land into our Wiki.

scss/_alert.scss
```diff
@@ -3,13 +3,24 @@
 //
 
 .alert {
+  // scss-docs-start alert-css-vars
+  --#{$variable-prefix}alert-bg: transparent;
+  --#{$variable-prefix}alert-padding: #{$alert-padding-y $alert-padding-x};
+  --#{$variable-prefix}alert-margin-bottom: #{$alert-margin-bottom};
+  --#{$variable-prefix}alert-color: #{$alert-padding-y $alert-padding-x};
```
louismaximepiton, line 10: That's a great color ! 😂 + not used

scss/_alert.scss
```diff
@@ -3,13 +3,24 @@
...
+  --#{$variable-prefix}alert-border-color: transparent;
```
louismaximepiton, line 11: Shouldn't we use this on l.69 in this file ? Instead of ```border-color: $value```, having ```--#{$variable-prefix}alert-border-color: $value``` should be more flexible. However, I'm not sure if the flexible argument is a real one here.

scss/mixins/_alert.scss
```diff
@@ -1,8 +1,12 @@
 // scss-docs-start alert-variant-mixin
 @mixin alert-variant($background, $border, $color) {
-  color: $color;
-  @include gradient-bg($background);
-  border-color: $border;
+  --#{$variable-prefix}alert-color: #{$color};
```
louismaximepiton, line 3: Still not used right ?

scss/_alert.scss
julien-deramond: Yep :) It is tackled directly in Bootstrap via https://github.com/twbs/bootstrap/commit/a2c056e1f6d5925a5e6fef1f5d709f07a2a535d1

site/content/docs/5.1/helpers/position.md
```diff
@@ -40,3 +40,22 @@ Responsive variations also exist for `.sticky-top` utility.
...
+
+## Responsive sticky bottom
+
+Responsive variations also exist for `.sticky-bottom` utility.
+
+```html
+<div class="sticky-sm-bottom">Stick to the bottom on viewports sized SM (small) or wider</div>
+<div class="sticky-md-bottom">Stick to the bottom on viewports sized MD (medium) or wider</div>
+<div class="sticky-lg-bottom">Stick to the bottom on viewports sized LG (large) or wider</div>
+<div class="sticky-xl-bottom">Stick to the bottom on viewports sized XL (extra-large) or wider</div>
```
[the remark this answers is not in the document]
julien-deramond: I think you're right. IMO both `sticky-xxl-bottom` and `sticky-xxl-top` are missing. I let you create a PR for Bootstrap to fix it and it'll come back with the merges.

scss/_alert.scss
julien-deramond: Not sure to understand what you propose here :innocent:

julien-deramond:
I've taken into account your 2 comments about the mixin by reintegrating the next Bootstrap commit (https://github.com/twbs/bootstrap/commit/a2c056e1f6d5925a5e6fef1f5d709f07a2a535d1) via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/989/commits/8e7ff4c989260ba0f11082547ed1d2aa018a438c.

scss/mixins/_alert.scss
```diff
@@ -1,11 +1,19 @@
...
-  border-color: $border;
+@mixin alert-variant($background, $border, $color, $background-image) { // Boosted mod
+  --#{$variable-prefix}alert-color: #{$color};
+  --#{$variable-prefix}alert-bg: #{$background};
+  --#{$variable-prefix}alert-border: #{$border};
 
-  .alert-link {
-    color: shade-color($color, 20%);
+  // Boosted mod
+  border: var(--#{$variable-prefix}alert-border);
```
louismaximepiton, line 8: You can remove this line as it is CSS variables

louismaximepiton (approved):
Just one line to remove ? I can't decide whether the background-image should or should not be in the mixin, so I'd let as-is.

julien-deramond:
@Lausselloic We struggle to decide if what I've modified for the alerts to be closer to Bootstrap has done is OK or not. Any opinion?

scss/_variables.scss
```diff
@@ -1568,6 +1568,7 @@ $modal-scale-transform:             scale(1.02) !default;
 // Define alert colors, border radius, and padding.
 
 // scss-docs-start alert-variables
+$alert-color:                       $black !default; // Boosted mod
```
Lausselloic, line 1571: why not use inherit as a value here? and then use this variable inside alert.scss?

julien-deramond: Yes indeed, that's better

louismaximepiton (approved):

PR #1004, opened 2021-12-28
feat(githubWorkflows): Adding Percy and Browserstack jobs

louismaximepiton:


.github/workflows/percy.yml
```diff
@@ -1,23 +1,15 @@
...
       - 'scss/**'
-      - 'site/content/docs/**/components'
-      - 'site/content/docs/**/forms'
+      - 'site/content/docs/**/components/**'
+      - 'site/content/docs/**/forms/**'
     branches:
       - main
-      - "!dependabot/**"
+      - v4-dev
   pull_request:
```
julien-deramond, line 12: The goal of this PR is to run both jobs when there is a push on _main_ or _v4-dev_ or on demand manually. It is needed to remove this line isn't it?

.github/workflows/browserstack.yml
```diff
@@ -21,41 +15,41 @@ env:
...
   browserstack:
     runs-on: ubuntu-latest
-    if: github.repository == 'Orange-OpenSource/Orange-Boosted-Bootstrap' && (!contains(github.event.commits[0].message, '[ci skip]') && !contains(github.event.commits[0].message, '[skip ci]'))
     timeout-minutes: 30
 
     steps:
       - name: Bail early
-        if: github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'skip:ci') || contains(github.event.pull_request.labels.*.name, 'skip:browserstack'))
         run: exit 0
+        if: github.actor == 'dependabot[bot]'
```
julien-deramond, line 23: If the job can only be launched by pushing _main_ or _v4-dev_ or manually, I don't understand why we should keep this line about Dependabot which can't do those 2 actions.

.github/workflows/percy.yml
```diff
@@ -30,18 +22,18 @@ jobs:
 
     steps:
       - name: Bail early
-        if: github.event_name == 'pull_request' && ((contains(github.event.pull_request.labels.*.name, 'skip:ci') || contains(github.event.pull_request.labels.*.name, 'skip:percy')) || (github.event.pull_request.draft == true))
         run: exit 0
+        if: github.event_name == 'pull_request' || github.actor == 'dependabot[bot]'
```
julien-deramond, line 26: All those lines for Percy are not coherent with Browsertack

.github/workflows/percy.yml
```diff
@@ -30,18 +22,18 @@ jobs:
...
 
       - name: Clone repository
         uses: actions/checkout@v2
-        if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.pull_request.draft == false && !contains(github.event.pull_request.labels.*.name, 'skip:ci') && !contains(github.event.pull_request.labels.*.name, 'skip:percy'))
```
julien-deramond, line 38: We can remove `skip:percy` and `skip:browserstack` from everywhere if we can only launch those jobs manually or automatically on 2 branches.

julien-deramond (changes requested):
Depending on the actions after my comments, we'll probably need to do after that the PR will be merged:
- [x] Remove `skip:percy` and `skip:browserstack` labels
- [x] Explain in the DoDs that Browserstack needs to be launched manually at the end of the PR just before merging (and Percy as well when necessary).

julien-deramond (approved):
Sound good to me! We're going to try this version.

PR #1005, opened 2021-12-29
Drop `delay-cli` to avoid CVE-2021-33623 and use `sirv-cli`

julien-deramond:
Drop `delay-cli` to avoid CVE-2021-33623 vulnerability.

Note: could be fixed by https://github.com/sindresorhus/delay-cli/pull/6 as well.

The use of `sirv-cli` comes from the WIP Bootstrap branch https://github.com/twbs/bootstrap/blob/main-xmr-pa11y-ci.


julien-deramond:
Put it back in Draft mode cause I gotta check again if replacing `npm run dist` + `npm run release-sri` + `npm run docs-build` with `npm run docs` does the job.

julien-deramond:
1. Created https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1005/commits/3db32975fc95c52e23ad28e2cb722409e61032c1 that should be caught by pa11y-ci.
As you mentioned @Lausselloic, it is not caught by pa11y-ci because the CSS is not regenerated (see https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/runs/4775193184?check_suite_focus=true).
2. Put back the previous configuration via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1005/commits/005b7b92991bed1674fe4577fb4a29b10150ed82: pa11y-ci [catches the new error introduced in the Sass file](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/runs/4775236047?check_suite_focus=true). So, good catch :) I'll let the previous config.

PR #1014, opened 2022-01-04
chore/merge main@cebb843

julien-deramond:
Note: the last commit (https://github.com/twbs/bootstrap/commit/cebb8436c7f771424ae9978484379608b250d258) from Bootstrap doesn't impact Boosted since we don't want to update our image


louismaximepiton (approved):
`Notice.txt` > 2019 for Orange SA, does it need a change ? 

Whatever, the remaining sounds good to me

PR #1025, opened 2022-01-07
Dark accordions

louismaximepiton:
### Description

Adding a `.accordion-dark` for a dark variant of accordions.

### Motivation & Context

Need for orange footers

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would change existing functionality)

### Checklist

### Development

- [x] Should match specs (eg. either the Web UI Kit or any pattern from the WAI — or both…)
- [x] Docs added:
  - including the "Sass" part using `scss-docs` shortcode
  - in /about/overview/#custom-components if it is a new Orange custom component
  - in /getting-started/introduction/#components if it is a new Orange custom component that requires JavaScript (and Popper)
  - in /customize/overview#csps-and-embedded-svgs if it is a new Orange custom component that includes embedded SVGs in our CSS
  - in /forms/validation/?#supported-elements if it is a new Orange custom component that is a form control
  - in /forms/overview/ if it is a new Orange custom component that is a form control
- [x] Check (and fix) RTL version
- [x] Run linters
- [x] Run compilers
- [x] Tests added for JS-side
- [x] Run tests
- [ ] Manually run BrowserStack test
- [ ] Manually run Percy test
- [ ] Cross-browser test:
  - Firefox ESR
  - Latest Edge, Chrome, Firefox, Safari
  - iOS Safari
  - Chrome & Firefox on Android
- [ ] Clean up the branch using `rebase -i`
- [ ] Commited with `feat(…): …` message
- [ ] Mention it in Migration Guide (if `back-from-v4`): renamed variables, changes in markup requirement,  etc.

### Reviews

- [ ] Code review
- [ ] A11y review
- [ ] Design review

### Related issues

No issue

### [Live preview](https://deploy-preview-{your pr number}--boosted.netlify.app/docs/5.1/components/accordion/#dark-variant)

### DIFF
--- a/scss/_accordion.scss
+++ b/scss/_accordion.scss
@@ -115,3 +115,39 @@
     letter-spacing: $accordion-button-letter-spacing-lg;
   }
 }
+
+// Boosted mod
+.accordion-dark {
+  .accordion-button {
+    color: $accordion-dark-button-color;
+    background-color: $accordion-dark-button-bg;
+
+    &:not(.collapsed) {
+      color: $accordion-dark-button-active-color;
+      background-color: $accordion-dark-button-active-bg;
+    }
+
+    &:hover {
+      color: $accordion-dark-button-active-color;
+    }
+
+    &:focus {
+      color: $accordion-dark-button-active-color;
+    }
+  }
+
+  .accordion-header {
+    border: $accordion-border-width solid $accordion-dark-border-color;
+    border-width: $accordion-border-width 0 0;
+  }
+
+  .accordion-item {
+    color: color-contrast($accordion-dark-bg);
+    background-color: $accordion-dark-bg;
+
+    &:last-of-type {
+      border-bottom: $accordion-border-width solid $accordion-dark-border-color;
+    }
+  }
+}
+// End mod
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1387,7 +1387,7 @@ $card-footer-color:                 $gray-700 !default; // Boosted mod
 
 // scss-docs-start accordion-variables
 $accordion-padding-y:                     $spacer * .5 !default;
-$accordion-padding-x:                     $spacer * .5 !default;
+$accordion-padding-x:                     0 !default;
 $accordion-color:                         $body-color !default;
 $accordion-bg:                            $body-bg !default;
 $accordion-border-width:                  $border-width * .5 !default;
@@ -1405,6 +1405,14 @@ $accordion-button-bg:                     $accordion-bg !default;
 $accordion-transition:                    $btn-transition, border-radius .15s ease !default;
 $accordion-button-active-bg:              null !default;
 $accordion-button-active-color:           $accessible-orange !default;
+
+$accordion-dark-color:                    $white !default;
+$accordion-dark-bg:                       $black !default;
+$accordion-dark-border-color:             $gray-700 !default;
+$accordion-dark-button-color:             $accordion-dark-color !default;
+$accordion-dark-button-bg:                $accordion-dark-bg !default;
+$accordion-dark-button-active-bg:         null !default;
+$accordion-dark-button-active-color:      $brand-orange !default;
 
 // Boosted mod: no $accordion-button-focus-*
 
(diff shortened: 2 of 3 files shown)


scss/_accordion.scss
```diff
@@ -115,3 +115,39 @@
     letter-spacing: $accordion-button-letter-spacing-lg;
   }
 }
+
+// Boosted mod
+.accordion-dark {
```
Lausselloic, line 120: like button are button-inverse maybe use accordion-inverse instead of dark?

scss/_variables.scss
```diff
@@ -1387,7 +1387,7 @@ $card-footer-color:                 $gray-700 !default; // Boosted mod
 
 // scss-docs-start accordion-variables
 $accordion-padding-y:                     $spacer * .5 !default;
-$accordion-padding-x:                     $spacer * .5 !default;
+$accordion-padding-x:                     0 !default;
```
Lausselloic, line 1390: there's no change on visual accordion?

louismaximepiton: It avoids to have a negative margin that didn't fit well with dark-accordions. IMO, it doesn't change anything for accordions that are on the same bg-color and correct some visual issues that could happen in other cases. 

However, I think that accordions need to have a small refresh because their implementation sounds weird to me in Boosted.

scss/_accordion.scss
louismaximepiton: I don't really know about that, I think that `.btn-inverse` is there because there is already a `.btn-dark` variant.

julien-deramond: Indeed `.btn-inverse` has this name because `.btn-dark` is already taken.
We thought calling dark variants of Boosted components `.component-dark` because Bootstrap uses it for the [Carousel - Dark variant](https://getbootstrap.com/docs/5.1/components/carousel/#dark-variant) with their `.carousel-dark`.

scss/_variables.scss
Lausselloic: I agree for the need of small refactoring and improvement for accordions to match the guidelines. Agree with this null value, but need to remove this line also : https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/b67299a8f0b03babc1fc94ec322883ed7162efa8/scss/_accordion.scss#L11

louismaximepiton:
Differents solutions I've seen so far, and I'm confused about the one to choose : 

- Adding a background to `<code>` and a fix color to its text. **Issue **: Change every code, even for example code -> not good to me.
- Find a fixed color that goes well for black and white background color at least (#767676 seems to do it on AA level). **Issue **: not flexible + not AAA level.
- Change in every component the text color for `<code>`. **Issue **: Not easy to maintain.
- Change the color value to inherit. **Issue **: Hard to differenciate from the normal text.

julien-deramond:
This dark variant `<code>` issue can be treated separately not to block this PR. I would suggest to:
* remove the use of `<code>` in this dark variant accordion example
* create an issue regarding the dark variant `<code>` in which you can mention the different solutions you have studied and the fact that we need to reintroduce the original accordion example when it'll be fixed

julien-deramond:
> This dark variant `<code>` issue can be treated separately not to block this PR. I would suggest to:
> 
>     * remove the use of `<code>` in this dark variant accordion example
> 
>     * create an issue regarding the dark variant `<code>` in which you can mention the different solutions you have studied and the fact that we need to reintroduce the original accordion example when it'll be fixed

So this topic will be handled in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1036

julien-deramond (changes requested):
As discussed in DM, the issue regarding the `<code> happens too with the links (displayed black with a black bg). We should wait to identify a solution to tackle those 2 topics before merging this PR.

julien-deramond (approved):
Sound good to me after the dark texts PR and some minor modifications. @louismaximepiton I let you check my last commits on this branch. If it's OK for you then I can merge it.

PR #1033, opened 2022-01-11
Chore: Merge `main` @1726041

julien-deramond:


louismaximepiton (approved):
Sounds good to me.

However, one JS test doesn't work in local for me (same in main)

julien-deramond:
> Sounds good to me.
> 
> However, one JS test doesn't work in local for me (same in main)

This error is from Bootstrap. If you run `npm run test` on their _main_ branch, you'll have this error as well with Chrome Headless 97.0.4692. 
It is visible in https://github.com/twbs/bootstrap/runs/4773074647?check_suite_focus=true as well.

PR #1039, opened 2022-01-14
Orange stepbar

louismaximepiton:
Backport to v4 the design of Stepbars. (38px height to 30px)

Preview : [link](https://deploy-preview-1039--boosted.netlify.app/docs/4.6/components/stepbar/)

### DIFF
--- a/scss/_o-step-bar.scss
+++ b/scss/_o-step-bar.scss
@@ -14,7 +14,7 @@
 .stepbar-item {
   display: flex;
   flex: 1;
-  padding: map-get($spacers, 2);
+  padding: $o-stepbar-padding-y map-get($spacers, 2);
   margin-right: $o-stepbar-margin-r;
   font-weight: $font-weight-bold;
   text-align: center;
@@ -54,6 +54,8 @@
   max-width: 1.25ch;
   margin: auto;
   overflow: hidden;
+  font-size: $font-size-sm;
+  line-height: $o-stepbar-line-height;
   color: $white;
   text-decoration: if($link-decoration == none, null, none);
   white-space: nowrap;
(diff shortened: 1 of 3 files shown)


scss/_variables.scss
```diff
@@ -1404,8 +1404,10 @@ $o-scroll-up-bottom: (
 ) !default;
 
 // Step process bar
-$o-stepbar-height:      $spacer * 2 !default;
-$o-stepbar-arrow-width: 1.0625rem !default;
+$o-stepbar-padding-y:   divide(.875rem, 2) !default;
```
julien-deramond, line 1407: IMO you can directly use `.4375rem` rather than the division here.

scss/_o-step-bar.scss
```diff
@@ -14,7 +14,7 @@
 .stepbar-item {
   display: flex;
   flex: 1;
-  padding: map-get($spacers, 2);
+  padding: $o-stepbar-padding-y map-get($spacers, 2);
```
julien-deramond, line 17: I'd use a specific var for the 2nd part:
```scss
$o-stepbar-padding-x:   map-get($spacers, 2) !default;
```

scss/_o-step-bar.scss
```diff
@@ -54,6 +54,8 @@
   max-width: 1.25ch;
   margin: auto;
   overflow: hidden;
+  font-size: $font-size-sm;
+  line-height: $o-stepbar-line-height;
```
julien-deramond, line 58: I'd use here `line-height: $line-height-sm;` here.

Isn't there a bug with `$line-height-sm`? When I apply it, the set value is `calc(divide(16, 14))` and the browser can't use it. Small buttons seems broken as well.

scss/_o-step-bar.scss
```diff
@@ -81,27 +83,30 @@
...
       content: "";
       background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 17 40'><path fill='#{$black}' d='M0 0l14 20L0 40z'/><path fill='#{$white}' d='M3 0H0l14 20L0 40h3l14-20z'/></svg>"));
       background-size: $o-stepbar-arrow-width $o-stepbar-height;
       transform: translateX(#{"/*rtl:insert:-*/"}$o-stepbar-arrow-width) #{"/*rtl:insert:rotate(180deg)*/"};
     }
 
-    &:last-child::after {
-      display: none;
+    &:first-child {
+      padding-left: divide($o-stepbar-arrow-width, 2);
```
julien-deramond, line 105: Could be a specific variable.

julien-deramond (changes requested):
Since the size of this component has changed, the visible focus is more difficult to see. Especially when the step has a black background. This is not perfect in v5 neither but a bit better. Maybe you could look at it to see what we could do about it.

scss/_o-step-bar.scss
louismaximepiton: Yes it looks like a bug, I guess that it comes from lines 280 and 281 in the `_variables.scss`, the calc function is too much there. At least it works well without.

julien-deramond: Issue created: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1066

julien-deramond: This issue is now closed. Resolving this conversation and rebasing this PR.

scss/_o-step-bar.scss
julien-deramond: It isn't `$o-stepbar-padding-x-xc` rather than `$o-stepbar-paddingx-xc`? And what's `xc`?

louismaximepiton: `xc` stands for `extreme-child`

julien-deramond: I don't have something better to propose so let's go with that.

julien-deramond (approved):
Sounds good to me :rocket:

PR #1083, opened 2022-02-01
fix(docs): make visible 'Design guidelines' menu item when the page is active

julien-deramond:
Going onto [Orange Design System for web](https://deploy-preview-1083--boosted.netlify.app/docs/5.1/guidelines/) page make it active (and visible) in the menu bar.

Screenshot from 2022-02-01 14-40-54

If you do the same thing locally with the actual _main_ branch, the menu item won't be active.


louismaximepiton (approved):
Sounds good for me.

PR #1087, opened 2022-02-03
fix(accordion): Remove active text color on focus

razyon:
Orange Design System does not requires a font color change on focus for the accordion. We find it a bit confusing with the over state.


julien-deramond:
Thanks for the PR @razyon! 

In the current version of Boosted, the focus has the same color than the hover state for the accordions. We have this same behavior in the header for example: when a new menu item is focused the text is displayed in orange as well even if it is not yet active.

However from what I see in the [DSM](https://system.design.orange.com/0c1af118d/p/9586a0-accordion/b/4851ef) we should have this orange color only when it its pressed and/or expanded.
It would mean that indeed the focus state never needs to change the text color.

Let me just check with the designers to clarify some points about the different hover and pressed states.

julien-deramond:
Checked with the designers:
* Regarding the focus state the text colour does not change and to focus outline is added
* The listing of the states is not complete. They will update the DSM and we will modify Boosted accordingly.

This PR is good to merged. Thx for your contribution.

PR #1089, opened 2022-02-04
Sticky header example

louismaximepiton:

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -114,17 +114,19 @@
   }
 
   .nav-link {
-    white-space: nowrap;
+    display: block;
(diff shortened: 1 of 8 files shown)


Nurovek:
Some comments to remind us what to do to fix it:
- Fix things that have been rolled back (some css in `_orange-navbar`, ...)
- Use `.fixed-top` on both ```<nav>```s (supra + global)
    - Indeed the two bars must be sticky, not just the supra bar
- Once it's done, play with the positioning for the supra bar (if both are fixed-top, the supra bar is under the global header)
- Fix alignments (the supra is not aligned well with the global header)

Will edit this comment if I forgot something.

PR #1096, opened 2022-02-11
Chore/merge main@a794219

julien-deramond:


site/content/docs/5.1/components/badge.md
```diff
@@ -84,8 +84,16 @@ Use the `.rounded-pill` utility class to make badges more rounded with a larger
...
 {{< /example >}}
 
-## Sass
+## CSS
 
 ### Variables
 
+<small class="d-inline-flex px-2 py-1 font-monospace text-muted border rounded-3">Added in v5.2.0</small>
+
+As part of Boosted's evolving CSS variables approach, buttons now use local CSS variables on `.badge` for enhanced real-time customization. Values for the CSS variables are set via Sass, so Sass customization is still supported, too.
```
julien-deramond, line 93: FYI I created a PR in Bootstrap to change "buttons" into "badges"

scss/_badge.scss
```diff
@@ -1,22 +1,32 @@
+// stylelint-disable custom-property-empty-line-before
```
louismaximepiton, line 1: With the correction below, don't need this

scss/_badge.scss
```diff
@@ -1,22 +1,32 @@
...
+
 // Base class
 //
 // Requires one of the contextual, color modifier classes for `color` and
 // `background-color`.
 
 .badge {
+  // scss-docs-start badge-css-vars
+  --#{$variable-prefix}badge-padding: #{subtract($badge-padding-y, 1px) $badge-padding-x add($badge-padding-y, 1px)}; // Boosted mod
+  @include rfs($badge-font-size, --#{$variable-prefix}badge-font-size);
```
louismaximepiton, line 11: This line could be at the end of CSS var declarations IMO

louismaximepiton (commented):
Is there a need for small changes ?

scss/_badge.scss
julien-deramond: I'm going to merge as it is in order to avoid having discrepancies with Bootstrap (for easier merges in the future). But if the fix is OK, you could propose it to Bootstrap so we have it back with the next merges.

louismaximepiton (approved):
Sounds good to me !

PR #1124, opened 2022-02-28
implement default focus-visible

Lausselloic:
doesn't only relies on polyfill focus-visible but also implement native :focus-visible
https://caniuse.com/css-focus-visible support is not so bad at time


julien-deramond (approved):
Haven't seen any regressions so LGTM! 🚀 
Thx for this enhancement :)

PR #1143, opened 2022-03-09
Chore/merge main@3a57da4

julien-deramond:


scss/_popover.scss
```diff
@@ -146,15 +183,15 @@
...
-  color: $popover-header-color;
+  color: var(--#{$variable-prefix}popover-header-color);
 
   /* rtl:remove */
   letter-spacing: unset; // Boosted mod
-  background-color: $popover-header-bg;
-  border-bottom: $popover-border-width solid shade-color($popover-header-bg, 10%); // Boosted mod
+  background-color: var(--#{$variable-prefix}popover-header-bg);
+  border-bottom: var(--#{$variable-prefix}popover-border-width) solid shade-color($popover-header-bg, 10%); // Boosted mod
   @include border-top-radius($popover-inner-border-radius);
```
julien-deramond, line 195: I'll propose to Bootstrap to create a CSS var as for other components

scss/_popover.scss
```diff
@@ -99,30 +133,33 @@
...
   }
 }
 
 /* rtl:begin:ignore */
 .bs-popover-start {
   > .popover-arrow {
-    right: subtract(-$popover-arrow-height, $popover-border-width);
-    width: $popover-arrow-height;
-    height: $popover-arrow-width;
+    right: calc((var(--#{$variable-prefix}popover-arrow-height) * -1) - var(--#{$variable-prefix}popover-border-width)); // stylelint-disable-line function-disallowed-list
```
julien-deramond, line 146: I'll propose a patch for Bootstrap for those calculations; `- var(--#{$variable-prefix}popover-border-width)` is not present in Bootstrap and is mandatory for us at least.

scss/_popover.scss
```diff
@@ -146,15 +183,15 @@
 
 // Offset the popover to account for the popover arrow
 .popover-header {
-  padding: $popover-header-padding-y $popover-header-padding-x;
+  padding: var(--#{$variable-prefix}popover-header-padding-y) var(--#{$variable-prefix}popover-header-padding-x);
   margin-bottom: 0; // Reset the default from Reboot
   @include font-size($font-size-base);
```
louismaximepiton, line 188: Do we need to create a CSS var for this font-size ? If yes -> backport to Bs

scss/_variables.scss
```diff
@@ -1489,7 +1489,9 @@ $tooltip-margin:                    $spacer * .25 !default;
 
 $tooltip-arrow-width:               $spacer * .5 !default;
 $tooltip-arrow-height:              $tooltip-arrow-width * .5 !default;
-$tooltip-arrow-color:               $tooltip-bg !default;
+// fusv-disable
+$tooltip-arrow-color:               null !default; // Deprecated in Boosted 5.2.0 for CSS variables
```
louismaximepiton, line 1493: Shouldn't we keep the `$tooltip-bg` as it is for the popover ?

louismaximepiton (commented):
I think I get a bug on tooltips, but I wanna talk about it with you since it is a bit hard to reproduce : custom tooltip clicked, zoom on it and scroll vertically until the tooltip is pushed to the right : initialization of the component seems to be broken.

louismaximepiton (approved):

PR #1144, opened 2022-03-09

julien-deramond:
This PR is a proposal to modify the margin left value between the required symbol (the `*`)  and the corresponding label.

The current value in Boosted is 6px: https://boosted.orange.com/docs/5.1/forms/overview/#required-field.

Screenshot from 2022-03-09 11-47-56

The value [defined in the DSM](https://system.design.orange.com/0c1af118d/p/88ab5b-forms/b/599459/i/48901789) is 3px.

[**Preview of the modification**](https://deploy-preview-1144--boosted.netlify.app/docs/5.1/forms/overview/#required-field) (@CyriaqueBillard I let you check the rendering and tell us if it sounds good to you)

Screenshot from 2022-03-09 11-52-58


louismaximepiton (approved):
LGTM.

PR #1145, opened 2022-03-09

julien-deramond:
Backport of the fix explained in #1144. :warning: Wait for its review before merging this one.

This PR doesn't need a design review since it would have already been done in the other PR. Just check that this code does the same thing than in the other PR and covers all the rendering cases.


louismaximepiton (commented):
Maybe need a change in `_o-forms.scss` ? (l.85)

PR #1154, opened 2022-03-18
fix(pagination): add margin only after first element and before last …

### LINKED ISSUE #1120: Docs > Components > Pagination > Disabled and active states: Fix css
### Prerequisites

### Describe the issue

In Boosted, pagination numbers have a spacing (margin-left) of 10px when they don't in ODS => to be corrected to be conform to ODS.
In ODS : 
image

In Boosted : 
image

### Reduced test cases

NA

macOS

Chrome

v5.1.0

hannahiss:
…(if they are not alone)

Beware to the result when navigating with keyboard (focus-visible) : 
image

closes #1120

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -1,24 +1,27 @@
+// stylelint-disable custom-property-empty-line-before
+
 .pagination {
   // scss-docs-start pagination-css-vars
-  --#{$variable-prefix}pagination-padding-x: #{$pagination-padding-x};
-  --#{$variable-prefix}pagination-padding-y: #{$pagination-padding-y};
-  --#{$variable-prefix}pagination-color: #{$pagination-color};
-  --#{$variable-prefix}pagination-bg: #{$pagination-bg};
-  --#{$variable-prefix}pagination-border-width: #{$pagination-border-width};
-  --#{$variable-prefix}pagination-border-color: #{$pagination-border-color};
-  --#{$variable-prefix}pagination-border-radius: #{$pagination-border-radius};
-  --#{$variable-prefix}pagination-hover-color: #{$pagination-hover-color};
-  --#{$variable-prefix}pagination-hover-bg: #{$pagination-hover-bg};
-  --#{$variable-prefix}pagination-hover-border-color: #{$pagination-hover-border-color};
-  --#{$variable-prefix}pagination-focus-color: #{$pagination-focus-color};
-  --#{$variable-prefix}pagination-focus-bg: #{$pagination-focus-bg};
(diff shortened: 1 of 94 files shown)


hannahiss:
After some talks, I arrived to the conclusion that this change of focus is the new way to 

> 

After some discussion, I came to the conclusion that this change of focus (which is the new way of doing focuses) is not part of this issue. It must be treated with the new focuses for buttons. So far, this version seems correct. To confirm...

scss/_pagination.scss
```diff
@@ -136,15 +132,16 @@
...
   &:first-child .page-link {
+    margin-right: $pagination-margin-start;
+
     &:not(:empty) {
       padding-right: $pagination-padding-end;
     }
 
-    &::before {
-      /* rtl:raw:
+    /* &::before {
```
louismaximepiton, line 141: IMO, you should keep it as is since `/* rtl:raw */` is there to complete rtl version of Boosted.

ATM in rtl : 
image

scss/_pagination.scss
```diff
@@ -160,6 +157,12 @@
     }
   }
 
+  &:last-child {
+    &:not(:nth-child(2)) .page-link{
```
louismaximepiton, line 161: This selector could be on the same line

scss/_pagination.scss
```diff
@@ -78,10 +78,6 @@
 }
 
 .page-item {
-  &:not(:first-child) .page-link {
```
louismaximepiton, line 81: IMO, this should be kept as is since it comes directly from Bootstrap, but you could set `$pagination-margin-start` to 0

scss/_pagination.scss
```diff
@@ -136,15 +132,16 @@
   }
 
   &:first-child .page-link {
+    margin-right: $pagination-margin-start;
```
louismaximepiton, line 135: The scss var used here should be explicit (e.g. `$pagination-margin-x`)

louismaximepiton (changes requested):
Few things to change.

I don't know if it's worth a try, but- `l.81` maybe a `&:nth-child(2), &:last-child` selector could work here ? (I didn't tried but maybe you tried before ?)

scss/_pagination.scss
hannahiss: Ok, for my information, and next time, how can I know that it comes directly from Boosted please ?

louismaximepiton: We used to add `// Boosted mod` at the end of the lines when the modifications only affects few lines.
Otherwise, we used to add `// Boosted mod` and `// End mod` that encapsulate lines of code if it affects several lines.

hannahiss:
Sorry I had to close the PR but my branch is contaminated just like me

PR #1177, opened 2022-04-01

julien-deramond:
Backport of https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1144 in v4.

[Live Preview](https://deploy-preview-1177--boosted.netlify.app/docs/4.6/components/forms/#custom-styles)

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -8,7 +8,7 @@ label {
   }
 }
 
-.is-required::after {
+.is-required:not(.form-group)::after {
   margin-left: $label-required-margin-left;
   font-weight: $font-weight-bold;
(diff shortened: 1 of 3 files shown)


louismaximepiton (commented):
I think you could change the value in `o-forms.scss` ?

julien-deramond:
Good catch. I didn't know it was even possible to add a `.is-required` directly on the form group.

I fixed at the same time a weird behavior with `.form-group .is-required` and the double stars:
Screenshot from 2022-04-01 12-02-52

louismaximepiton (approved):
LGTM !

PR #1188, opened 2022-04-05
chore(docs): reorder alphabetically list of components

julien-deramond:
This PR is a proposal to reorder alphabetically the list of components in the documentation. IMHO it is simpler to find the components in the list and more natural.

**After the merge**
- [ ] If we agree on that, I'll create a PR for Bootstrap as well
- [x] Check if it can be backported to v4


julien-deramond:
> Maybe the one in `docs/5.1/getting-started/introduction.md > Quick start > JS > Components` could be handled in this PR too ?

Yep ofc, I thought it was already good; haven't seen the last one 🙃 Thx!

louismaximepiton (approved):
Sounds good to me.

PR #1213, opened 2022-04-20
fix(docs): replace '..' by '…' in the documentation title

julien-deramond:
Before this modification, screen reader said "and ergonomic components library dot dot".
With this modification, screen reader will say ""and ergonomic components library".

/CC @Aniort ;) Thanks again for your presentation about screen readers. I tried it with Orca on Ubuntu.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -54,7 +54,7 @@ module:
       target:                       static/docs/5.1/assets/js/lang
 
 params:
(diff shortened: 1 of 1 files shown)


louismaximepiton (approved):
Was a bit complicated to review but I can't see no regression 🚀

Issue #1121, opened 2022-02-28
Docs > Components > Buttons > Dark background: Fix border color for social buttons on active state


isabellechanclou:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

On https://boosted.netlify.app/docs/5.1/components/buttons/#dark-background, the value of border-color for social inverse buttons (= with a dark background) active state don't change, so on mouse down, visually, it seems that the buttons get smaller.

### Reduced test cases

NA

### What operating system(s) are you seeing the problem on?

Windows, macOS

### What browser(s) are you seeing the problem on?

Chrome, Firefox

### What version of Boosted are you using?

v5.1.0

hannahiss:
I don't know how to reach the branch in which  dark background version of social buttons have been implemented.

Lausselloic:
https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1010
https://deploy-preview-1010--boosted.netlify.app/docs/5.1/components/buttons/#dark-background

PR #1224, opened 2022-04-29
fix(social-btn): add white border on dark background for active state

hannahiss:
on dark background, need a white border on active tate so that the size of button doesn't decrease when active (report form branch main-his-social-border-on-dark)


hannahiss:
Linked to #1121

louismaximepiton:
Just to remind what we've said in private : 
The `.active` should be back-ported to v4, since this class is actually brought by the `button-variant` mixin.

hannahiss:
> Just to remind what we've said in private : The `.active` should be back-ported to v4, since this class is actually brought by the `button-variant` mixin.

You can review. Thanks

scss/_buttons.scss
```diff
@@ -190,14 +190,17 @@ fieldset:disabled a.btn {
     }
 
     &,
-    &.btn-inverse:active {
+    &.btn-inverse:active,
+    &.btn-inverse.active {
       background-color: $white;
       background-image: escape-svg(url("data:image/svg+xml,#{$icon}"));
+      border-color: $white;
```
louismaximepiton, line 197: This line should maybe be in another selector since it breaks the normal social buttons

louismaximepiton (changes requested):
:warning: Normal social buttons are broken ⚠️

julien-deramond (approved):

PR #1226, opened 2022-04-29
fix(color): add indigo to colors (existing but not documented).

### LINKED ISSUE #1225: Docs > Customize > Color: $indigo is missing in All colors.
Even if it is the same color than <code class="notranslate">$purple` it may be interesting to make it appear in the documentation because it exists!

hannahiss:
Closes #1225

### DIFF
--- a/site/content/docs/5.1/customize/color.md
+++ b/site/content/docs/5.1/customize/color.md
@@ -28,14 +28,14 @@ Check out [our Sass maps and loops docs]({{< docsref "/customize/sass#maps-and-l
 
 ## All colors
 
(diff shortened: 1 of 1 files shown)


louismaximepiton (approved):
LGTM, I would have wrote something like: `... the indigo colors are the same as the purple ones.` But as-is looks fine to me.

hannahiss:
> LGTM, I would have wrote something like: `... the indigo colors are the same as the purple ones.` But as-is looks fine to me.
C'est un projet complet, on apprend l'anglais aussi ;-)

julien-deramond (approved):

Issue #901, opened 2021-11-09
Boosted v4 ⭢ v5 (remaining tasks)


julien-deramond:
This issue tracks all the remaining things to backport from Bootstrap v4.

For all actions done from this epic issue:
* Don't forget to update the migration page
* Ensure to add deprecation warnings in v4 when required (see [#560](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/560) / [#562](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/562)  for examples).

- [x] Gather and sort items from https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/383 then delete it as duplicate
- [x] Check the content of https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/640 and define an action_
- [x] Compare our v5 components with the design specifications and note the differences
- [x] Merge this list with the ones in comments
- [x] Study each item and create an issue for each one or delete it
- [x] Document somewhere ([Discrepancies between Boosted and Bootstrap](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/wiki/Discrepancies-between-Boosted-and-Bootstrap)?)Bootstrap parts dropped in Boosted (and the reasons as an history) and add in the DoDs to update this doc
  * no `.floating-label`
  * no `.form-control-sm`, `.form-select-sm`, `.input-group-sm`, `.col-form-label-sm` -> Should change with #1230
  * no forms' `.#{$state}-tooltip`
  * no `table-row-variants()`, `.table-bordered`, `.table-borderless`, `.table-striped` -> Should change with #859 
  * no `.btn-outline`
  * no `.pagination-sm` nor `.pagination-lg`
  * no active state for `.list-group-item-action` -> Should change with #1467
  * no floating labels
  * no carousel dark variant
  * alerts sizes & icons
  * [Tabs light](https://boosted.orange.com/docs/5.1/components/navs-tabs/#tabs-light)
  * Responsive pagination with `pagination-max-items()`
  * Progress sizes
  * Spinners sizes
  * Growing spinners -> Might have changed with #1041
  * `:focus-visible` is still here
  * Tools: bots Mattermost & Slack, Percy, Pa11y, Calibre App ;
  * Form validation: success and errors do not share the same layout, and icons aren't shown as inputs' `backgrounds`
  * Range input with filled thumb
[...]

PR #1228, opened 2022-04-29
fix(doc): add checks and radios without labels

hannahiss:
Closes [#1127 ](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1227)


hannahiss:
I'm not sure if it has to be done: it is asked in issue #901  but code specifies: 
`<!-- Boosted mod: don't promote "without label" -->`

louismaximepiton (approved):
Ah your right about it, I don't have any idea about this one. Maybe we should warn about it with an ODS alert ? Any thoughts @julien-deramond ?

julien-deramond (approved):
IMO it is not really related to ODS constraints so let's bring it back as it is.

Issue #1084, opened 2022-02-02
ODS > Navigation > Carousel Navigation


B3nz01d:
# Carousel Navigation

## Requirement

**For** a developer
**Who** is creating an Orange website
**The** carousel navigation
**is a** component
**that** allows to manipulate a carousel
**Boosted** will provide this ready to use component compatible with Orange design specifications and a11y requirements

[DSM reference](https://system.design.orange.com/0c1af118d/p/7773e1-carousel-navigation/b/99a7b7)

## General description

Carousel navigation allows users to navigate forward and backwards through content in a carousel component.

### Anatomy

Carousel navigation is composed of:
1. an arrow
2. a rounded container

### States

Carousel navigation has four possible states:
* Default
* Hover
* Pressed
* Disabled

### Appearance

Carousel navigation includes a black chevron which is set within a 50% white container to allow them to remain visible when placed over different colored backgrounds or imagery.
The 20% transparency is applied to the Default & Disabled States.
The carousel navigation is not displayed anymore under the Medium Screen size break point.

### Placement

The carousel navigation is positioned to the left and right edges of the carousel and centered vertically. For more examples of the carousel navigation used in hero banners, see the promotions examples in modules.

### Interaction

* Mouse: the component must be clickable
* Keyboard: the component will inherit the button interaction (focus + press "Enter")
    * Note: the carousel can also be manipulated with "left" and "right" keys but it will handle itself this behavior
* Touch (mobile): the component will inherit the button behavior on mobile
    * Note: the carousel can be manipulated by swiping the content but it will handle itself this behavior

### Animations

No animation for this component

## Accessibility

A good practice on [carousel accessibility](https://www.w3.org/WAI/tutorials/carousels/working-example/)

### Visual Accessibility

The image won't be used by screen readers; a visually hidden text will be used to inform the purpose of the commands: "Previous" / "Next".

#### Question: How the visible focus will be displayed?

- [ ] Check the current implementation to see if it is possible to have a visible focus (an outline) just around the button

### Usage Accessibility

N/A

## Technical specifications

This component won't be considered as a component but rather as a part of the Carousel component. So it won't be provided on its own.

### HTML Structure

The base HTML element will be a `<button>`.
The icon will be displayed within `<span>` (that must be hidden for screen reader) with a `:before` rule.
The image must be embedded by Boosted.

### JavaScript

No JavaScript directly for this component. JS will be handled by the Carousel component.

[...]

PR #1242, opened 2022-05-09
Carousel navigation focus

MewenLeHo:
Closes [#1084 ](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1084)

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -135,7 +135,6 @@
       outline: 3px solid $brand-orange;
       outline-offset: 2px;
       box-shadow: 0 0 0 2px $white;
-
     }
   }
   // End mod
@@ -219,12 +218,22 @@
     &:hover,
     &:focus {
       background-color: color-contrast($carousel-indicator-active-bg);
-      transform: scale($carousel-indicator-hover-scale);
 
       &::before {
         transform: translate3d(-50%, -50%, 0) scale($carousel-indicator-active-scale);
       }
     }
+
+    &:hover {
+      transform: scale($carousel-indicator-hover-scale);
+    }
+
+    &:focus {
+      outline: 3px solid $brand-orange;
+      outline-offset: 2px;
+      box-shadow: 0 0 0 2px $white;
+    }
+
   }
 
   .active {
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -104,7 +104,7 @@
   text-align: center;
   background: none;
   border: 0;
-  outline: 0;
+  outline: 0; // Boosted mod
   opacity: $carousel-control-opacity;
   @include transition($carousel-control-transition);
 
@@ -129,7 +129,6 @@
     color: $carousel-control-color;
   }
 
-  &:focus:focus-visible,
   &:focus {
     &[data-focus-visible-added] {
       & > span {
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -104,6 +104,7 @@
   text-align: center;
   background: none;
   border: 0;
+  outline: 0;
   opacity: $carousel-control-opacity;
   @include transition($carousel-control-transition);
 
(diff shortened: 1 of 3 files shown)


scss/_carousel.scss
```diff
@@ -127,6 +128,15 @@
     --#{$boosted-prefix}control-bg: #{$carousel-control-icon-active-bg}; // Boosted mod
     color: $carousel-control-color;
   }
+
+  &:focus:focus-visible,
+  &:focus {
```
louismaximepiton, line 133: I think this selector could be upgraded with a `[data-focus-visible-added]` in order to not have the focus state on click

scss/_carousel.scss
```diff
@@ -127,6 +128,15 @@
...
+    & > span {
+      outline: 3px solid $brand-orange;
```
louismaximepiton, line 135: Think using rem could be better in the style page

scss/_carousel.scss
```diff
@@ -208,12 +218,22 @@
     &:hover,
     &:focus {
```
louismaximepiton, line 219: I guess that the weird behavior I can see is due to these selectors : 
- Focus the current page indicator
- There's a semi-moon that rolls over

louismaximepiton (changes requested):
I think the js should be improved ? Maybe on Bs side

Focus on carousel doesn't seem to pause it
Focus on indicators is weird : the current page indicator is kinda scaled up

MewenLeHo:
> I think the js should be improved ? Maybe on Bs side
> 
> Focus on carousel doesn't seem to pause it Focus on indicators is weird : the current page indicator is kinda scaled up

Should the carousel pauses on focus? Maybe. It may be useful for keyboard users.
What do you think about it @julien-deramond? It's a question for @Aniort and our accessibility experts.
However it's not the subject of this pull request.

julien-deramond:
> Should the carousel pauses on focus? Maybe. It may be useful for keyboard users. What do you think about it @julien-deramond? It's a question for @Aniort and our accessibility experts. However it's not the subject of this pull request.

Not sure it should since you're focusing the arrow, not the carousel itself (the content). But I let @Aniort gives us his feedback.
Agree that's not something to take into account in this PR. As said by @louismaximepiton if we change this behavior let's propose it directly to Bootstrap.

scss/_carousel.scss
```diff
@@ -104,6 +104,7 @@
   text-align: center;
   background: none;
   border: 0;
+  outline: 0;
```
julien-deramond, line 107: Missing `// Boosted mod`

scss/_carousel.scss
julien-deramond: What's the status of this comment? Resolved?

scss/_variables.scss
```diff
@@ -563,7 +563,8 @@ $container-padding-x: $grid-gutter-width !default;
 $border-width:                .125rem !default;
 $border-widths: (
   1: $border-width * .5,
-  2: $border-width
+  2: $border-width,
+  3: $border-width * 1.5
```
julien-deramond, line 567: By doing this you're adding a `.border-3` utility so it must be documented in https://deploy-preview-1242--boosted.netlify.app/docs/5.1/utilities/borders/#width by adding the following example: `<span class="border border-3"></span>`.

Note: an impact is observed where `.border-3` is used in the docs but is handle directly via https://github.com/twbs/bootstrap/pull/36476.

scss/_carousel.scss
```diff
@@ -124,9 +125,20 @@
   }
 
   &:active {
-    --#{$boosted-prefix}control-bg: #{$carousel-control-icon-active-bg}; // Boosted mod
+    --#{$boosted-prefix}control-bg: #{$carousel-control-icon-active-bg};
     color: $carousel-control-color;
   }
+
+  &:focus:focus-visible,
```
julien-deramond, line 132: Just out of curiosity, why do we need this `&:focus:focus-visible`? In other component just having `&:focus { &[data-focus-visible-added] {` seems sufficient.

scss/_carousel.scss
```diff
@@ -208,12 +220,24 @@
...
+    }
 
+    &:hover {
+      transform: scale($carousel-indicator-hover-scale);
       &::before {
         transform: translate3d(-50%, -50%, 0) scale($carousel-indicator-active-scale);
       }
     }
+
+    &:focus {
```
julien-deramond, line 232: :thinking: Buttons in `.carousel-indicators` are focusable?

julien-deramond (changes requested):

scss/_variables.scss
MewenLeHo: Done

PR #1253, opened 2022-05-11
Cards : Adding CSS var

louismaximepiton:


scss/_card.scss
```diff
@@ -18,6 +18,7 @@
   --#{$prefix}card-cap-color: #{$card-cap-color};
   --#{$prefix}card-height: #{$card-height};
   --#{$prefix}card-color: #{$card-color};
+  --#{$prefix}card-font-weight: #{$card-font-weight}; // Boosted mod
```
julien-deramond, line 21: IMO the naming is misleading: the usage is limited to `card-header` and `card-footer` and not to the entire font-weight of cards. Can't we have a name more specific here?

julien-deramond (approved):

PR #1255, opened 2022-05-12
🎤 A dollar is what I need

julien-deramond:
Two missing dollars in CSS variables definitions. Just check that there's no regression.


louismaximepiton (approved):
Couldn't spot any element where the `.go` was used + couldn't spot any regression.

PR #1257, opened 2022-05-13
chore/merge-main: finish remaining modifications after Bootstrap design doc refactoring

julien-deramond:
- [ ] Remove and take into account all `////////////////////////////////////////////////////////////` comments
- [ ] Remove and take into account all `<!---------------------------------------------------->` comments


julien-deramond:
@louismaximepiton Proposing via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1257/commits/be2196b6a5fa306f24fa34fe2410c8f0bed93c9e to remove the specific masthead h1 rule in order to have the responsive behavior of `.display-1` and have a nice size in mobile. LGTY?

site/assets/scss/_clipboard-js.scss
```diff
@@ -44,37 +37,19 @@
 
   // Boosted mod
   .bi {
-    width: 1.4em;
-    height: 1.4em;
+    width: 1.8125rem;
```
julien-deramond, line 40: I'd stick with `1.4em`. It is rather huge when displayed in the action bar next to the "Edit on StackBlitz" button.

site/assets/scss/_search.scss
```diff
@@ -10,7 +10,7 @@
 .DocSearch-Container {
   --docsearch-hit-shadow: none;
 
-  z-index: 1030;
+  z-index: 1100; // value 11000000000 seems to hide scrollbars under the backdrop
```
julien-deramond, line 13: 11000000000?

site/assets/scss/_clipboard-js.scss
louismaximepiton: I agree that the size is way better but should we keep it even if they are a bit blurry ?

PR #1259, opened 2022-05-16
Title bars : Add CSS var

louismaximepiton:


julien-deramond (changes requested):
It is `--#{$prefix}` everywhere instead of `--#{prefix}`

julien-deramond (approved):

PR #1269, opened 2022-05-19
fix(pagination): Remove space between pagination numbers in v4

hannahiss:
Also remove margin at start

Backport of #1120 to v4

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -8,7 +8,6 @@
 
 .page-link {
   // Boosted mod
-  // position: relative;
   display: flex;
   justify-content: center;
   min-width: $pagination-item-size;
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -207,10 +207,7 @@
       display: none;
     }
   }
-  // End mod
 
-
-  // Boosted mod
   &:nth-child(2):not(:last-child) .page-link {
     margin-left: $pagination-margin-x;
   }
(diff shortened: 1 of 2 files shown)


scss/_pagination.scss
```diff
@@ -209,6 +208,17 @@
     }
   }
   // End mod
+
+
+  // Boosted mod
```
louismaximepiton, line 213: Could be packed with the previous one IMO

scss/_variables.scss
```diff
@@ -932,6 +932,7 @@ $dropdown-header-padding:           $dropdown-padding-y !default; // Boosted mod
 // Pagination
 
 $pagination-margin-y:               $spacer !default;
+$pagination-margin-x:               .5rem !default; // Boosted mod
```
louismaximepiton, line 935: Should be `.625rem`

louismaximepiton (changes requested):
Focus state on the active one looks kind of weird to me

scss/_variables.scss
hannahiss: ah oui, j'ai gardé la même valeur sans vérifier sur le deisgn...

louismaximepiton (approved):
The selectors aren't the same in v4 and v5, does that impact anything ? 
However, LGTM

julien-deramond (approved):
LGTM! FYI made some minor changes in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1269/commits/6d12751b1f09ca9ffbc03a9b3b19e020b13e3fe2.

julien-deramond:
> The selectors aren't the same in v4 and v5, does that impact anything ? However, LGTM

Just to answer to your question @louismaximepiton. The structure is not exactly the same so no problem to have different selectors in v4 and v5 in this case.

PR #1270, opened 2022-05-19
Added online version of autoprefixer

rahulkumarsingh73690:
Added online version of Autoprefixer (Featured on mozilla). So users can instantly use it without installing it.


julien-deramond:
Thanks @rahulkumarsingh73690 . We are a fork of Bootstrap so we'll stick with their decision in https://github.com/twbs/bootstrap/pull/36393.

PR #1272, opened 2022-05-20
fix(table): row selection: space after checkbox incorrect

### LINKED ISSUE #1236: Docs > Tables > Row selection: space after checkbox incorrect
### Prerequisites

### Describe the issue

In Docs > Content > Tables > Row selection, the space between the checkbox and the following text is too large compared to what's in ODS on https://system.design.orange.com/0c1af118d/p/21b04a-data-tables/b/56a1a9/i/66611045 "WEB-TBL-STD-002" => space to be fixed.
image

When looking at the problem quickly, it can be solved by adding a `margin-right: -12px`in .formcheck`, but another solution would certainly be better.

### Reduced test cases

NA

macOS

Chrome

v5.1

hannahiss:
Add a negative right margin to adjust space in that case

Closes #1236


scss/_tables.scss
```diff
@@ -48,6 +48,14 @@
   > tbody {
     vertical-align: inherit;
   }
+
+  // Boosted mod
+  // Add negative right margin for selectable rows
+  td:first-child .form-check,
+  th:first-child .form-check {
+    margin-right: -.75rem;
```
louismaximepiton, line 56: Is there any reason for this specific value ? IMO, a greater value would be better here ?

louismaximepiton (commented):
Out of context but, did you try changing the html structure ? (without the `div.form-check` for example)

scss/_tables.scss
hannahiss: It seems to me that form-check is needed to surround checkbox input with form-check-input. 
Moreover, removing form-check does not solve the problem and removing padding is not ok (it is a left padding and checbox image is added as a background).

hannahiss: For the value, it has been proposed by Isabelle but maybe you're right, -1rem could be nicer...

With -0.75rem :
image

With -1rem :
image

louismaximepiton: Yes the 1rem looks better regarding the specs ! If I were a bit picky, I'd say that `1.0625rem` would be pixel perfect and IMO, `$spacer * .85` would be even more understandable.
The management of checkboxes seems to be a bit weird to me even on Bs side, wondering if there is anything to change on their side ... I don't understand from where the spacing after the checkboxes come. The solution here looks fine to me

louismaximepiton (approved):
LGTM 🚀

julien-deramond:
There's no way here to have 40px x 40px? It seems to be like that [in the DSM](https://system.design.orange.com/0c1af118d/p/21b04a-data-tables/b/56a1a9/i/66611045).

Screenshot from 2022-06-09 08-07-09

julien-deramond:
With `margin-right: -1.024rem` we have a `<td>` width of:
- Ubuntu / Firefox: 40px
- Ubuntu / Chrome: 40.02px
What about Windows? If we have the same values (really close to 40px) we could go for this `-1.024rem` :shrug:

louismaximepiton:
After digging a bit more inside tables, it appears that the columns take as much place as they can. This is why we have such weird sizes (you can try resize your window and you won't have the same size). Here is a workaround to have exactly 40px at each breakpoint (on first column): 

```html
<table class="table table-sm table-hover table-responsive">

  <colgroup>
    <col style="2.5rem">
  </colgroup>

  <thead>
  </thead>
  <tbody>
  </tbody>
</table>
```

Here the 2.5rem is arbitrary, I didn't check if another value could be better here

hannahiss:
Louis-Maxime's solution seems good and well supported by all browsers, so I added colgroup to fix first column width. 
Concerning the negative right margin, it is still needed to prevent second column to move to the right, but 1rem seems ok and cleaner that 1,0625rem or -1.024rem.

hannahiss:
@julien-deramond I'm going to update migration guide too ;-)

site/content/docs/5.2/content/tables.md
```diff
@@ -643,6 +643,10 @@ The selection behavior isn't implemented yet. This feature will be delivered wit
 
 <div>
   <table class="table table-sm table-hover table-responsive">
+    <colgroup>
```
julien-deramond, line 646: Not a big fan of adding a `colgroup` here. Can't be handle everything by using the same thing that was done in v4?

Using a `.has-checkbox` in `<table>` plus a rule similar (+ some adjustements) to:

```scss
// Boosted mod
.has-checkbox tr > :first-child {
  width: $spacer;
}
// end mod
```

site/content/docs/5.2/migration.md
```diff
@@ -33,6 +33,11 @@ If you need more details about the changes, please refer to the [v5.2.2 release]
 
 - <span class="badge bg-warning">Warning</span> All forms examples have been modified to add a `.mt-2` to all submit buttons in order to always have 30px between the last form control and the button. Please reflect this modification into your websites.
 
+### Contents
+
+- **Tables**
+  - When using row selection (checkboxes), use `.has-checkbox` on the table to get correct spacing on the first column.
```
julien-deramond, line 39: Suggestion:

```suggestion
  - Reintroduction of `.has-checkbox` from v4 to display correctly the row selection in the first column of tables. 
```

julien-deramond (approved):
LGTM. Just a suggestion to improve the migration guide note.

scss/_tables.scss
```diff
@@ -48,6 +48,18 @@
...
     vertical-align: inherit;
   }
+
+  // Boosted mod
+  // When using checkboxes in the first column, force width of this column and add negative right margin
+  // to checkbox, to ensure correct spacing on the first column.
+  &.has-checkbox tr > :first-child {
+    width: $spacer * 2;
+
+    .form-check {
```
louismaximepiton, line 58: The whole rule on `.form-check` could be replaced by `max-width: $spacer * 2;`. Not sure what's the best here 🤷

julien-deramond: Can someone try it? If we can do the same thing with a reference to `$spacer`, it would be better than a negative margin and magical value.

site/content/docs/5.2/migration.md
hannahiss: I also add a warning because, if this is a migration guide from 5.1 to 5.2, adding has-checkbox class is needed...

scss/_tables.scss
hannahiss: Ok it works with both width and max-width... 
Tested on Windows/Chrome, Windows/Firefox and on Browserstack : iOS/Safari and MacOS/Safari

PR #1280, opened 2022-05-24
Use Footer component in docs

### LINKED ISSUE #1221: Use the footer component in Boosted
We need to use the footer component to display our footer.

hannahiss:
After an exchange with Cyriaque, we agreed to 

- Position the text to the right of the navigation links (so after these links).
- Delete the line break after the first sentence.
- Delete the list that surrounds the version number and the Netlify link.

Closes #1221

### DIFF
--- a/site/assets/scss/_footer.scss
+++ b/site/assets/scss/_footer.scss
@@ -3,17 +3,13 @@
 //
 
 .bd-footer {
-  @include font-size(.875rem);
-  position: sticky;
-  top: 100vh;
-  font-weight: 700;
+  // Boosted mod
+  .bd-footer-info {
+    border-top: var(--#{$prefix}accordion-border-color) solid var(--#{$prefix}accordion-border-width);
 
-  a {
-    color: var(--#{$prefix}link-color); // Boosted mod: use CSS var instead of $link-color
-
-    &:hover,
-    &:focus {
-      color: var(--#{$prefix}link-hover-color); // Boosted mod: use CSS var instead of $link-hover-color
+    @include media-breakpoint-up(lg) {
+      border-top: none;
     }
   }
+  // End mod
 }
--- a/site/assets/scss/_footer.scss
+++ b/site/assets/scss/_footer.scss
@@ -2,14 +2,4 @@
 // Footer
 //
 
-.bd-footer {
-  // Boosted mod
-  .bd-footer-info {
-    border-top: var(--#{$prefix}accordion-border-color) solid var(--#{$prefix}accordion-border-width);
-
-    @include media-breakpoint-up(lg) {
-      border-top: none;
-    }
-  }
-  // End mod
-}
+// Boosted mod: no .bd-footer
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -19,6 +19,8 @@
                      aria-describedby="headingLinks">Docs</a></li>
               <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/"
                      aria-describedby="headingLinks">Examples</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/guidelines/"
+                     aria-describedby="headingLinks">Design guidelines</a></li>
               <!-- Boosted mod: removed .Site.Params.themes and .Site.Params.blog -->
             </ul>
           </div>
@@ -37,8 +39,6 @@
 
               <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/getting-started/"
                      aria-describedby="headingGuides">Getting started</a></li>
-              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/starter-template/"
-                     aria-describedby="headingGuides">Starter template</a></li>
               <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/getting-started/webpack/"
                      aria-describedby="headingGuides">Webpack</a></li>
               <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/getting-started/parcel/"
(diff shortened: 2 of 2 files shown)


site/layouts/partials/footer.html
```diff
@@ -1,56 +1,113 @@
...
-          <li class="list-inline-item mt-3 mt-md-0"><a href="https://www.netlify.com">
-            <img src="https://www.netlify.com/img/global/badges/netlify-dark.svg" alt="Deploys by Netlify" width="89" height="40"/>
-          </a></li>
-        </ul>
+<footer class="footer bg-dark navbar-dark">
+  <h2 class="visually-hidden">Boosted sitemap & information</h2>
+  <div class="container-xxl footer-nav">
+    <nav class="accordion accordion-dark" id="sitemapAccordion" aria-label="Boosted sitemap footer">
+      <div class="row">
+        <div class="footer-column col-md-2 offset-md-1">
```
louismaximepiton, line 6: I think the `offset-md-1` could be removed here or moved to the last column

site/layouts/partials/footer.html
```diff
@@ -1,56 +1,113 @@
...
+              <li><a class="nav-link" href="/" aria-describedby="headingLinks">Home</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/"
+                     aria-describedby="headingLinks">Docs</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/"
+                     aria-describedby="headingLinks">Examples</a></li>
+              <!-- Boosted mod: removed .Site.Params.themes and .Site.Params.blog -->
+            </ul>
+          </div>
+        </div>
+        <div class="footer-column col-md-2">
```
louismaximepiton, line 26: We should maybe consider using `col-md-6 col-lg-2` since some links are not on one line -> Need to change the last paragraphs management

site/layouts/partials/footer.html
```diff
@@ -1,56 +1,113 @@
...
+          <div id="collapseCommunity" class="container-xxl accordion-collapse collapse" data-bs-parent="#sitemapAccordion">
+            <ul class="navbar-nav">
+              <li><a class="nav-link" href="{{ .Site.Params.repo }}/issues" aria-describedby="headingFour2">Issues</a>
+              </li>
+              <li><a class="nav-link" href="{{ .Site.Params.repo }}/discussions" aria-describedby="headingFour2">Discussions</a>
+              </li>
+            </ul>
+          </div>
+        </div>
+        <div class="footer-column col-md-3 mb-3 pt-3 pt-md-4 px-1 px-md-0 fw-bold small">
```
louismaximepiton, line 89: These paragraphs might be under the back-to-top button in some cases, we could maybe try to handle it in a different way ? + in small viewports, missing accordion border between these paragraphs and the last accordion item. Maybe we should improve the the footer CSS to handle this kind of cases ? (not sure about that)

louismaximepiton (changes requested):
Some visual issues + one question

site/layouts/partials/footer.html
hannahiss: Like we said, this problem with back to top must be processed in another issue

hannahiss: Border has been added in small viewports

louismaximepiton (changes requested):
This solution looks good to me !

louismaximepiton (approved):
As said before, LGTM!

site/assets/scss/_footer.scss
```diff
@@ -3,17 +3,13 @@
 //
 
 .bd-footer {
-  @include font-size(.875rem);
-  position: sticky;
-  top: 100vh;
-  font-weight: 700;
+  // Boosted mod
+  .bd-footer-info {
```
julien-deramond, line 7: Not possible to use `.accordion-header` and get rid of this new class and so get rid of the `.bd-footer` here (and in the HTML) as well?

site/layouts/partials/footer.html
```diff
@@ -1,56 +1,113 @@
...
+            <span class="d-none d-md-flex">Links</span>
+          </h3>
+          <div id="collapseLinks" class="container-xxl accordion-collapse collapse" data-bs-parent="#sitemapAccordion">
+            <ul class="navbar-nav">
+              <li><a class="nav-link" href="/" aria-describedby="headingLinks">Home</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/"
+                     aria-describedby="headingLinks">Docs</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/"
+                     aria-describedby="headingLinks">Examples</a></li>
+              <!-- Boosted mod: removed .Site.Params.themes and .Site.Params.blog -->
```
julien-deramond, line 22: We should add "Design guidelines" in this section.

site/layouts/partials/footer.html
```diff
@@ -1,56 +1,113 @@
...
+              Guides
+            </button>
+            <span class="d-none d-md-flex">Guides</span>
+          </h3>
+          <div id="collapseGuides" class="container-xxl accordion-collapse collapse" data-bs-parent="#sitemapAccordion">
+            <ul class="navbar-nav">
+
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/getting-started/"
+                     aria-describedby="headingGuides">Getting started</a></li>
+              <li><a class="nav-link" href="/docs/{{ .Site.Params.docs_version }}/examples/starter-template/"
```
julien-deramond, line 40: We should update /examples/starter-template or remove this link. We can remove it now and create an issue to try to reintroduce a Boosted version for this example page if necessary.

julien-deramond (changes requested):

site/assets/scss/_footer.scss
julien-deramond: Fixed via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1280/commits/bc8db3e877b684e034727d1b5ff0a3eaa8ba8512

Issue #1189, opened 2022-04-05
ODS > Active Focus


B3nz01d:
# Active Focus

## Requirement

**As a** user
**I want** to be able to use a keyboard arrow
**So That** I can select the various items in a page and that they are visually identified

[DSM reference](toBeUpdtaed)

## General description
A selection sould hoghligh all avalable components to understand that they are currently selected

#### Component
* all focus items are using the following color:
  * white background: orange accessible
  * black background: brand orange 
  * other background (using the supporting colors): black
* the focus outline is sitting 2px away from the edge of the texte or the contained item
* the focus line is 3px thikness
* on the drop down the opening arrow would stay in white (not use the black hoover state when selected)
* /!\ on the drop down selected items check if a white separator could be added in) - this could also be applied to the table selected row
* the focus should follow the form of the focus item
* the toggle switch is focusing on the entire item
* on the tag item the focus should be applied to the entire item and not specifically on the delete crosss
* the focus state should not resize the component itself
* the focus state can overlay on top of other items of the componants
* the focus state shouldn't be animated
* the focus state should appear on keyboard focus only

Link to the new wcag 2.2: https://www.w3.org/TR/WCAG22/

### Variants
* Drop down: the selected item should alow to have a white outline between the black selected item and the focus line
    * propose focus inside the list
    * propose focus with the focus outside le list  <- prefered option
* Stepper: is it possible to have the outline follow the exact step form or does it need to be a rectangle outline ???
* Lists: check the selected item gray background and the orange contrast, might need to use the white spacing here as well
* Images for carousel:
  *  carourousel navigation: should use a white outline + an orange focus
  *  pagination dots: the pill holding the dots would switch to full white background and the orange rounded focus is applyed to the size of the dot
* Global header icon with a badge: could the we make the focus go under the badge so that te notification number is still visible???
*  

### States
* NA

### Internationalization
* NA

### Interaction
* All: 
  * The user is able to select a text using the default OS behavior
* Touch: 
  * Nothing Specific
* Keyboard: 
  * Nothing specific

### Animations
* there should not be any animation of the focus with zoom effect - /!\ check if this has an impact from the bootstrap source

### Visual Accessibility
* taken into account within the design creation process

### Usage Accessibility
* (?) Focus order - NA
* (?) Focus loss expected behavior - NA
* (?) Screen Reader - default OS behavior

## Implementation tasks
- [x] #1190
- [ ] #1237

[...]

PR #1281, opened 2022-05-25
CSS var: stepped process

louismaximepiton:
CSS var for stepped process


scss/_stepped-process.scss
```diff
@@ -2,123 +2,141 @@
...
 .stepped-process-link {
-  flex: 1 0 $step-link-width;
-  max-width: $step-link-width;
+  flex: 1 0 var(--#{$prefix}stepped-process-link-width);
+  max-width: var(--#{$prefix}stepped-process-link-width);
   margin: auto;
   overflow: hidden;
-  line-height: 1;
-  color: $step-link-color;
+  line-height: 1; // not enough
```
louismaximepiton, line 94: Needs to be tackled in another PR ?

louismaximepiton (commented):
On hold until there's an answer for the second comment

scss/_stepped-process.scss
```diff
@@ -2,123 +2,141 @@
...
+  --#{$prefix}stepped-process-item-next-bg: #{$step-item-next-bg};
+  --#{$prefix}stepped-process-item-drop-shadow: none;
+  --#{$prefix}stepped-process-item-arrow-width: 0;
+  --#{$prefix}stepped-process-item-arrow-shape: #{$step-item-arrow-shape};
+  --#{$prefix}stepped-process-link-width: #{$step-link-width};
+  --#{$prefix}stepped-process-link-color: #{$step-link-color};
+  --#{$prefix}stepped-process-link-active-color: #{$step-link-active-color};
+  --#{$prefix}stepped-process-link-marker: #{$step-link-marker};
+  --#{$prefix}stepped-process-link-text-decoration: #{$step-link-text-decoration};
+  // scss-docs-end stepped-process-css-vars
```
julien-deramond, line 24: We didn't have the case before but shouldn't we include what's in `@include media-breakpoint-up` as well in the documentation inside some `scss-docs-end`? In the same as `stepped-process-css-vars` or in separate section explained in the docs.

scss/_stepped-process.scss
julien-deramond: Yes could you please create an issue about that and remove this comment?

scss/_variables.scss
```diff
@@ -1967,26 +1967,29 @@ $back-to-top-icon-height:        subtract(1rem, 1px) !default;
...
+$stepped-process-counter:     step !default; // Used as a counter name
+
+$step-item-padding:       .5rem !default;
+$step-item-margin-end:    $border-width !default;
+$step-item-bg:            $black !default;
+$step-item-active-bg:     $primary !default;
+$step-item-next-bg:       $gray-400 !default;
+$step-item-shadow-size:   $border-width * 1.5 !default;
+$step-item-drop-shadow:   drop-shadow($step-item-shadow-size 0 0 $white) #{"/* rtl:"} drop-shadow(-$step-item-shadow-size 0 0 $white) #{"*/"} !default;
+// $step-item-padding-end:  $step-item-padding * 2 !default;
```
julien-deramond, line 1982: If not used anymore we should use `// fusv-disable` and `// fusv-enable` to handle the deprecation of the var in v5.2.0 (and then remove it afterwards)

scss/_variables.scss
```diff
@@ -1967,26 +1967,29 @@ $back-to-top-icon-height:        subtract(1rem, 1px) !default;
...
+
+$step-item-arrow-width:   1rem !default;
+$step-item-arrow-shape:   polygon(0% 0%, subtract(100%, $border-width) 50%, 0% 100%) #{"/* rtl:"} polygon(100% 0%, $border-width 50%, 100% 100%) #{"*/"} !default; // Used in clip-path
+
+$step-link-width:         1.25ch !default; // Matches width of a single number
+$step-link-color:         $white !default;
+$step-link-active-color:  $black !default;
+$step-link-marker:        counter(var(--bs-stepped-process-counter)) inspect("\A0") !default;
+$step-link-marker-lg:     counter(var(--bs-stepped-process-counter)) inspect(".\A0") !default;
+$step-link-text-decoration: $link-decoration !default;
```
julien-deramond, line 1992: All `$step-*` and `$stepped-*` values should be vertically aligned in the file.

julien-deramond (changes requested):
I haven't finished the technical review (CSS vars) but here are some first feedback.

I think the result observed in this PR is better but is it normal that the visible focus has changed? Not at all breakpoints but the shape is sometimes different (smaller).

scss/_stepped-process.scss
louismaximepiton: #1337

julien-deramond:
@louismaximepiton Reverted the creation of `--#{$prefix}stepped-process-max-items` via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1281/commits/c94be7ba24d8132e1da58170a90e0c1659a5c602 because of `@for $i from 1 through $stepped-process-max-items {`. Indeed you won't be able to change the behavior of this loop by changing only the CSS var and having the value stored in this CSS var doesn't seem very useful in terms of reuse. We can discuss about that when you're back if I missed something and you want it to be reintroduced.

julien-deramond (approved):
LGTM! Visible focus has changed but will be revisited via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1189.

PR #1340, opened 2022-06-27
Card cap color fix

### LINKED ISSUE #1302: Usage of `--bs-card-cap-color` in card footer
### Prerequisites

### Describe the issue

Modifying `--bs-card-cap-color` to `#fd0` give the following rendering:

Screenshot from 2022-06-09 13-32-24

The card footer background shouldn't be yellow. We should use `--bs-card-cap-bg` or have a specific CSS var for this use case.

### Reduced test cases

The change can be made manually in the browser.

Linux

Firefox

v5.2.0

louismaximepiton:
Closes #1302 

I had a hard time on this issue since we don't have spec for this kind of element. I tried to keep it as-is. So I moved to `--bs-card-bg` instead of `--bs-card-cap-bg` since the last one might change the aspect of the card. I was wondering if the background of the card footer could be transparent or could inherit from card itself but the actual solution looks more flexible to me. I don't know if you have any other ideas ?


julien-deramond (approved):
Since we don't have Orange design specifications for this use case and that's not a priority atm, let's go with this patch which is more logical.

PR #1361, opened 2022-07-06
Fix the background color for validated input

### LINKED ISSUE #1358: Input text field background-color changing when valid
### Prerequisites

### Describe the issue

Background color of input text gets transparent when input is valid.
Background of input should remain White but gets transparent when valid.

### Reduced test cases

Reproducible on boosted and bootstrap demo sites.
https://getbootstrap.com/docs/5.1/forms/validation/
https://getbootstrap.com/docs/5.1/forms/validation/

you need to set the background to any color to notice the bug. (background-color: #333;)
.bd-example {
    padding: 1.5rem;
    margin-right: 0;
    margin-left: 0;
    border-width: 1px;
    background-color: #333;
}

Windows

Chrome

5.1

louismaximepiton:
Fixes #1358 

Bs only change the background-image so we do the same.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1001,8 +1001,8 @@ $input-placeholder-color:               $gray-700 !default;
 $input-plaintext-color:                 var(--#{$prefix}body-color) !default;
 
 $input-height-inner:                    add($input-line-height * 1em, $input-padding-y * 2) !default;
-// Boosted mod: no $input-height-inner-half
-// Boosted mod: no $input-height-inner-quarter
+$input-height-inner-half:               $spacer !default; // Boosted mod
+$input-height-inner-quarter:            map-get($spacers, 2) !default; // Boosted mod
 
 $input-height:                          2.5rem !default;
 // Boosted mod: no $input-height-sm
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -1001,8 +1001,8 @@ $input-placeholder-color:               $gray-700 !default;
 $input-plaintext-color:                 var(--#{$prefix}body-color) !default;
 
 $input-height-inner:                    add($input-line-height * 1em, $input-padding-y * 2) !default;
-// Boosted mod: no $input-height-inner-half
-// Boosted mod: no $input-height-inner-quarter
+$input-height-inner-half:               $spacer !default; // Boosted mod
```
julien-deramond, line 1004: Why don't we use the same values as Bootstrap?
```scss
$input-height-inner-half:               add($input-line-height * .5em, $input-padding-y) !default;
$input-height-inner-quarter:            add($input-line-height * .25em, $input-padding-y * .5) !default;
```

julien-deramond: Let's keep it as you developed it for now. We'll tackle this issue for precisely when the design for form validation will be in the DSM and final.

julien-deramond (approved):

Issue #846, opened 2021-10-06
v5.2.0 Ship List


julien-deramond:
## :rocket: Features

- **Ratios**: New 9x16 aspect ratio to display portrait videos ([#906](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/906))
- **Utilities**: New classes (`.ll-sm` and `.ll-md`) to limit the extent of line lengths ([#883](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/883) / [#925](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/925))
- **Components**: New Orange navbar component ([#795](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/795) / [#1099](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1099))
- **Buttons**: Reintegration of the `.btn-inverse` class to handle buttons with dark background ([#1010](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1010)) ([#1064](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1064))
- **Forms**: New Star Rating component ([#887](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/887))
- **Spinners**: Reintegration of the growing spinners from v4 ([#1041](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1041))
- **Dark background**: Handle basic elements like links and code to be displayed on dark background automatically ([#1038](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1038))
- **Dark background**: Accordions dark variant ([#1025](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1025))
- **Dark background**: Alerts dark variant ([#1012](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1012))
- **Forms**: New component: Quantity Selector ([#860](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/860))
- **Components**: New component: Sticker ([#1023](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1023)) ([#1251](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1251))
[...]

PR #1406, opened 2022-07-25
Added in v5.2.0 label for all new features

julien-deramond:
Add an 'Added in v5.2.0' label for all of our new features.

Please check that nothing's missing (see https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/846 as reference).


site/content/docs/5.2/components/offcanvas.md
```diff
@@ -253,7 +253,7 @@ Since the offcanvas panel is conceptually a modal dialog, be sure to add `aria-l
 
 ### Variables
 
-<small class="d-inline-flex px-2 py-1 fw-semibold text-success bg-success bg-opacity-10 rounded-2">Added in v5.2.0</small>
+{{< added-in "5.2.0" >}}
```
isabellechanclou, line 256: The Added in v5.2.0 has a green (#EAF9EA) background instead of the grey (#CCCCCC) one => green color to be replaced by the grey one.

isabellechanclou (approved):

louismaximepiton (commented):
Wondering if the `added-in` should be on top of the page in `orange-navbar.md`.
In `carousel.md`, add a tag for play/pause button. (and add it in ship list ?)
In `check-radios.md`, add a tag for without labels part ?
In `borders.md`, add a tag for variables ? (maybe on Bs)
Do we need to add a tag on dark-offcanvas ? (on Bs side ?)

EDIT: masthead.html doesn't have the same code but it might not be an issue

julien-deramond:
> Wondering if the `added-in` should be on top of the page in `orange-navbar.md`. In `carousel.md`, add a tag for play/pause button. (and add it in ship list ?) In `check-radios.md`, add a tag for without labels part ? In `borders.md`, add a tag for variables ? (maybe on Bs) Do we need to add a tag on dark-offcanvas ? (on Bs side ?)
> 
> EDIT: masthead.html doesn't have the same code but it might not be an issue

- Orange Navbar didn't exist in v5.1.3 so it must be on top of the page
- Carousel with pause/play button. Yes I'm adding it (also in the ship list, IDK why it isn't there)
- Check & Radios without labels comes from Bootstrap so let's let it as it is
- Same thing for Borders CSS variables (Bootstrap)
- Same thing for Dark OffCanvas (Bootstrap)

Issue #1326, opened 2022-06-22
Docs > Components > Buttons > Sizes: Fix Alert/Warning message 


isabellechanclou:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

This issue comes from the work done in Issue https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1201 List in all the doc the incompatibility with ODS in which all Boosted components (variants included) have been inspected in order to identify the ones that do not respect ODS specifications and therefore need an alert message saying "Do not use in ODS".
Doing this work, design bugs were noticed.

**Design bug to be corrected:**
- docs/v5.1/components/buttons/sizes: replace the sentence "It is not recommended by Orange Design System..." by the dedicated Alert/Warning message.

![image](https://user-images.githubusercontent.com/89514509/175004937-5c9769ed-50ee-494a-aac9-6322c693b9d5.png)

Dedicated Alert/Warning message to use:
![image](https://user-images.githubusercontent.com/89514509/175005106-ac113285-12fe-4e3b-b75c-1ab117024c96.png)

### Reduced test cases

NA

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What version of Boosted are you using?

v5.1

isabellechanclou:
I'm closing this issue because it will be treated in issue #1199.

Issue #1199, opened 2022-04-07
Add callouts in all the documentation when incompatibility with ODS


julien-deramond:
Some components (or some variants of them) are inherited from Bootstrap and should not be used in Orange websites.
To avoid an extra maintenance or simply to let the developer understand how some attributes or variants are possible, we keep what's from Bootstrap documentation.
So we use callouts as an information for the users to know that. This is an example for the Navbar component:
![Screenshot from 2022-04-07 16-23-06](https://user-images.githubusercontent.com/17381666/162222266-cb041597-0caf-483e-aaad-d248de2a8e9d.png)
ℹ️  All callouts must use this template which is in `site/layouts/shortcodes/ods-incompatibility-alert.html` file.
The description text of the callout needs to be written in the component .md file. It has to be adapted for each different case. Try as much as possible to use the same description texts for similar cases. 

This epic story will contain the list of pages and components that should have those callouts.

## List of pages or components
=> For the moment, there is a unique callout template with a "danger" alert icon. So do not take into account the fact that is it written below "Add **warning**" alert or "Add **danger** alert".

   💡 **To discuss with the design**:
Instead of using the Alert/Warning icon, it might be more efficient and clear to have an icon representing ODS or at least letting know that this callout targets a design topic and not a technical one.
This idea was implemented in Boosted V4.3 with this message on the Introduction page introducing these Docs callouts:
        
> ![image](https://user-images.githubusercontent.com/89514509/183386978-a0fca654-9b6c-4a66-9002-de33b30f2650.png)

and the use of the icon where needed in the Docs pages:
        
> ![image](https://user-images.githubusercontent.com/89514509/183386626-f9bae466-cfac-4053-9059-960bafc669d1.png)

        => **Decided on this topic with the design** : 
                - Keep the Alert icon which is the most efficient
                - As it was done in v4.3, add an explanation text in the `Getting started / Introduction` page to introduce and explain thje why of our design callout messages.
                - 2 possible contents for these callout messages where tested: 
                a) detailed messages to explain why a component or the variant can not be used (`kept version`).
                b) a generic and simpler text proposed by Cyriaque:  "In its structure or in its visual, this component variant inherited from Bootstrap is incompatible with the Orange Design System." The callout title remains the same "Incompatibility with ODS" with a link on it to go to the introduction page.

=>  **In the  ℹ️  description text proposals below, the ones in bold are the one put in the code at the moment.**

[...]

PR #1414, opened 2022-07-26
Fix warning message

MewenLeHo:
Closes [#1326 ](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1326)

### DIFF
--- a/site/content/docs/5.2/components/buttons.md
+++ b/site/content/docs/5.2/components/buttons.md
@@ -211,7 +211,7 @@ Fancy larger or smaller buttons? Add `.btn-lg` or `.btn-sm` for additional sizes
 {{< /example >}}
 
 {{< ods-incompatibility-alert >}}
-in some cases you could roll your own custom sizing with CSS variables like in the example just under this alert.
+In some cases you could roll your own custom sizing with CSS variables like in the example just under this alert.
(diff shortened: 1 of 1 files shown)


site/content/docs/5.2/components/buttons.md
```diff
@@ -210,7 +210,9 @@ Fancy larger or smaller buttons? Add `.btn-lg` or `.btn-sm` for additional sizes
 <button type="button" class="btn btn-secondary btn-sm">Small button</button>
 {{< /example >}}
 
-**It is not recommended by Orange Design System** but in some cases you could even roll your own custom sizing with CSS variables:
+{{< ods-incompatibility-alert >}}
+In some cases you could roll your own custom sizing with CSS variables like in the example just under this alert.
```
isabellechanclou, line 214: To be more consistent with the other callout messages as written so far (see PR #1199), it would propose to write "This variant is not recommended by the Orange Design Specifications, nevertheless, in some cases, you could roll your own custom sizing with CSS variables."
Now, as there a topic to raise about the callout messages, it may be wiser to wait before considering changing this wording...

isabellechanclou:
I closed the issue #1326 because it will be treated in issue #1199 instead. Therefore, we don't need this PR anymore. So, I'm closing it.

PR #1417, opened 2022-07-27
List group: Trim unused CSS var

louismaximepiton:


scss/_variables.scss
```diff
@@ -1739,7 +1739,7 @@ $list-group-border-radius:          $border-radius !default;
 $list-group-item-padding-y:         .875rem !default;
 $list-group-item-padding-x:         $spacer !default;
 
-$list-group-hover-bg:               null !default; // Boosted mod
+// Boosted mod: no $list-group-hover-bg:
```
julien-deramond, line 1742: You might want use `// fusv-disable` to deprecate the Sass variable

julien-deramond (approved):
LGTM! IMO it is safe to remove it like that without handling it as deprecated since it was never used in the past.

PR #1418, opened 2022-07-27
Modal: Remove unused CSS variables

louismaximepiton:


julien-deramond:
Maintenance question: isn't it "better" to keep the variables and set them to `$btn-close-padding`? It would be closer to Bootstrap.
```scss
.btn-close {
    margin: calc(#{$btn-close-padding} * -1) calc(#{$btn-close-padding} * -1) calc(#{$btn-close-padding} * -1) auto; // Boosted mod: using custom properties for padding
  }
```

louismaximepiton:
That's a good question and I don't have the answer right now. Maybe it's better, I'll give it a try !

julien-deramond (approved):
Let's do it like this for the moment. It is consistent to other files: `_accordion.scss`, `_list-group.scss` and `_navbar.scss`.

Issue #1332, opened 2022-06-23
Docs > Components > Navs-tabs: Fix hover color on a selected item


isabellechanclou:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

This issue comes from the work done in Issue https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1201 List in all the doc the incompatibility with ODS in which all Boosted components (variants included) have been inspected in order to identify the ones that do not respect ODS specifications and therefore need an alert message saying "Do not use in ODS".
Doing this work, design bugs were noticed.

**Design bug to be corrected:**
- docs/v5.1/components/navs-tabs:
    - [x] **#tabs-light**: there shouldn't be a change of text color on hover, pressed, focused on a selected tab. Only hover and pressed states will to be fixed in this issue. The focus state will be fixed in issue #1237
    - [x] **#nested-tabs**: there shouldn't be a change of text color on hover, pressed, focused on a selected tab. Only hover and pressed states will to be fixed in this issue. The focus state will be fixed in issue #1237
![image](https://user-images.githubusercontent.com/89514509/175297494-f0573373-5497-47e9-aac9-dca7fa913472.png)

### Reduced test cases

NA

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What version of Boosted are you using?

v5.1

PR #1420, opened 2022-07-27
Remove hover for active tab

MewenLeHo:
Closes [#1332](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1332)


isabellechanclou (approved):
Both fixes ok for me.

julien-deramond:
It is said in the issue not to take into account focus modifications here. But still there is a regression when an active item is focused:
Capture d’écran 2022-07-29 à 13 50 12

It should remain orange.

scss/_nav.scss
```diff
@@ -127,7 +127,7 @@
     // stylelint-disable-next-line function-disallowed-list
     border-width: 0 0 calc(var(--#{$prefix}nav-tabs-border-width) * 4);
 
-    &:hover,
+    &:not(.active):hover,
```
julien-deramond, line 130: IDK if this rule is useful.

MewenLeHo: It was indeed useful for the previous version of the issue.
I had to revamp it a bit for the updated version.

scss/_nav.scss
```diff
@@ -135,6 +135,10 @@
       outline-color: currentcolor;
     }
 
+    &.active:hover {
+      color: var(--#{$prefix}nav-tabs-link-active-color);
+    }
+
```
isabellechanclou, line 141: OK with this.

 ⚠️ There is a remaining visual behavior to be fixed in PR #1437: on mouse down (active state) on a selected item, the text color stays black as expected, but when the click is released and the mouse is moved away, 
the text color turns orange when it shouldn't. It has to do with the focus management.

julien-deramond:
Regression: Active + focused tab item changes color (+ border color) when hovered. 

Peek 2022-08-24 14-13

louismaximepiton:
It should maybe be merged once the #1437 is merged ?

scss/_nav.scss
```diff
@@ -135,6 +135,10 @@
       outline-color: currentcolor;
     }
 
+    &.active:hover {
+      color: var(--#{$prefix}nav-tabs-link-active-color);
```
louismaximepiton, line 139: Why not writing this line inside the `.active` below ?

julien-deramond (approved):
LGTM!

PR #1424, opened 2022-07-28
Alert: Adding CSS var

louismaximepiton:
Is there some variables that need to go in Bs ? Or maybe I revert the changes that aren't done in Bs yet ?

### Checks

- [ ] There's no regression between [deploy](https://deploy-preview-1424--boosted.netlify.app/docs/5.2/components/alerts/) and [main](https://main--boosted.netlify.app/docs/5.2/components/alerts/).


julien-deramond:
There's a conflict that needs to be resolved.
Issue with Bundlewatch as well.

julien-deramond:
@louismaximepiton Some conflicts to resolve here 🙏

scss/_alert.scss
```diff
@@ -39,7 +49,7 @@
 
 // Provide class for links that match alerts
 .alert-link {
-  font-weight: $alert-link-font-weight;
+  font-weight: var(--#{$prefix}alert-link-font-weight);
```
julien-deramond, line 52:
```suggestion
  font-weight: var(--#{$prefix}alert-link-font-weight); // Boosted mod
```

scss/_alert.scss
```diff
@@ -48,14 +58,15 @@
 // Expand the right padding and account for the close button's positioning.
 
 .alert-dismissible {
-  padding-right: $alert-dismissible-padding-r;
+  padding-right: var(--#{$prefix}alert-dismissible-padding-right);
```
julien-deramond, line 61:
```suggestion
  padding-right: var(--#{$prefix}alert-dismissible-padding-right); // Boosted mod
```

scss/_alert.scss
```diff
@@ -85,31 +96,34 @@
...
   line-height: 1.5;
 }
 
 //// Icons
 .alert-icon {
   flex-shrink: 0;
-  width: $alert-icon-size;
-  margin: -#{$alert-icon-margin-y * .5} 0;
+  width: var(--#{$prefix}alert-icon-size);
+  margin: calc(var(--#{$prefix}alert-icon-margin-y) * -.5) 0; // stylelint-disable-line function-disallowed-list
```
julien-deramond, line 106:
```suggestion
  margin: calc(-.5 * var(--#{$prefix}alert-icon-margin-y)) 0; // stylelint-disable-line function-disallowed-list
```

scss/_alert.scss
```diff
@@ -8,17 +8,27 @@
...
   --#{$prefix}alert-color: #{$alert-color}; // Boosted mod
   --#{$prefix}alert-border-color: transparent;
   --#{$prefix}alert-border: #{$alert-border-width} solid var(--#{$prefix}alert-border-color);
   --#{$prefix}alert-border-radius: #{$alert-border-radius};
+  --#{$prefix}alert-logo-size: #{$alert-logo-size}; // Boosted mod
+  --#{$prefix}alert-icon-size: #{$alert-icon-size}; // Boosted mod
+  --#{$prefix}alert-icon-margin-y: #{$alert-icon-margin-y}; // Boosted mod
+  --#{$prefix}alert-link-font-weight: #{$alert-link-font-weight}; // Boosted mod
+  --#{$prefix}alert-heading-font-weight: #{$alert-heading-font-weight}; // Boosted mod
+  --#{$prefix}alert-dismissible-padding-right: #{$alert-dismissible-padding-r}; // Boosted mod
```
julien-deramond, line 22: Since the button is positioned with absolute values, it seems that `$alert-dismissible-padding-r` becomes useless, and so is this CSS variable.
If we confirm that, I will create a PR for Bootstrap since it is probably the same over there.

scss/_alert.scss
```diff
@@ -119,26 +133,17 @@
...
+  --#{$prefix}alert-line-height: #{$line-height-sm};
   --#{$prefix}alert-border: 0;
+  --#{$prefix}alert-logo-size: #{$alert-logo-size-sm};
+  --#{$prefix}alert-icon-size: #{divide($alert-icon-size, 2)};
+  --#{$prefix}alert-btn-close-offset: #{$alert-btn-close-offset-sm};
   // scss-docs-end alert-sm-css-vars
-
-  font-size: $font-size-sm;
-  line-height: $line-height-sm;
+  --#{$boosted-prefix}icon-spacing: #{$btn-close-padding-sm};
```
julien-deramond, line 143: Can we have button close with small alerts? If not, I'm not sure this CSS var is useful.

scss/_alert.scss
```diff
@@ -119,26 +133,17 @@
   --#{$prefix}alert-padding-x: 0;
   --#{$prefix}alert-padding-y: #{$alert-padding-sm};
   --#{$prefix}alert-margin-bottom: 0;
+  @include rfs($font-size-sm, --#{$prefix}alert-font-size);
+  --#{$prefix}alert-line-height: #{$line-height-sm};
   --#{$prefix}alert-border: 0;
+  --#{$prefix}alert-logo-size: #{$alert-logo-size-sm};
+  --#{$prefix}alert-icon-size: #{divide($alert-icon-size, 2)};
```
julien-deramond, line 140: Why do we use a division here instead of a multiplication?

julien-deramond (changes requested):

scss/_alert.scss
louismaximepiton: I think it's to avoid long text to be under the close button.

scss/_alert.scss
louismaximepiton: IMO, this should be tackled inside another PR. That said, I think that every alerts should be dismissible (even if we don't show it on our doc).

julien-deramond: OK you can let this modification in the PR but let's create an issue to track it

scss/_alert.scss
julien-deramond: So reading the new version, do we need a `$alert-icon-size-sm` to be consistent?

Issue #1324, opened 2022-06-22
Docs > Components > Alerts : Add a close button + fix height


isabellechanclou:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

This issue comes from the work done in Issue https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1201 List in all the doc the incompatibility with ODS in which all Boosted components (variants included) have been inspected in order to identify the ones that do not respect ODS specifications and therefore need an alert message saying "Do not use in ODS".
Doing this work, design bugs were noticed.

**Design bug to be corrected:**
- docs/v5.1/components/alerts:
    - ~~**#examples**: Add a close button (black cross) on the right for each example~~
    - [x] **#additional-content**: 
        - ~~Add a close button (black cross) on the right for each example~~
        - [x] The component height is incorrect. It is 75.59px while it should be 80px
    - ~~**dark-variant**: Add a close button (white cross) on the right for each example~~

### Reduced test cases

NA

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What version of Boosted are you using?

v5.1

MewenLeHo:
"dark-variant: Add a close button (black cross) on the right for each example"

@isabellechanclou: shouldn't close buttons be white for dark variants?

isabellechanclou:
Yes, of course 👍. I do the correction in the above issue description.

MewenLeHo:
@isabellechanclou: as discussed during the daily and according to @julien-deramond, we are not supposed to have a close button everywhere.

Could we clarify that?

PR #1425, opened 2022-07-28
Fix alerts height

MewenLeHo:
Closes [#1324](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1324)


scss/_alert.scss
```diff
@@ -31,6 +31,7 @@
   margin: 0;
   font-size: inherit;
   font-weight: inherit;
+  line-height: 1.125;
```
louismaximepiton, line 34: Not sure about that, but since we inherit font-size, maybe we should inherit line-height too ?

scss/_alert.scss
```diff
@@ -85,6 +86,7 @@
 
 .alert-heading ~ p {
   font-weight: $font-weight-normal;
+  line-height: 1.5rem;
```
louismaximepiton, line 89: `1.5` should be better here to be responsive on font-size

louismaximepiton (commented):
Few things

louismaximepiton (approved):
LGTM!

julien-deramond (approved):

PR #1426, opened 2022-07-28
Buttons: Adding CSS var

louismaximepiton:
One question remains: Do we need to use `button-variant` in `.btn-no-outline` and `.btn-social` ?
I did it the other way (writting down all the variables) because from what I've seen, it lightens the bundle but I'm open minded on every argument.

It can be reviewed commit by commit if prefered and/or can be splited in different PRs.


scss/_buttons.scss
```diff
@@ -150,27 +150,19 @@
 
 // Boosted mod: border-only on :hover, :focus and :active
 .btn-no-outline {
```
louismaximepiton, line 152: Hard to know if this is a good way to process or if a button-variant could the trick.

scss/_buttons.scss
```diff
@@ -232,23 +232,28 @@
 
 // Boosted mod: social button
 .btn-social {
-  border-color: currentcolor;
+  --#{$prefix}btn-border-color: currentcolor;
+  --#{$prefix}btn-hover-color: #{$white};
+  --#{$prefix}btn-hover-bg: var(--#{$boosted-prefix}network-color, $black);
+  --#{$prefix}btn-hover-border-color: var(--#{$boosted-prefix}network-color, $black);
+  --#{$prefix}btn-active-color: #{$white};
+  --#{$prefix}btn-active-bg: #{$black};
```
isabellechanclou, line 244: Why not:
- `btn-hover-bg-color` instead of `btn-hover-bg`?
- `btn-active-bg-color` instead of `btn-active-bg`?

scss/_buttons.scss
```diff
@@ -232,23 +232,28 @@
...
+  --#{$prefix}btn-active-border-color: #{$black};
+  --#{$prefix}btn-disabled-color: #{$gray-500};
+  --#{$prefix}btn-disabled-bg: transparent;
```
isabellechanclou, line 247: Why not `btn-disabled-bg-color` instead of `btn-disabled-bg`?

scss/_buttons.scss
```diff
@@ -232,23 +232,28 @@
...
 
-  &:active,
-  &.active {
-    color: $white;
-    background-color: $black;
-    border-color: $black;
+  &.btn-inverse {
+    --#{$prefix}btn-color: #{$white};
+    --#{$prefix}btn-active-color: #{$black};
+    --#{$prefix}btn-active-bg: #{$white};
```
isabellechanclou, line 258: Why not `btn-active-bg-color` instead of `btn-active-bg`?

isabellechanclou:
1st commit : No need of CSS vars for `$link-decoration`and `$link-hover-decoration`?

scss/_buttons.scss
louismaximepiton: Because we reuse the CSS var coming from `.btn`. Since they come from Bootstrap, I'd say that it is smaller so more understandable 🤔

scss/_buttons.scss
louismaximepiton: Same as above.

scss/_buttons.scss
louismaximepiton: Same as above.

louismaximepiton:
> 1st commit : No need of CSS vars for `$link-decoration` and `$link-hover-decoration`?

It was done this way in order to be closer to Bs !

julien-deramond (approved):
Just a comment regarding an apparently useless CSS var redefinition. Haven't looked at all the CSS vars redefinitions in details but there's no regressions so that's good to me.

PR #1427, opened 2022-07-28
Navs: Adding CSS var

louismaximepiton:
WIP


scss/_nav.scss
```diff
@@ -225,7 +231,7 @@
   // Boosted mod
   > .tab-pane-with-nested-tab {
     display: none;
-    margin: $spacer * -1 calc(subtract($nav-tabs-link-padding-x, $nav-tabs-border-width) * -1); // stylelint-disable-line function-disallowed-list
+    margin: calc(var(--#{$prefix}tab-content-padding-y) * -1) calc(var(--#{$prefix}tab-content-padding-x) * -1); // stylelint-disable-line function-disallowed-list
```
isabellechanclou, line 234: OK for me but makes the whole thing a bit harder/longer to read/understand...

isabellechanclou (approved):
LGTM (both commits) even though a slight comment on the 2nd commit.

scss/_nav.scss
julien-deramond: :warning: `calc(var(--#{$prefix}tab-content-padding-y) * -1) ` -> `calc(-1 * var(--#{$prefix}tab-content-padding-y))` (check all the PR)

julien-deramond (commented):
Sounds good to me. Just a comment regarding the naming of one of the CSS vars.

julien-deramond (approved):

Issue #891, opened 2021-11-09
Local Headers > Breadcrumb


julien-deramond:
## References

- **Orange Design System**: [Modules > Local Headers](https://system.design.orange.com/0c1af118d/p/774477-local-headers/b/743cd0):
  * Breadcrumb On White - WEB-LOC-BCR-001
  * Breadcrumb On Black - WEB-LOC-BCR-001
  * Breadcrumb On Supporting Colour - WEB-LOC-BCR-003

## More info

Two cents from @ffoodd:

Well, I guess it already kind of work: to fully support colored background, I think we'd need to add some color inheritance to breadcrumb items.

```css 
.breadcrumb-item a {
  color: inherit;
}
```

Also the separator should get that color too, so it'd need some `mask-image` love: 
1. set `content` to an empty string (`""`)
2. apply the SVG in `var()` as a `mask-image` (and probably set `mask-size` and `mask-position` alongside)
3. inherit color as background, thanks to `background-color: currentColor;`

Should do the trick! See Back-to-top link or social buttons for examples.

- Tweak breadcrumbs to use `mask` and inherit `color`, so that `.bg-*` utilities work on them.

louismaximepiton:
ODS Breadcrumb won't move anymore unless for fixes, let's close it.

PR #1430, opened 2022-07-29
Breadcrumb: Add dark variant

louismaximepiton:
Not sure if this is the best solution but this is likely the smallest one since we define the color to inherit (cascading from `*-dark` rule) and background-color to transparent + active state of links are linked to `--#{$prefix}link-hover-color`. 
Related to #891.

### Checks

Please check that [dark breadcrumbs](https://deploy-preview-1430--boosted.netlify.app/docs/5.2/components/breadcrumb/#dark-variant) are fine.

### DIFF
--- a/scss/_breadcrumb.scss
+++ b/scss/_breadcrumb.scss
@@ -46,3 +46,7 @@
     color: var(--#{$prefix}breadcrumb-item-active-color);
   }
 }
+
+.breadcrumb-dark .breadcrumb-item::before {
+  filter: $invert-filter;
+}
--- a/scss/_breadcrumb.scss
+++ b/scss/_breadcrumb.scss
@@ -52,10 +52,10 @@
 // Boosted mod
 .breadcrumb-dark {
   // scss-docs-start breadcrumb-dark-css-vars
-  --#{$prefix}breadcrumb-color: #{$breadcrumb-dark-color};
-  --#{$prefix}breadcrumb-bg: #{$breadcrumb-dark-bg};
-  --#{$prefix}breadcrumb-divider-color: #{$breadcrumb-dark-divider-color};
-  --#{$prefix}breadcrumb-item-active-color: #{$breadcrumb-dark-active-color};
+  --#{$prefix}breadcrumb-color: #{$breadcrumb-dark-color}; // Boosted mod
+  --#{$prefix}breadcrumb-bg: #{$breadcrumb-dark-bg}; // Boosted mod
+  --#{$prefix}breadcrumb-divider-color: #{$breadcrumb-dark-divider-color}; // Boosted mod
+  --#{$prefix}breadcrumb-item-active-color: #{$breadcrumb-dark-active-color}; // Boosted mod
   // scss-docs-end breadcrumb-dark-css-vars
 
   .breadcrumb-item::before {
--- a/scss/_breadcrumb.scss
+++ b/scss/_breadcrumb.scss
@@ -52,7 +52,10 @@
 // Boosted mod
 .breadcrumb-dark {
   // scss-docs-start breadcrumb-dark-css-vars
-  --#{$prefix}breadcrumb-color: #{$breadcrumb-dark-color}; // Boosted mod
+  --#{$prefix}breadcrumb-color: #{$breadcrumb-dark-color};
+  --#{$prefix}breadcrumb-bg: #{$breadcrumb-dark-bg};
+  --#{$prefix}breadcrumb-divider-color: #{$breadcrumb-dark-divider-color};
+  --#{$prefix}breadcrumb-item-active-color: #{$breadcrumb-dark-active-color};
   // scss-docs-end breadcrumb-dark-css-vars
 
   .breadcrumb-item::before {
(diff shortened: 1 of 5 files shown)


isabellechanclou:
LGTM

isabellechanclou (approved):

site/content/docs/5.2/components/breadcrumb.md
```diff
@@ -90,6 +90,20 @@ You can also remove the divider setting `--bs-breadcrumb-divider: '';` (empty st
 $breadcrumb-divider: none;
 ```
 
+## Dark variant
+
+Breadcrumbs come with a dark variant: `.breadcrumb-dark`.
+
```
julien-deramond, line 97: If you change it to `bg-warning` for example, the dark variant doesn't have the correct colors:
Screenshot from 2022-08-24 13-15-46

julien-deramond (changes requested):

site/content/docs/5.2/components/breadcrumb.md
louismaximepiton: Ah yes nice catch, baited by the utility once again >< Correcting it!

scss/_breadcrumb.scss
```diff
@@ -5,6 +5,7 @@
   --#{$prefix}breadcrumb-margin-bottom: #{$breadcrumb-margin-bottom};
   @include rfs($breadcrumb-font-size, --#{$prefix}breadcrumb-font-size);
   --#{$prefix}breadcrumb-font-weight: #{$breadcrumb-font-weight}; // Boosted mod
+  --#{$prefix}breadcrumb-color: #{$breadcrumb-color}; // Boosted mod
```
louismaximepiton, line 8: Should maybe be in Bootstrap ?

site/content/docs/5.2/components/breadcrumb.md
```diff
@@ -106,6 +124,20 @@ As part of Boosted's evolving CSS variables approach, breadcrumbs now use local
...
 
+<!-- Boosted mod: Dark variant -->
+Customization through CSS variables can be seen on the `.breadcrumb-dark` modifier class where we override specific values without adding duplicate CSS selectors.
+
+{{< scss-docs name="breadcrumb-dark-css-vars" file="scss/_breadcrumb.scss" >}}
+<!-- End mod -->
+
 ### Sass variables
 
+Variables for all breadcrumbs:
```
louismaximepiton, line 135: Should maybe be in Bs (in all files)

julien-deramond: Let's see the feedback from https://github.com/twbs/bootstrap/pull/37034

scss/_breadcrumb.scss
```diff
@@ -46,3 +48,15 @@
...
 }
+
+// Boosted mod
+.breadcrumb-dark {
+  // scss-docs-start breadcrumb-dark-css-vars
+  --#{$prefix}breadcrumb-color: #{$breadcrumb-dark-color}; // Boosted mod
+  // scss-docs-end breadcrumb-dark-css-vars
+
+  .breadcrumb-item::before {
+    filter: $invert-filter;
```
julien-deramond, line 59: Isn't there a way to do something like `--#{$prefix}breadcrumb-divider-color: #{$breadcrumb-dark-divider-color}` rather than using an invert filter?

In the same spirit, if we change `--bs-breadcrumb-divider-color` value, isn't there a way to make it applied to http://boosted.orange.com/docs/5.2/components/breadcrumb/#example as well? It is difficult to understand that `--bs-breadcrumb-divider-color` is only applied for http://boosted.orange.com/docs/5.2/components/breadcrumb/#dividers.

scss/_variables.scss
```diff
@@ -1806,6 +1806,7 @@ $breadcrumb-padding-y:              .5rem !default;
 $breadcrumb-padding-x:              0 !default;
 $breadcrumb-item-padding-x:         $spacer * .5 !default;
 $breadcrumb-margin-bottom:          1rem !default;
+$breadcrumb-color:                  $black !default; // Boosted mod
```
julien-deramond, line 1809: Not to do in this PR but we should think of a refactoring in the future by using `$body-color` instead of `$black` directly. For dark variant, we could introduce a `$body-dark-color`.

scss/_breadcrumb.scss
louismaximepiton: `--#{$prefix}breadcrumb-divider-color` is on Bs side and is only used for non-svg dividers. As said together, we should maybe try to tweak the variable name on BS side to be more consistent with what the variable does.

However, it doesn't work at the moment anyway, so I'll handle that with the next commit.

scss/_variables.scss
```diff
@@ -1843,6 +1844,13 @@ $breadcrumb-divider-flipped:        $breadcrumb-divider !default;
 $breadcrumb-border-radius:          null !default;
 // scss-docs-end breadcrumb-variables
 
+// scss-docs-start breadcrumb-dark-variables
+$breadcrumb-dark-color:             $white !default; // Boosted mod
+$breadcrumb-dark-bg:                null !default; // Boosted mod
+$breadcrumb-dark-divider-color:     $black !default; // Boosted mod: since the divider is inverted for svg reasons
```
julien-deramond, line 1850:
```suggestion
$breadcrumb-dark-divider-color:     $black !default; // Boosted mod: since the divider is inverted for SVG reasons
```

PR #1433, opened 2022-07-29
Pagination: Adding dark variant

louismaximepiton:
First draft idea: Add Scss var for each CSS var


site/content/docs/5.2/components/pagination.md
```diff
@@ -141,6 +141,26 @@ To avoid pagination wrapping or overflowing, we limit the number or pagination i
...
+
+{{< example class="bg-dark" >}}
+<nav aria-label="Dark page navigation example">
+  <ul class="pagination pagination-dark">
+    <li class="page-item disabled">
+      <a class="page-link">Previous</a>
+    </li>
+    <li class="page-item"><a class="page-link" href="#">1</a></li>
+    <li class="page-item"><a class="page-link" href="#">2</a></li>
+    <li class="page-item"><a class="page-link" href="#">3</a></li>
```
isabellechanclou, line 157: It could be interesting to add an `active` class on one of the 3 numbered items to illustrate the look of a selected item.

louismaximepiton: You're totally right, fixed via last commit !

isabellechanclou (approved):

scss/_variables.scss
```diff
@@ -1432,8 +1432,34 @@ $pagination-icon:                   var(--#{$boosted-prefix}chevron-icon) !defau
 $pagination-icon-size:              subtract($spacer * 2, $border-width * 2) !default;
 $pagination-icon-width:             add(.5rem, 1px) !default;
 $pagination-icon-height:            subtract(1rem, 1px) !default;
+
 $pagination-active-item-bg:         $primary !default;
+$pagination-active-item-color:      color-contrast($pagination-active-item-bg) !default;
 $pagination-active-item-border:     $pagination-active-item-bg !default;
+
+// Dark variant
+$pagination-dark-bg:                        $black !default;
```
julien-deramond, line 1441: In Bootstrap, CSS dark vars are separated in the docs (https://getbootstrap.com/docs/5.2/components/dropdowns/#variables). I find it interesting, we probably want to do the same.

julien-deramond (changes requested):
Rendering LGTM!
Requested some minor changes in the documentation.
Please inform the designers that the DSM should be updated: the hover state for dark variant isn't described!

site/content/docs/5.2/components/pagination.md
```diff
@@ -141,6 +141,28 @@ To avoid pagination wrapping or overflowing, we limit the number or pagination i
   </ul>
 </nav>
 {{< /example >}}
+
+## Dark variant
+
+Pagination comes with a dark variant: `.pagination-dark`.
```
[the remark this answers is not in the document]
louismaximepiton: I might have clicked a bit too fast, I used the [Accordion](https://boosted.orange.com/docs/5.2/components/accordion/#dark-variant) syntax. Which one would you prefer ?

scss/_variables.scss
louismaximepiton: Actually it seems that they don't have a specific method on this part regarding carousel, offcanvas, ... But I agree, we should propose it on their side !

site/content/docs/5.2/components/pagination.md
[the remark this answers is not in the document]
julien-deramond: The one I proposed is closer to what's in Bootstrap docs for dark variants. Something like this.
The Accordion syntax seems less correct.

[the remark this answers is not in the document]
louismaximepiton: Ok, I'll try to propagate the syntax everywhere !

site/content/docs/5.2/components/pagination.md
```diff
@@ -153,10 +175,20 @@ As part of Boosted's evolving CSS variables approach, pagination now uses local
 
 {{< scss-docs name="pagination-css-vars" file="scss/_pagination.scss" >}}
 
+Customization through CSS variables can be seen on the `.pagination-dark` modifier class where we override specific values without adding duplicate CSS selectors.
+
+{{< scss-docs name="pagination-dark-css-vars" file="scss/_pagination.scss" >}}
+
 ### Sass variables
 
+Variables for all pagination:
```
julien-deramond, line 184:
```suggestion
Variables for all paginations:
```

julien-deramond (approved):

PR #1437, opened 2022-08-02
Apply visible focus design guidelines in all components

### LINKED ISSUE #398: Enhance `:focus-visible` states color contrast
Our primary color `#f16e00` leads to insufficient contrasts against `#ffffff`.

Thanks to the `:focus-visible` polyfill, we're now able to distinguish the focus state when using keyboard—which already got enhanced with `outline`s.

I suggest we go a step further by ensuring contrasts (mostly using a darker orange) for components using `#f16e00` in either `color` or `background-color` when focused.

- [ ] inventory cases with `#f16e00` when focused
- [ ] select the minimum viable orange: for now, `$orange-700` which results in `#a44b00`, see [current `v5-dev` deploy](https://v5-dev--boosted.netlify.app/docs/5.0/customize/color/#all-colors)
- [ ] apply it in listed cases.

### LINKED ISSUE #659: Select focus state not visible enough
hello,
threre's an issue when the focus enter a <select> hml form, the focus status indicator is just displayed by a light gray to black 2px border. Even if it's enough for other f'orm controls (because the flashing cursor appears) it's not the case for a select.
We should increase the focus visiblity by, for example, changing the backgroàund ceolor of' the select or anything else...

### LINKED ISSUE #1035: Outline not visible on iOS and Safari on Star Rating
### Prerequisites

### Describe the issue

The outline for focus-visible isn't there for iOS and Mac. Maybe search for tabindex="0" or for options in MacOS directly.

### Reduced test cases

Go to [link](https://deploy-preview-887--boosted.netlify.app/docs/5.1/forms/checks-radios/#star-rating) and click on stars (or touch).

### What operating system(s) and browser(s) are you seeing the problem on?

(Source Browserstack)
* macOS + Safari
* iOS - Chrome

_main_ branch (5.2.0)

### LINKED ISSUE #1189: ODS > Active Focus

MewenLeHo:
Fixes #398.
Fixes #659.
Fixes #1035.
Fixes #1189.
Fixes #1189.

Commit that could be removed: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1437/commits/449d2c9e9d1651e6f9776045f2b5305afd2a56c7 (here to reduce bundle size ~.05KB once packed, but may reduce comprehension ?)


louismaximepiton:
Maybe you should try to have a `z-index: 3;` in your mixin in order to avoid problems having the same component next to other ones.

louismaximepiton:
@MewenLeHo could you please review [my fixes](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1437/commits/87ae05ffdfdab81eeeb8c649379164068a1a41dc) please ? 😄

louismaximepiton:
FYI, 

For stepped-process, the outline won't follow the shape of the arrows. Even if this is feasible, it appears as a draft and is too rounded to be branded. (Here is [the workaround](https://codepen.io/louismaximepiton/pen/oNqOEWG)).

Moreoverv, the way the stepped process is built leads to some overlapping with outline (or box-shadow) due to `z-index`. Each item create its own pile context so outline goes under the previous item.

The way I chose here is the default one (for links) without box-shadow because it could overlap on text and is very not readable. By default the outline is set to its color here.

louismaximepiton:
Should try to get closer to Bs once #1513 is merged: revert https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1513#issuecomment-1237748797. We should go to `:focus-visible` too!

louismaximepiton (changes requested):
We should move to `:focus-visible` and drop the `focus-visible` dependency. Implies to check that it's well implemented in all forms + check that we don't have any reference to it in scss files.

Should be rebased on main with new focus rule from Bootstrap!

Should we introduce Sass variable for each component redefining their own focus style ? any thoughts on this particular question @julien-deramond ?

On other files:
- [ ] Should refactor in `search.scss`
- [ ] Should refactor in `reboot.scss` the `[type="search"]` (Someone with Mac @isabellechanclou)
- [ ] In `button-group.scss`, l.137 is useful ?
- [x] Do we need rules on offset and focus inside `anchor.scss`
- [ ] In `buttons.scss`, check if we could revert https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1513#issuecomment-1237748797 once rebased
- [ ] In `variables.scss`:
  - [ ] `$input-btn-focus-width` should take `$outline-offset` value
  - [ ] `$input-btn-focus-box-shadow` should be almost the same as `$btn-focus-box-shadow`
- [ ] In `colored-links.scss`, we should remove `:focus` rule
- [ ] In `mixin/list-group` should remove `:focus`
- [ ] Is there anything to change on `.visually-hidden` ?
- [ ] In all files, see if we can remove easily the `:not(focus)` rules (in button-group)
- [ ] Check every focus rules inside `assets/scss/**.scss` (boosted, clipboard, navbar)
- [ ] [OUT OF PR] [Checks&radios](https://deploy-preview-1437--boosted.netlify.app/docs/5.2/forms/checks-radios/#with-icon) should have a different id.
- [ ] [OUT OF PR] [Carousel indicators](https://main--boosted.netlify.app/docs/5.2/components/carousel/#with-indicators) are broken if we hover for too long

louismaximepiton:
Need to see if the box-shadow could be changed on dark backgrounds : 
image

scss/_variables.scss
```diff
@@ -834,7 +833,7 @@ $input-btn-font-family:       inherit !default;
 $input-btn-font-size:         $font-size-base !default;
 $input-btn-line-height:       1.25 !default;
 
-$input-btn-focus-width:       .125rem !default;
+$input-btn-focus-width:       $outline-offset !default;
```
louismaximepiton, line 836: Boosted mod

scss/_variables.scss
```diff
@@ -875,7 +874,7 @@ $btn-border-width:            $input-btn-border-width !default;
 
 $btn-font-weight:             $font-weight-bold !default;
 $btn-box-shadow:              null !default;
-$btn-focus-width:             $border-width !default;
+$btn-focus-width:             $outline-offset !default;
```
louismaximepiton, line 877: We should use `$input-btn-focus-width` to be closer to Bootstrap

scss/_variables.scss
```diff
@@ -889,7 +888,7 @@ $btn-border-radius:           $border-radius !default;
 $btn-border-radius-sm:        $border-radius-sm !default;
 $btn-border-radius-lg:        $border-radius-lg !default;
 
-$btn-transition:              $transition-focus !default;
+$btn-transition:              null !default; // boosted mod
```
louismaximepiton, line 891: B

scss/_variables.scss
```diff
@@ -1515,11 +1514,20 @@ $accordion-body-padding-x:                $spacer !default; // Boosted mod
 $accordion-button-padding-y:              $accordion-padding-y !default;
 $accordion-button-padding-x:              $accordion-padding-x !default;
 $accordion-button-color:                  $accordion-color !default;
-$accordion-button-bg:                     var(--#{$prefix}accordion-bg) !default;
+$accordion-button-bg:                     transparent !default; // Boosted mod
 $accordion-transition:                    $btn-transition, border-radius .15s ease !default;
 $accordion-button-active-bg:              null !default; // Boosted mod
 $accordion-button-active-color:           $accessible-orange !default; // Boosted mod
 
+// Boosted mod: dark variant
```
louismaximepiton, line 1522: Might be a rebase conflict, they are already here below.

louismaximepiton (commented):
~~Package-lock.json shouldn't be commited maybe related to https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1521/commits/ff9ed90a2fff44129fdc55a51936ffd4b8ed8453~~

site/assets/scss/_anchor.scss
```diff
@@ -10,7 +10,6 @@
     content: "#";
   }
 
-  &:focus,
```
louismaximepiton, line 13: This one should stay IMO, otherwise we loose focus

scss/_variables.scss
```diff
@@ -1370,9 +1367,8 @@ $dropdown-header-color:             $black !default;
 $dropdown-header-padding-x:         $dropdown-item-padding-x !default;
 $dropdown-header-padding-y:         $spacer !default; // Boosted mod
 // fusv-disable
-$dropdown-header-padding:           $dropdown-header-padding-y $dropdown-header-padding-x !default; // Deprecated in v5.2.0
+$dropdown-header-padding:           $dropdown-header-padding-y $dropdown-header-padding-x !default; // Deprecated in v5.3.0
 // fusv-enable
-// scss-docs-end dropdown-variables
```
louismaximepiton, line 1375: Should be revert

scss/_variables.scss
```diff
@@ -1370,9 +1367,8 @@ $dropdown-header-color:             $black !default;
 $dropdown-header-padding-x:         $dropdown-item-padding-x !default;
 $dropdown-header-padding-y:         $spacer !default; // Boosted mod
 // fusv-disable
-$dropdown-header-padding:           $dropdown-header-padding-y $dropdown-header-padding-x !default; // Deprecated in v5.2.0
+$dropdown-header-padding:           $dropdown-header-padding-y $dropdown-header-padding-x !default; // Deprecated in v5.3.0
```
louismaximepiton, line 1370: Should be revert too

scss/_reboot.scss
```diff
@@ -90,17 +90,8 @@ body {
 //    See https://github.com/WICG/focus-visible
 // 3. Using the :focus-visible pseudo-class if supported by the browser
```
louismaximepiton, line 91: Should remove the explanation above

louismaximepiton (commented):
As we said: should we add a focus-box on button-group only (and not on every button) ?

We may handle the deploy preview, ~~any thoughts @julien-deramond ?~~ Sorry for the ping, Seems to be still imported in generate-sri file.

scss/_buttons.scss
```diff
@@ -50,25 +49,17 @@
...
-    outline-offset: $outline-width; // Boosted mod
-    &[data-focus-visible-added] { // Boosted mod: added `&[data-focus-visible-added]`
-      // Avoid using mixin so we can pass custom focus shadow properly
-      @if $enable-shadows {
-        box-shadow: var(--#{$prefix}btn-box-shadow), var(--#{$prefix}btn-focus-box-shadow);
-      } @else {
-        box-shadow: var(--#{$prefix}btn-focus-box-shadow);
-      }
-    }
+  &:focus-visible {
```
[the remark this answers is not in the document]
isabellechanclou: Yes, I think we should

scss/mixins/_focus.scss
```diff
@@ -0,0 +1,15 @@
...
+@mixin focus-visible($zindex: $outline-zindex, $color: $outline-color, $width: $outline-width, $offset: $outline-offset) {
+  z-index: $zindex;
+  outline: $color solid $width;
+  outline-offset: $offset;
+}
+// scss-docs-end focus-visible
+
+// scss-docs-start focus-box
+@mixin focus-box($offset: $outline-offset, $box-color: $outline-box-color) {
+  @if $enable-shadows {
```
isabellechanclou, line 11: This `@if $enable-shadows {` condition should be deleted in order to have a white circle between the carousel left/right arrow and its focus. Nevertheless, this deletion might break things somewhere else => to be investigated before deleting.

isabellechanclou:
In `docs/5.2/components/footer/#example` in small screen, the focus is a bit too narrow around the subcategory title => we should see how to can fix that.
image

julien-deramond:
Just a question here.
We propose to remove the focus-visible polyfill. Have you tried manually (or with BrowserStack; when it works) the old browsers we're supposed to support? (e.g. Firefox 68, etc.)

julien-deramond:
> > Just a question here. We propose to remove the focus-visible polyfill. Have you tried manually (or with BrowserStack; when it works) the old browsers we're supposed to support? (e.g. Firefox 68, etc.)
> 
> Yes, focus-visible is now well supported by all browsers except IE11: https://caniuse.com/?search=focus-visible

For Firefox 4-84, it is mentioned that it needs the prefix `-moz-`. That's why I'm asking if you've tried it for information.
On top of that, caniuse is a good source of information but can contain errors.

julien-deramond:
IDK if there's something to do here but Tags have been merged yesterday.
And btw, there are some conflicts around here

isabellechanclou:
In `Components > Navbar` for small screens, the hamburger menu (it's a `<button>`) is not centered inside the focus rectangle. It due to the padding of the `<button>` that is not symmetrical between left and right => See how to fix that to meet the design specifications.
image

isabellechanclou:
In `Components > Stepped process`, the look of the focus doesn't match the design design specifications (shape, color sometimes) in small and larger screens. Examples of the differences:
- in design specs: 
image
image

- in boosted:
image

    image

    image

isabellechanclou:
In Forms, is it normal that the focus for Text fields standard when "Active filled" and "Error" don't have the same look as in the design spec?
- In design specifications;
image
    image

- In Boosted > Forms > Validation (on MacOS):
image
image

louismaximepiton (approved):
LGTM

MewenLeHo:
**Focus meeting - Minutes**

- Stepped process: current focus is different from specifications because it is not possible to add the arrow parts in the focus. 
=> We decided that the actual version is ok so we will have to update design specifications.
=> In addition, we found an accessibility issue on this compoment that we will need to address in a specific PR: the current step must not be a link, we must replace it with plain text by removing the `href` attribute.

- Inputs: the problem here is that inputs can only be active and focused at the same time resulting in a possible edge case.
=> We decided to keep only the active border and to update the design specifications accordingly.

- Nav links: the 'current page' indicator can't be included in the focus on full size navbar but is included in the minimized version.
=> I will make a test with a smaller outline in order to widden the space between the bottom of the outline and the current page indicator.

julien-deramond:
What's the status here? PR still in Draft. It is ready for a final review or not yet?

Aniort (changes requested):
hello,
the orange #f16e00 focus indicator on white background is not enough contrasted
so, maybe, the solution is to put a black focus indicator on white (or light backgrounds) to be WCAG compliant!

scss/_carousel.scss
```diff
@@ -105,9 +105,9 @@
   opacity: $carousel-control-opacity;
```
louismaximepiton, line 105: Missing `box-shadow: none`

scss/_reboot.scss
```diff
@@ -91,7 +91,7 @@ body {
...
 // scss-docs-start focus-visibility
 :focus {
-  outline: $outline-width solid; // 1
+  @include focus-visible(); // 1
```
louismaximepiton, line 94: We should remove the box-shadow on click

scss/_variables.scss
```diff
@@ -584,8 +588,12 @@ $border-radius-2xl:           null !default;
 $border-radius-pill:          50rem !default;
 // scss-docs-end border-radius-variables
 
-$outline-width:               $border-width !default;  // Boosted mod
-$outline-offset:              $outline-width !default; // Boosted mod
+$outline-zindex:              5 !default; // Boosted mod
+$outline-width:               3px !default;  // Boosted mod
+$outline-offset:              2px !default; // Boosted mod
```
louismaximepiton, line 593: Should be 3px

louismaximepiton (changes requested):
I'll finish the review once this has been changed !

scss/_variables.scss
```diff
@@ -584,8 +588,12 @@ $border-radius-2xl:           null !default;
...
+$outline-color:               var(--#{$prefix}focus-outline-outer-color) !default; // Boosted mod
+$outline-box-width:           3px !default; // Boosted mod
```
louismaximepiton, line 595: After asking to design, this should be to 2px

scss/_reboot.scss
```diff
@@ -85,21 +85,16 @@ body {
...
 // Boosted mod: focus state
 //
-// 1. Default focus state
-// 2. Using the :focus-visible polyfill to hide outline defensively
+// 1. Using the :focus-visible polyfill to hide outline defensively
 //    See https://github.com/WICG/focus-visible
-// 3. Using the :focus-visible pseudo-class if supported by the browser
+// 2. Using the :focus-visible pseudo-class if supported by the browser
 // scss-docs-start focus-visibility
-:focus {
```
louismaximepiton, line 93: Reintroduce it otherwise there is'nt any focus anymore

scss/_reboot.scss
```diff
@@ -85,21 +85,16 @@ body {
...
-  @include focus-visible(); // 1
-}
-
 .js-focus-visible :focus:not([data-focus-visible-added]),
-.js-focus-visible .focus:not([data-focus-visible-added]) { // 2
+.js-focus-visible .focus:not([data-focus-visible-added]) { // 1
   outline: 0 !important;
```
louismaximepiton, line 99: Add `box-shadow: none;`

scss/_reboot.scss
```diff
@@ -85,21 +85,16 @@ body {
...
 }
 
-:focus:not(:focus-visible) { // 3
+:focus:not(:focus-visible) { // 2
   outline: 0 !important;
```
louismaximepiton, line 103: Add `box-shadow: none;`

louismaximepiton (changes requested):

scss/_variables.scss
```diff
@@ -1070,12 +1075,6 @@ $form-star-size:                  1.5625rem !default;
 $form-star-size-sm:               1.25rem !default;
 $form-star-margin-between:        -.125rem !default;
 
-$form-star-focus-box-shadow:      $input-btn-focus-box-shadow !default;
```
louismaximepiton, line 1073: Same here, deprecate rather than remove

scss/_variables.scss
```diff
@@ -1536,7 +1540,7 @@ $accordion-body-padding-x:                $spacer !default; // Boosted mod
 $accordion-button-padding-y:              $accordion-padding-y !default;
 $accordion-button-padding-x:              $accordion-padding-x !default;
 $accordion-button-color:                  $accordion-color !default;
-$accordion-button-bg:                     var(--#{$prefix}accordion-bg) !default;
+$accordion-button-bg:                     transparent !default; // Boosted mod
```
louismaximepiton, line 1543: Is this change still needed since we changed the focus on accordion ?

louismaximepiton (changes requested):
We have a box-shadow around some no outline components (inputs, ...)

julien-deramond:
Apparently this PR can also close #398

scss/_accordion.scss
```diff
@@ -50,9 +50,9 @@
   letter-spacing: var(--#{$prefix}accordion-btn-letter-spacing); // Boosted mod
   background-color: var(--#{$prefix}accordion-btn-bg);
   border: 0;
+  outline-offset: add($outline-offset, var(--#{$prefix}accordion-border-width)); // Boosted mod
```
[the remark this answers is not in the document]
louismaximepiton: Tackled in #1727.

louismaximepiton (changes requested):
I found an issue that made me reconsider almost everything of `box-shadow: none`: if people forget about the context, they might have some trouble with focus on changing background (e.g. version dropdown issue, list-group contextual classes, star-rating, ...).

Some other comments:
~~- [ ] Need to add `.dropdown-menu-dark` inside the Boosted header version dropdown otherwise we don't have focus anymore.~~
- [ ] Need to tackle this ? Was already there
image
- [x] Do we need to tackle the contextual classes knowing that it will change very soon + shouldn't be used atm ? (example secondary)
image
- [x] Need to remove the box-shadow on offcanvas (with the `outline:0`)
- [x] Pagination overrides the box-shadow so there isn't any
- [x] Should add `box-shadow:none` everywhere we find `outline:0`
- [x] Navbar-toggler doesn't have box-shadow

scss/_accordion.scss
```diff
@@ -50,9 +50,9 @@
...
   @include border-radius(0);
   overflow-anchor: none;
-  @include transition(var(--#{$prefix}accordion-transition));
```
julien-deramond, line 55: Haven't checked all the PR yet but IMO you can let it here without actually being an issue. Otherwise you'd have to deprecate `--#{$prefix}accordion-transition` which is only used here, and add a "Boosted mod" comment.

scss/_accordion.scss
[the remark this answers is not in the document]
julien-deramond: Yep change is required if possible

louismaximepiton (approved):
LGTM!

scss/_variables.scss
```diff
@@ -584,14 +584,24 @@ $border-radius-2xl:           null !default;
 $border-radius-pill:          50rem !default;
 // scss-docs-end border-radius-variables
 
-$outline-width:               $border-width !default;  // Boosted mod
```
julien-deramond, line 587: TODO: `$outline-width` and `$outline-offset` should be deprecated

scss/_navbar.scss
```diff
@@ -186,13 +187,13 @@
     padding-top: subtract(var(--#{$prefix}navbar-nav-link-padding-y), var(--#{$prefix}navbar-border-width));
     margin-left: calc(-1 * var(--#{$prefix}navbar-nav-link-padding-x)); /* stylelint-disable-line function-disallowed-list */
     word-break: break-all;
-    outline-offset: $spacer;
-    @include transition($navbar-transition);
```
julien-deramond, line 190: TODO: double check that it must be indeed removed. More generally do we keep `@include transition`s or not.

scss/_stepped-process.scss
```diff
@@ -98,32 +98,29 @@
...
   }
 
-  &:focus {
-    text-decoration: var(--#{$prefix}stepped-process-link-text-decoration);
-    outline-offset: $spacer * .25;
-  }
-
   .active & {
     color: var(--#{$prefix}stepped-process-link-active-color);
+    outline-color: $black;
```
julien-deramond, line 118: Shouldn't we use a focus visible custom property here?

site/content/docs/5.2/migration.md
```diff
@@ -156,6 +167,7 @@ If you need more details about the changes, please refer to the [v5.2.3 release]
       <li><code>$helper-icon</code></li>
       <li><code>$input-border-radius-sm</code></li>
       <li><code>$input-font-size-sm</code></li>
+      <li><code>$step-link-active-outline</code></li>
```
julien-deramond, line 170: TODO: a lot more vars (CSS and/or Sass) are deprecated or new

julien-deramond (approved):
Sounds good to go into v5.2.3. GG 🚀 
We'll make the design review/a11y review asynchronally after exceptionnally because we really need to release v5.2.3. Part of it was already reviewed but let's do it in more details.

PR #1441, opened 2022-08-05
Re-introduce thumbnail images

louismaximepiton:
Wondering if we should introduce paragraph to explain how to manipulate thumbnails with utilities. IMHO we don't need it but I prefer to ask.


MewenLeHo (approved):
🆗

scss/_variables.scss
```diff
@@ -1758,7 +1758,7 @@ $list-group-action-active-bg:       $list-group-active-bg !default; // Boosted m
 // Image thumbnails
 
 // scss-docs-start thumbnail-variables
-$thumbnail-padding:                 $spacer * .5 !default;
+$thumbnail-padding:                 $spacer !default; // Boosted mod
```
julien-deramond, line 1761: Why do we change the padding here? Is there a ODS guideline about it?

Rendering seems weird to me in Examples with this value:

Capture d’écran 2022-08-29 à 08 52 05

louismaximepiton: Ah yes, you're right, I haven't seen those. For what I remember the design comment on thumbnails was: 2px border and 20px adjustable padding. So I guess that we could stay with the previous value. Would you be fine with this ? 😄

julien-deramond: I let you show that to the designers in order to find a rule we agree on.

louismaximepiton:
Returns from spec review:
- The borders should be `$gray-500`
- There should be no padding

julien-deramond:
@louismaximepiton Based on the last comment, I propose https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1441/commits/dfcc545044e434e766495c6f80a6c0d8c5441910 to remove the padding and change the border color.
I let you check the impacts wherever `.img-thumbnail` is used and tell me if this is OK for you. In this case, I'll merge this PR.

site/content/docs/5.2/content/images.md
```diff
@@ -14,7 +14,13 @@ Images in Boosted are made responsive with `.img-fluid`. This applies `max-width
 {{< placeholder width="100%" height="250" class="bd-placeholder-img-lg img-fluid" text="Responsive image" >}}
 {{< /example >}}
 
-<!-- Boosted mod: no rounded thumbnails -->
+## Image thumbnails
+
+In addition to our [border-radius utilities]({{< docsref "/utilities/borders" >}}), you can use `.img-thumbnail` to give an image a 2px border appearance.
+
+{{< example >}}
+{{< placeholder width="200" height="200" class="img-thumbnail" title="A generic square placeholder image with a white border around it, making it resemble a photograph taken with an old instant camera" >}}
```
louismaximepiton, line 22:
```suggestion
{{< placeholder width="200" height="200" class="img-thumbnail" title="A generic square placeholder image with a white border around it, making it resemble a photograph taken with an old instant camera" background="#eee" >}}
```

julien-deramond (approved):

PR #1445, opened 2022-08-08
Error messages: Fix margin-top

louismaximepiton:
Fix that works for `.form-text` too (IMO, it's better to change both at the same time).


MewenLeHo (approved):
🆗

julien-deramond:
Is it possible to enrich the description in order to get what's fixed in this PR?

PR #1446, opened 2022-08-08
Links: fix hover state on dark background

louismaximepiton:
Fix hover state for links.


MewenLeHo (approved):
🆗

julien-deramond (approved):
Made a change to simplify the rule

PR #1447, opened 2022-08-08
Buttons: fix `letter-spacing`

louismaximepiton:
Fix for the `letter-spacing` inside large buttons. Fix for mixin `button-size`.


MewenLeHo (approved):
:ok:

scss/mixins/_buttons.scss
```diff
@@ -41,12 +41,13 @@
 // Boosted mod: no `@mixin button-outline-variant`
 
 // scss-docs-start btn-size-mixin
-@mixin button-size($padding-y, $padding-x, $font-size, $border-radius, $line-height: null, $icon-spacing: $btn-icon-padding-x) {
+@mixin button-size($padding-y, $padding-x, $font-size, $border-radius, $line-height: null, $letter-spacing: $letter-spacing-base, $icon-spacing: $btn-icon-padding-x) {
```
julien-deramond, line 44: Can't we add this parameter at the end?
Right now, it is a breaking change. If it is a the end, since it has a default value, it wouldn't be a breaking change anymore for people using this mixin.

julien-deramond: Done in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1447/commits/ddc7a9ed1a9f983139c6214b69f89802c58717d2. @louismaximepiton can you please double check this modification before I merge this PR?

julien-deramond (approved):

scss/mixins/_buttons.scss
louismaximepiton: Fine for me !

PR #1449, opened 2022-08-08
Cards: Header padding-top fix

louismaximepiton:
Not sure that the actual implementation is the best but if we think for the future and the small and large cards will change this value as well.


MewenLeHo (approved):
🆗

scss/_card.scss
```diff
@@ -4,6 +4,7 @@
 
 .card {
   // scss-docs-start card-css-vars
+  --#{$prefix}card-spacer-top: #{$card-spacer-top}; // Boosted mod
```
julien-deramond, line 7: IMO it's disturbing to have a `--bs-card-spacer-top` that acts on the top padding and then something named `--bs-card-spacer-y` only used for the bottom padding.
As a user, I don't know how to use those variables; will `--bs-card-spacer-top` override the value of the `--bs-card-spacer-y`, which one has the priority, etc.

louismaximepiton: Another solution could be to set the padding top as a ratio of the `card-spacer-y`

site/content/docs/5.2/migration.md
```diff
@@ -132,9 +133,18 @@ If you need more details about the changes, please refer to the [v5.2.3 release]
     </ul>
   </details>
 
+- <details class="mb-2">
+    <summary><span class="badge bg-danger">Breaking</span> Dropped Sass variables:</summary>
```
julien-deramond, line 137: Note for myself: replace "Dropped" by "Deprecated" in this migration guide

julien-deramond (approved):
LGTM!

PR #1450, opened 2022-08-08
Cards: quote fix

louismaximepiton:
Make the quotes look the same everywhere.

### DIFF
--- a/site/content/docs/4.6/components/card.md
+++ b/site/content/docs/4.6/components/card.md
@@ -202,7 +202,7 @@ Card headers can be styled by adding `.card-header` to `<h*>` elements.
   </div>
   <div class="card-body">
     <blockquote class="blockquote mb-0">
-      <p>A well-known quote, contained in a blockquote element.</p>
+      <p class="mb-0">A well-known quote, contained in a blockquote element.</p>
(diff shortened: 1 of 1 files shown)


julien-deramond (changes requested):
You can't change the markdown for v4 here since in v5 you don't have this `class="mb-0"`. It will impact the migration of the users. If you really want to do this modification, you'd have to change the `.blockquote-footer` rule (by handling the negative top margin as done in v5).
**Note:** In Boosted v5, we could change this value. Both elements seem to be to close to each other.

louismaximepiton:
> You can't change the markdown for v4 here since in v5 you don't have this `class="mb-0"`. It will impact the migration of the users. If you really want to do this modification, you'd have to change the `.blockquote-footer` rule (by handling the negative top margin as done in v5). **Note:** In Boosted v5, we could change this value. Both elements seem to be to close to each other.

This is done this way in [Content > Typography](https://boosted.orange.com/docs/4.6/content/typography/#naming-a-source) (Same in Bootstrap v4). However, the rendering is different between Typography and Card, so I'm wondering if this should be changed on Bs side or not at all ?

julien-deramond (approved):

julien-deramond:
Oups, bad manipulation. I'm closing it since v4 won't be maintained anymore. Let's focus on v5.

PR #1451, opened 2022-08-09
Dropdown: `min-width` on toggle-split dropdown

louismaximepiton:
After a discussion with design team in spec meeting (long time ago), we arrived at the conclusion that every `.dropdown-toggle-split` should have the same width, independently of the arrow direction. I think that there isn't any specs since they don't want directions on arrows in dropdown.

Maybe a better solution could be to set a width and remove padding-x.


MewenLeHo (approved):
🆗

julien-deramond:
@louismaximepiton Could you complete the description to explain what this PR fixes or what this modification is needed for? Not sure to understand the context.

julien-deramond (approved):

PR #1459, opened 2022-08-12
Text break: remove for RTL and add a callout

louismaximepiton:
Backport of #578. No need to backport on Bootstrap since they didn't support rtl.


isabellechanclou (approved):
LGTM

julien-deramond (approved):

PR #1460, opened 2022-08-12
Brand doc: Fix clear space that overlap on certain breakpoints

louismaximepiton:


louismaximepiton (commented):
Thanks a lot for these changes, it looks way better now !

isabellechanclou (approved):

julien-deramond (changes requested):
There's an issue with the horizontal alignment:
Screenshot from 2022-08-22 12-42-20

Should stay like this:

Screenshot from 2022-08-22 12-42-41

julien-deramond (approved):

PR #1465, opened 2022-08-17
Delete .mark padding in Docs

isabellechanclou:
The `.mark` padding of 0.2em used in Docs>About>Brand>#master-artwork and in Docs>Content>Typography>#inline-text-elements was vertically too much, occurring an overlapping on the text right above in a paragraph. To avoid that overlapping, the padding is removed on vertical axis in this PR.

**With the padding:**
image

**Without the paddding:**
image

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -416,7 +416,7 @@ $blockquote-letter-spacing:   -.015625rem !default; // Boosted mod
 $hr-border-color:             $gray-300 !default;
 $hr-border-width:             $border-width * .5 !default;
 
-$mark-padding:                0 .2em !default;
+$mark-padding:                0 .2em !default; // Boosted mod
 
 $dt-font-weight:              $font-weight-bold !default;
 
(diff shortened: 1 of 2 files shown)


louismaximepiton (approved):
LGTM 🚀

scss/_variables.scss
```diff
@@ -416,7 +416,7 @@ $blockquote-letter-spacing:   -.015625rem !default; // Boosted mod
 $hr-border-color:             $gray-300 !default;
 $hr-border-width:             $border-width * .5 !default;
 
-$mark-padding:                .2em !default;
+$mark-padding:                0 .2em !default;
```
julien-deramond, line 419:
```suggestion
$mark-padding:                0 .2em !default; // Boosted mod
```

julien-deramond (approved):
Please inform the designers that this element should be described in the DSM because [`<mark>` is a basic HTML element](https://developer.mozilla.org/fr/docs/Web/HTML/Element/mark).

This must be backported to v5.

isabellechanclou:
OK for informing the designers.

OK for backporting to v5, knowing that in V5.2, `<mark>` is no more used in is used only in `Docs>About>Brand>#master-artwork`, but only in `Docs/content/typography/#inline-text-elements` and in `Example/cheatsheet` not visible at the moment. In these page, depending on the usage context of <mark>, the change might not show.

julien-deramond:
`<mark>` can be used by our users in their projects.

isabellechanclou:
ℹ️ Backport to v5 in PR #1484

PR #1466, opened 2022-08-17
List-group: Branded

### LINKED ISSUE #1452: List group: make the component branded with the specs.
### Proposal

- [ ] Make the hoverable and clickable list group branded, following the design specifications in Zeplin.
- [ ] Make the contextual variant look great.

louismaximepiton:
Fixes #1452.

On hold until design review of specs.


isabellechanclou:
**Notes:**
- The **focus state** doesn't have the correct appearance in this PR. It will be corrected in the focus dedicated PR #1437

- In the **#Flush paragraph**, it might be useful to add a design callout saying that this flush variant shall be used inside a container with a top and a bottom border, to have at the end the same design as illustrated on the design specifications.

    --> **Design specifications illustration:**
    image

    --> **Boosted illustration:**

    image

- In the **#contextual-classes paragraph second example**, the design specifications have to be checked with designers for the color of icons on the different interactive states (hover, pressed/active, ...) for what they've described for the light version is not consistent with what they've described for the dark version. In the light version, the icons keep the same color no matter the state when they change color in the dark version.

site/content/docs/5.2/components/list-group.md
```diff
@@ -156,7 +158,9 @@ Add `.list-group-horizontal` to change the layout of list group items from verti
 
 ## Contextual classes
 
-Use contextual classes to style list items with a stateful background and color.
+<!-- Boosted mod: Add situational icons instead of changing the bg-color -->
+Use contextual classes to style list items with a stateful icon.
+<!-- End mod -->
 
 {{< example >}}
```
isabellechanclou, line 165: It could be useful to show in this example the selected state on one of the item with icon to show how it looks.

isabellechanclou:
In `List-group.md` file,  paragraph `#custom-content`, the police size of `<small>And some small print.</small>` shall be 16px and not 14px.

louismaximepiton:
>     * paragraph `#custom-content`, the police size of `<small>And some small print.</small>` shall be 16px and not 14px.

I don't think so, we don't have specs for this specific case + `<small>` should display a font-size of 14px. Maybe only a callout should do the trick ? Linked to #1199.

>     * paragraph `#checkboxes-and-radios`, I think the checkbox/radio button left margin should be 10px and not more, for in the design specification "Numbered", the number has a left margin of 10px. To me, it the same type of case.

Spoke together and we'll propose a shifted design for a design review: `padding-left: 15px` for every variant even Numbered.

louismaximepiton:
Adding what was said in design spec meeting:
- The specs for active states is great for the dark variant and should be backported to the white variant. -> Nothing to change atm
- Flush variant should stay as-is but we might introduce some examples to show how to build the different variants. -> Open another issue to tackle it ?

louismaximepiton:
Please see #1467. Closing this one.

PR #1484, opened 2022-08-25
Fix .mark padding (backport from v4.6)

isabellechanclou:
Backport form v4.6: Fix of `<mark>` vertical padding to avoid overlapping on the lines above and below.  See #1465.

ℹ️  `<mark>` is no more used in Docs>About>Brand>#master-artwork, but only in Docs/content/typography/#inline-text-elements and in Example/cheatsheet not visible at the moment. In these pages, depending on the usage of the context, the change might not show.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -772,7 +772,7 @@ $dt-font-weight:              $font-weight-bold !default;
 
 $list-inline-padding:         $spacer * .25 !default;
 
-$mark-padding:                0 .2em !default; // Boosted mod
+$mark-padding:                0 .1875em !default; // Boosted mod
 $mark-bg:                     $primary !default;
 // scss-docs-end type-variables
 // End mod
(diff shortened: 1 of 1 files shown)


scss/_variables.scss
```diff
@@ -772,7 +772,7 @@ $dt-font-weight:              $font-weight-bold !default;
 
 $list-inline-padding:         $spacer * .25 !default;
 
-$mark-padding:                .1875em !default;
+$mark-padding:                0 .2em !default; // Boosted mod
```
louismaximepiton, line 775:
```suggestion
$mark-padding:                0 .1875em !default; // Boosted mod
```

louismaximepiton (commented):
Shouldn't we keep the Bootstrap value to be closer to Bs version ?

scss/_variables.scss
isabellechanclou: Sure, we can. It can indeed be more relevant that way than staying closer to the value we put in Boosted v4.6...
@julien-deramond @louismaximepiton: Tell me what you like better.

PR #1488, opened 2022-08-29
Docs: Change api responsive values

louismaximepiton:
Set the responsive values to ours instead of the Bootstrap ones.


julien-deramond (approved):
Good catch!

PR #1489, opened 2022-08-29
Close button: adding missing variants without new class

louismaximepiton:
Adding all the DSM variants inside an `Eco-design variant` section.

### Live previews

- <https://deploy-preview-1489--boosted.netlify.app/docs/5.3/components/close-button/#eco-design-variant>


site/content/docs/5.2/components/close-button.md
```diff
@@ -33,7 +33,7 @@ Change the default `.btn-close` to be white with the `.btn-close-white` class.
 
 ## Without specific class
 
-Close buttons can also be created without `.btn-close`.
+Close buttons can also be created without `.btn-close` in Boosted.
```
julien-deramond, line 36:
```suggestion
Close buttons can also be created without `.btn-close`.
```

site/content/docs/5.2/components/close-button.md
```diff
@@ -45,6 +45,38 @@ Close buttons can also be created without `.btn-close`.
   <svg width="1.25rem" height="1.25rem" fill="currentColor" aria-hidden="true" focusable="false"><use xlink:href="/docs/{{< param docs_version >}}/assets/img/boosted-sprite.svg#delete"></use></svg>
   <span class="visually-hidden">Close</span>
 </button>
+
+<button type="button" class="btn btn-icon btn-secondary">
```
julien-deramond, line 49: Where do we use those buttons in the DSM?

site/content/docs/5.2/components/close-button.md
```diff
@@ -45,6 +45,38 @@ Close buttons can also be created without `.btn-close`.
...
+</button>
+{{< /example >}}
+
+{{< example class="bg-dark" >}}
+<button type="button" class="btn btn-icon btn-no-outline btn-inverse">
+  <svg width="1.25rem" height="1.25rem" fill="currentColor" aria-hidden="true" focusable="false"><use xlink:href="/docs/{{< param docs_version >}}/assets/img/boosted-sprite.svg#delete"></use></svg>
+  <span class="visually-hidden">Close</span>
+</button>
+
+<button type="button" class="btn btn-icon btn-no-outline btn-inverse" disabled>
```
julien-deramond, line 66: There is an issue here, it should be grayish instead of white (like this one `<button type="button" class="btn btn-icon btn-inverse btn-secondary" disabled>`). Otherwise we can't differentiate it from a regular button.

julien-deramond (changes requested):

site/content/docs/5.2/components/close-button.md
louismaximepiton: I think the main issue is that `.btn-no-outline` doesn't have a disabled state for now but it might be fixed via #1426.

site/content/docs/5.2/components/close-button.md
louismaximepiton: In the [variant section of the DSM](https://system.design.orange.com/0c1af118d/p/278ebc-buttons-standard/b/247486/t/12ce14). Do you think we should add a variant like `.btn-close-bordered` for the `.btn-close`?

julien-deramond: You should ask this to the designers. I'm not sure to have seen components using this variant as a close button. IMO the link you're referring to is only an example of an outline button with an icon in it; it doesn't necessarily mean we should insert it as a "Close button".

louismaximepiton: Ah you might be right, it's a bit confusing, I'll ask as soon as possible to be clarified on this point!

louismaximepiton: Will be merged with #1705.

site/content/docs/5.2/components/close-button.md
louismaximepiton: Done

MewenLeHo (commented):
Actually note sure why the whole section called 'Without specific class' exists?
It is basically 'use the btn-close component without using .btn-close'. But this class was made for this component so why not using it instead of using an alternative way to do the same thing?

It seems to come from this PR: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/960/files

I also found this commit: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/960/commits/c64c5a25e4287e4524adce5f8c2402c961babba5

```
{{< callout warning >}}
This section might change in the future with [#521]({{<param repo>}}/issues/521). The class **`.btn-close` might be removed**.
{{< /callout >}}
````

It was in 2021 and .btn-close` is still here?

File: site/content/docs/5.3/components/close-button.md [hunk 4]
```diff
@@ -47,6 +47,38 @@ Close buttons can also be created without `.btn-close` to reduce the size of you
   <svg width="1.25rem" height="1.25rem" fill="currentColor" aria-hidden="true" focusable="false"><use xlink:href="/docs/{{< param docs_version >}}/assets/img/boosted-sprite.svg#delete"></use></svg>
   <span class="visually-hidden">Close</span>
 </button>
+
```
MewenLeHo, line 50: As discussed with our tech lead, we can keep this part but we will need to add a callout at the beginning of the page to explain why we need the examples in this section in Boosted.

MewenLeHo (commented):
Callout needed.

louismaximepiton:
FWIW, the close button takes `.2KB` for now without the outlined variant.

MewenLeHo (approved):
LGTM 🚀

Franco-Riccitelli:
All looks good to me.

julien-deramond (approved):
LGTM

PR #1491, opened 2022-08-29
Docs: reorder Carousel doc parts

louismaximepiton:
Main goal: be closer to Bootstrap.

### DIFF
--- a/site/content/docs/5.2/components/carousel.md
+++ b/site/content/docs/5.2/components/carousel.md
@@ -183,34 +183,6 @@ Add captions to your slides easily with the `.carousel-caption` element within a
 </div>
 {{< /example >}}
 
-### Disable touch swiping
-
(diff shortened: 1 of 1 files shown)


site/content/docs/5.2/components/carousel.md
```diff
@@ -155,20 +155,23 @@ Add captions to your slides easily with the `.carousel-caption` element within a
       <div class="carousel-caption d-none d-md-block">
         <h5 class="mb-0">First slide label</h5>
         <p class="mb-0">Some representative placeholder content for the first slide.</p>
+        <a href="#" title="Example link #1">Example link #1</a>
```
julien-deramond, line 158: Since there's no dedicated CSS to render those links and since apparently there's no guidelines in the DSM, I'm not sure we need to add this in our docs.

louismaximepiton: Yeah, I'll remove them from the PR since it was only a proposal of backport from v4.

julien-deramond (approved):

PR #1496, opened 2022-08-29
Docs: Typography hide `display-5` and  `display-6`

louismaximepiton:
Hide `.display-5` and `.display-6` from the documentation since they don't bring anything. Might be revert when we'll have the Specifications for font-size.


julien-deramond:
IMO this class exist in Boosted so should be documented even if the rendering doesn't bring anything.

louismaximepiton:
I totally agree with you, it was just a proposal. I think we can close this PR.

PR #1513, opened 2022-09-05
chore/merge-main@2504b89

julien-deramond:
- [x] https://github.com/twbs/bootstrap/commit/a329575d82a65660ffa7b18cbd9f291e0ee7eef5
- [x] https://github.com/twbs/bootstrap/commit/585146a6a7aa70faf25442d7d28636ce57e29588
- [x] https://github.com/twbs/bootstrap/commit/75e09b1c0f5ae5f51078c7a25fe36d892c5cfcfe
- [x] https://github.com/twbs/bootstrap/commit/b8880e5eec6bb4f33578578ba2413f0d91424382
  - @louismaximepiton Please carefully check this change 🙏  
- [x] https://github.com/twbs/bootstrap/commit/32c457db4b6ff389efbd35772b24746c7ffb0b6d
  - [x] Need to double-check all use cases to avoid regressions
- [x] https://github.com/twbs/bootstrap/commit/2504b8995095d2bb41c9686afa175f9eaa91bec2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "clean-css-cli": "^5.6.1",
         "cross-env": "^7.0.3",
         "eslint": "^8.23.0",
-        "eslint-config-xo": "^0.41.0",
+        "eslint-config-xo": "^0.42.0",
         "eslint-plugin-import": "^2.26.0",
         "eslint-plugin-markdown": "^3.0.0",
         "eslint-plugin-unicorn": "^42.0.0",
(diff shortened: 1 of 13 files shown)


louismaximepiton (approved):
I think it would be easier to remind if we had every multiplication in the same way. But looks fine to me atm.

Regarding the buttons changes, I couldn't spot any breaking change, so I think that it could be merged. However, 2 questions come to my mind:
- Do we need to change the way we manage `.btn-check` in Boosted (do a button-variant)
- Should we introduce `:focus-visible` inside Boosted instead of the library `focus-visible` ? If yes, might be reported in #1189.

julien-deramond:
> I think it would be easier to remind if we had every multiplication in the same way. But looks fine to me atm.

Interesting. Could be a solution to trick our brains.

> Regarding the buttons changes, I couldn't spot any breaking change, so I think that it could be merged. However, 2 questions come to my mind:
>     * Do we need to change the way we manage `.btn-check` in Boosted (do a button-variant)

I haven't looked at it in details. If you think this is a good way of handling `.btn-check` you can try a PR.

>     * Should we introduce `:focus-visible` inside Boosted instead of the library `focus-visible` ? If yes, might be reported in #1189.

We definitely need to think about it and check if we still need our `focus-visible` dependency.

julien-deramond:
I will rollback the modifications regarding the buttons. There's regressions in [No outline icon buttons](https://deploy-preview-1513--boosted.netlify.app/docs/5.2/components/buttons/#no-outline) (click) and [Social buttons](https://deploy-preview-1513--boosted.netlify.app/docs/5.2/components/buttons/#social-buttons) on hover.
It is safer to keep what we already developed and checked.

louismaximepiton:
Ah yep you're right, only checked the check&radios ><

louismaximepiton:
Seems like the [button tag](https://deploy-preview-1513--boosted.netlify.app/docs/5.2/components/buttons/#button-tags)'s link broken too! But it's due to the a:focus rule, should we correct that here ?

PR #1428, opened 2022-07-29
Back to top: dark variant + small doc fix

louismaximepiton:
Commits can be split if needed.

- Adding dark-variant to the component
- Correct table in small viewport that caused overflow-x on the page

PR #1520, opened 2022-09-07
Back to top: remove not branded variant

louismaximepiton:
Commits can be split if needed.

- Adding a missing CSS var on Back to top
- Removing the not branded variant.

Following the https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1428#issuecomment-1239036156.


julien-deramond (approved):
LGTM!

PR #1525, opened 2022-09-08
List-group: Add dark variant

louismaximepiton:
First round of dark variant for list-group.


julien-deramond:
The disabled state is a bit difficult to read because not in bold. I suppose this will be tackled with the "re-branding" of the list groups in another PR?

Screenshot from 2022-09-09 09-56-59

louismaximepiton:
Yes exactly, every link will be in bold with the incoming re branding. I think it doesn't causes any a11y issue since these are disabled items.

julien-deramond (approved):
LGTM! 🚀

PR #1528, opened 2022-09-09
Navbar: Allow `flex-grow` on each `.navbar-collapse`

### LINKED ISSUE #1514: Right alignment in Navar text example
### Prerequisites

### Describe the issue

In [Navbar text example](https://boosted.orange.com/docs/5.2/components/navbar/#text), the text "Navbar text with an inline element" should be right-aligned in desktop mode and in small screen mode.

Screenshot from 2022-09-06 08-17-51

Bootstrap reference: https://getbootstrap.com/docs/5.2/components/navbar/#text-1

In desktop mode, this is due to the absence of `flex-grow: 1` mentioned in the source code as having side effects maybe with the Orange Navbar component.

For this issue, fix the navbar CSS while trying to:
- not changing the markup from Bootstrap
- not to add regressions in Orange Navbar

Check also the other examples in this page compared to Bootstrap.

### Reduced test cases

N/A

Linux

Firefox

v5.2.0

louismaximepiton:
Fixes #1514.

Some few changes that'd allow to have `flex-grow`. This might have some breaking changes in the markup but I can't see any visual regression.

Might simplify a bit the markup and I think it would be way better this way.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -265,26 +265,28 @@
 // When collapsed, prevent the toggleable navbar contents from appearing in
 // the default flexbox row orientation. Requires the use of `flex-wrap: wrap`
 // on the `.navbar` parent.
-.navbar-collapse.show, // Boosted mod: always keep nav-items at full width when transitioning
-.navbar-collapse.collapsing {
-  flex-basis: 100%;
-  // Boosted mod: no flex-grow to avoid weird behavior in expanded breakpoints and .show
+.navbar-collapse {
+  // Bosted mod: flex-basis only on .show and .collapsing
+  flex-grow: 1;
   // For always expanded or extra full navbars, ensure content aligns itself
   // properly vertically. Can be easily overridden with flex utilities.
   align-items: center;
+  justify-content: flex-end; // Boosted mod
 
-  // Boosted mod: Handle icons support and padding of expanded navbars
-  &[class*="d-"] {
-    border-top: var(--#{$prefix}navbar-border-width) solid var(--#{$prefix}navbar-border-color);
+  // Boosted mod: Handle multiple collapse
+  // Always keep nav-items at full width when transitioning
+  &.show,
+  &.collapsing {
+    flex-basis: 100%;
+    justify-content: flex-start;
 
-    @include media-breakpoint-up(md) {
-      margin-bottom: calc(-1 * var(--#{$prefix}navbar-padding-y)); /* stylelint-disable-line function-disallowed-list */
-    }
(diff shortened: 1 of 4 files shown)


scss/_navbar.scss
```diff
@@ -457,6 +447,25 @@
         .navbar-collapse {
           display: flex !important; // stylelint-disable-line declaration-no-important
```
MewenLeHo, line 448: Can work without the important statement

site/content/docs/5.2/components/navbar.md
```diff
@@ -323,7 +323,7 @@ Mix and match with other components and utilities as needed.
           <a class="nav-link" href="#">Pricing</a>
         </li>
       </ul>
-      <span class="navbar-text ms-3">
+      <span class="navbar-text">
         Navbar text with an inline element
```
MewenLeHo, line 327: We may have a problem if the inline element is  a longer text, i.e. more than one line for instance. It should not happen but we never know. 
It may be possible to prevent this by using flex-basis or flex-shrink.

MewenLeHo (commented):
Good but some problems on smaller resolutions

scss/_navbar.scss
louismaximepiton: It comes from Bootstrap so let's keep it as-is if you're fine with it!

site/content/docs/5.2/components/navbar.md
louismaximepiton: For me the actual behavior with more than one line is the one wanted when it was developed I think. The fact is that the whole content should be contained inside the navbar whatever happens. The solution we tried when we worked on implementation was a `word-break: keep-all` inside nav-links but on too much items it would go outside the navbar. 

For the proposed solutions, I'd like to discuss it a bit more since I can't see how it would work but I'm not expert with flexboxes!

Not sure to understand well the Edit issue 😞

MewenLeHo: The actual behaviour is the desired behaviour so ok form me 🚀

MewenLeHo:
> Good but some problems on smaller resolutions

The actual behaviour is the desired behaviour so ok form me 🚀

MewenLeHo (approved):
ok

julien-deramond:
@louismaximepiton Could you please add the necessary information in the migration guide since there are changes in the markup.

vprothais:
I close this PR after adding it to "Boosted shelf" Milestone to keep track of existing PR on Boosted if needed.

Issue #1553, opened 2022-10-14
CSS: Remove `$boosted-prefix` 


louismaximepiton:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Proposal

Remove the custom `$boosted-prefix` to only have one.

⚠️ `-webkit-mask-` introduced in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1574/files could probably be removed. (more info in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1574#discussion_r1006434287)

### Motivation and context

Be more handful for users.

PR #1574, opened 2022-10-26
Fix close button inside tags on Chrome

MewenLeHo:
### Description

Missing one vendor prefix for Chrome in order to display close button correctly.
screenshot-deploy-preview-1574--boosted netlify app-2022 10 26-17_06_15

### Motivation & Context

We saw with @louismaximepiton that the close button inside some tags was missing in Chrome because there was a vendor prefix missing for this browser.

⚠️ : need to fix it before release by @julien-deramond 

### Types of change

### Live previews

* https://deploy-preview-1574--boosted.netlify.app/docs/5.2/components/tags/

### Checklist

### DIFF
--- a/scss/_tags.scss
+++ b/scss/_tags.scss
@@ -47,7 +47,7 @@
     width: var(--#{$prefix}tag-close-size);
     min-width: var(--#{$prefix}tag-close-size); // Here to avoid weird behavior on wrap
     height: var(--#{$prefix}tag-close-size);
-    padding: 0;
+    padding: 0; // Force padding on button to have a circle on iOS - Safari and Chrome
     margin: add(calc(-.5 * var(--#{$prefix}tag-close-size)), var(--#{$prefix}tag-font-shift)) var(--#{$prefix}tag-close-margin-end) calc(-.5 * var(--#{$prefix}tag-close-size)) var(--#{$prefix}tag-close-margin-start);
     color: inherit;
     background-color: transparent;
--- a/scss/_tags.scss
+++ b/scss/_tags.scss
@@ -47,6 +47,7 @@
     width: var(--#{$prefix}tag-close-size);
     min-width: var(--#{$prefix}tag-close-size); // Here to avoid weird behavior on wrap
     height: var(--#{$prefix}tag-close-size);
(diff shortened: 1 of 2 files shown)


louismaximepiton (approved):
LGTM 🚀 (Nice catch 😄)

julien-deramond:
> Small issue with iOS: image
> 
> I can see this on both Safari and on Chrome. Tags seems to introduce an overflow on x on small screens. We might add `.flex-wrap` to tag' containers.

Confirmed on iPhone 14/13 with Safari/Chrome

scss/_tags.scss
```diff
@@ -69,6 +69,7 @@
       height: 100%;
       content: "";
       background-color: currentcolor;
+      -webkit-mask-image: escape-svg(var(--#{$boosted-prefix}close-icon)); //stylelint-disable-line property-no-vendor-prefix
```
julien-deramond, line 72: I'm really wondering why we need to add this. Autoprefixer should do it automatically since it does it already for other definitions of `mask-image` in our project. Wondering if it has something to do with CSS vars.
Let's keep this modification but I'm gonna try to find why it does that.

julien-deramond: If the content is anything else than `mask-image: var(--o-close-icon);` the transformation done by Autoprefixer is OK... 🤔 
As soon as the CSS var is called `--o-*` it doesn't work 🤔

julien-deramond: Oh I think I know where it comes from 💡 
In Autoprefixer we have this kind of code `mistakes: ['-khtml-', '-ms-', '-o-']`. Our Boosted prefix si `o-`. There must be a conflict. If I change our prefix to `$boosted-variable-prefix:     ms-` it doesn't work neither.
So I think we can keep this modification but you should add a comment in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1553 that we could remove this line when the `$boosted-prefix` will be removed.

julien-deramond: See https://github.com/postcss/autoprefixer/issues/1484

julien-deramond (approved):

Issue #1626, opened 2022-11-15
Storybook enhancements


julien-deramond:
### Context

https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/565 introduced the usage of Storybook in Boosted as a link between Boosted documentation and the [Design System Manager (DSM)](https://system.design.orange.com/).
In some DSM "Develop" tabs, an iframe shows some components in action that are in fact Storybook stories from our "hidden" URL https://boosted.orange.com/storybook/.

However, the integration of Storybook within Boosted can be improved. This story is an epic one listing independent micro-tasks that should be tackled in different PRs. No need to create sub-issues.

### Enhancements

- [x] Introducing new dependencies ended up having some vulnerable dependencies in our `package-lock.json`
* `trim-newlines` that is patched in 3.0.1
* `glob-parent` that is patched in 5.1.2
* `trim` (because of `@storybook/addon-essential@6.5.13` and `@storybook/html@6.5.13`) that is patched in 0.0.3

Tackled by #2103 ?

---

- [x] Error message when running `npm run storybook-build` and then `npm run storybook`

```
...
> boosted@5.2.1 docs-vnu
> node build/vnu-jar.js

"file:/Users/ju/Orange-Boosted-Bootstrap/_site/storybook/index.html":1.38-1.60: info warning: Self-closing tag syntax in text/html documents is widely discouraged; it's unnecessary and interacts badly with other HTML features (e.g., unquoted attribute values). If you're using a tool that injects self-closing tag syntax into all void elements, without any option to prevent it from doing so, then consider switching to a different tool.
"file:/Users/ju/Orange-Boosted-Bootstrap/_site/storybook/index.html":1.87-^C
```

`vnu-jar.js` doesn't like things generated in Storybook build. `rm -rf site/storybook _site/storybook/` can help to reset the environment but a fix should be found to avoid having this error.

---

- [ ] Fix deprecation warning displayed when running `npm run storybook`

```
(node:24011) DeprecationWarning: Default PostCSS plugins are deprecated. When switching to '@storybook/addon-postcss',
you will need to add your own plugins, such as 'postcss-flexbugs-fixes' and 'autoprefixer'.

See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-default-postcss-plugins for details.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

--- 

[...]

PR #1660, opened 2022-11-23
fix(storybook): generate SB website in `_site` intead of `site`

julien-deramond:
### Related issues

Linked to the following issue mentioned in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1626:

Error message when running `npm run storybook-build` and then `npm run storybook`

```
...
> boosted@5.2.1 docs-vnu
> node build/vnu-jar.js

"file:/Users/ju/Orange-Boosted-Bootstrap/_site/storybook/index.html":1.38-1.60: info warning: Self-closing tag syntax in text/html documents is widely discouraged; it's unnecessary and interacts badly with other HTML features (e.g., unquoted attribute values). If you're using a tool that injects self-closing tag syntax into all void elements, without any option to prevent it from doing so, then consider switching to a different tool.
"file:/Users/ju/Orange-Boosted-Bootstrap/_site/storybook/index.html":1.87-^C
```

### Description

This error is generated by the HTML validator that scans our `site` directory. Since Storybook website is generated in `site`, the HTML validator is not very happy with the self-closing tags.

But in fact, there's no reason to generate Storybook website in `site`. After the release, we're going to copy `_site` and not `site` into `gh-pages`.

When running locally `npm run storybook`, Storybook website is not generated as HTML files.

So the only thing to change here is to keep it working with Netlify which needs to have `site/storybook` directory. I have created a special npm command to differenciate the normal build and the one reserved to Netlify.

### Non-regression testing

* `npm run storybook` works locally and doesn't generate `_site/storybook` nor `site/storybook`
* running `npm run storybook-build` and then `npm run storybook` works without the error mentioned in the description
* `npm run storybook-build` creates `_site/storybook` but nothing in `site`
* `npm run release` creates `_site/storybook` but nothing in `site`

### Types of change

### Live previews

* https://deploy-preview-1660--boosted.netlify.app/storybook

### Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,5 +6,4 @@
 /js/tests/integration/
 /site/static/sw.js
 /site/layouts/
-/site/storybook/
 /stories/auto/
--- a/.gitignore
+++ b/.gitignore
@@ -51,4 +51,3 @@ Thumbs.db
 
 # Storybook
 /stories/auto
-/site/storybook
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "release-zip-examples": "node build/zip-examples.js",
     "dist": "npm-run-all --aggregate-output --parallel css js fonts",
     "test": "npm-run-all lint dist js-test docs-build docs-lint",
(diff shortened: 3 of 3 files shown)


louismaximepiton (approved):
Sounds great ! Couldn't spot any breaking change and will clear some stuff.

PR #1676, opened 2022-12-02
A11y: remove some ambiguous links

louismaximepiton:
### Description

Remove `Examples` titles in order to be consistent in the whole doc + remove ambiguous links between the header and the ToC (search and replace `# Examples` to `# Example`).

### Motivation & Context

Remove rectifiable issues raised by aXe devtools in the browser over the Boosted doc.

### Types of change

### Live previews

* https://deploy-preview-1676--boosted.netlify.app/

### DIFF
--- a/site/content/docs/5.2/components/alerts.md
+++ b/site/content/docs/5.2/components/alerts.md
@@ -8,7 +8,7 @@ aliases:
 toc: true
 ---
 
-## Example
+## Examples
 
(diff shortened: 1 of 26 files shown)


julien-deramond:
Just raising a question here. "Example" if the section contains only one example and "Examples" for multiple examples. So we could have both depending on the components. Is it really an issue in terms of accessibility to have that? In their own context, they are perfectly valid, so is there no way to differenciate them only for a11y screen readers?

louismaximepiton:
Following https://github.com/gohugoio/hugo/issues/10507, it seems like this isn't needed anymore since most of our links are inside a structure `<ul><li></li></ul>`. Furthermore it seems like the links with the same label aren't spotted anymore by aXeDevtools.

Issue #1594, opened 2022-11-07
Docs > Components > Tooltip > Directions: Fix tooltips margins


isabellechanclou:
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

This issue comes from the work done in Issue https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1201 List in all the doc the incompatibility with ODS in which all Boosted components (variants included) have been inspected in order to identify the ones that do not respect ODS specifications and therefore need an alert message saying "Do not use in ODS".
Doing this work, design bugs were noticed.

**Design bug to be corrected:**
In docs/v5.1/components/tooltip/**#directions**,  fix margins between tooltip and the elements on witch it applies. See if it is possible to do so by adding some css.

ℹ️ **In Boosted:**

![image](https://user-images.githubusercontent.com/89514509/203764753-50918912-6a8a-474e-ae03-bb282f5abf2f.png)

ℹ️ **ODS specification:**

![image](https://user-images.githubusercontent.com/89514509/203763838-96c1918f-934f-497f-a754-31b7f493781d.png)

### Reduced test cases

NA

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What version of Boosted are you using?

v5.2

PR #1689, opened 2022-12-06
Fix tooltips and arrows positions

MewenLeHo:
### Related issues

Closes [#1594 ](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1594)

### Description

Change tooltips and related arrows positions depending on positions.

### Motivation & Context

Changes needed to respect Orange Design System specifications.

### Types of change

### Live previews

* https://deploy-preview-1689--boosted.netlify.app/docs/5.2/components/tooltips/#directions

### Checklist


julien-deramond (changes requested):
The change doesn't seem to be applied to popovers. And IMO it should.
Before doing the code review for this one (since it is rather simple in terms of code review), could you please check with the designers before if the value used (and the rendering) is OK so we have their approval?

louismaximepiton:
Just leaving here another way to fix the issue, haven't had the time to check if this could fit the need: `data-bs-offset="0,10"`

MewenLeHo:
I received the answer from the designers, this change must also affect popovers.

Franco-Riccitelli:
Hi, here is my feedback on the Tooltip and Popover pointer spacing.

**Tooltip**
All look good except for the example **Tooltip with HTML**. This is slightly too close to the button. Try moving 2px.

**Popover**
The size of the pointer on Popovers is slightly larger than for Tooltips as Popovers are generally bigger.
The "point" of the pointer is still 5px away from the item it is pointing to (same as Tooltips), but the pointer height is 10px (see attached).

Popover-Spacing

MewenLeHo:
**Tooltips:** ok

**Popovers:** in **Bootstrap**, popovers arrows are made of two half-sized arrows to make it looks like a white arrow with a grey border:

screenshot-getbootstrap com-2022 12 23-11_22_13

It's possible to resolve this PR easily by removing one of this 'half-arrow' and make the second one full size but we currently have a design issue about going tward a similar design, I even made a prototype about it.

I think we should decide what to do with this prototype and update our code accordingly. Then we will decide how to handle this PR.

MewenLeHo:
Closed because of a change in specifications.
I wil start a new one from scratch with the correct informations.

PR #1674, opened 2022-12-01
fix(tables): preserve table row heights with icons and thumbnails

hannahiss:
### Related issues

Fixes #1357 

### Description

Table lines with icons or thumbnails of 30px x 30px should have a 40px height for standard tables (having .table-sm class) and 50px for large tables (having only .table class).

### Motivation & Context

After a design review, it has been concluded that:
- there are two sizes of table rows : standard (40px) and large (50px) in a design point of view
- the row heights should be preserved even when containing icons or thumbnaisl (having a 30 x 30 size)

So, the action to do are:
- ODS examples should be reviewed to comply with these elements (Franco's actions)
- Boosted code should be adjusted to comply with these elements: 
  - implement new shortcode to add blue Boosted callout (info)
  - rename Boosted class bd-ods-incompatibility-alert to bd-ods-alert because it is no longer used only for ODS incompatibilities, but also for recommendations
  - add Boosted info callout at the top of tables page to notice developpers that recommended tables should have the class .table-sm
  - add Boosted info callout in paragraph "Small tables" to notice developpers  that this is the recommended tables
  - add negative margins to img and svg to preserve row heights
  - update migration guide to notice new variables Sass
  - add class bd-example when missing

### Types of change

- Bug fix (non-breaking which fixes an issues)
- New feature (non-breaking change which adds functionality)
- Refactoring (non-breaking change)
- Breaking change (fix or feature that would change existing functionality)

### Live previews

* https://deploy-preview-1674--boosted.netlify.app/docs/content/tables/#with-icons-or-thumbnails

### Checklist

- [x] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)
- [x] My code follows the code style of the project _(using `npm run lint`)_
- [x] My change introduces changes to the documentation
- [x] My change introduces changes to the migration guide
- [x] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1710, opened 2022-12-15
Move table-responsive class from <table> tag to <div> tag above

hannahiss:
### Related issues

Linked to issue https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/1357 and PR https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1674

### Description

`.table-responsive` class was misused on some examples: it was positioned on the `<table>` tag, whereas the documentation says to wrap the table inside a HTML tag (`<div>`) with `.table-responsive`. 

### Motivation & Context

I saw this misuse while correcting some tables row height, on rich content tables examples. 

### Types of change

### Live previews

* https://deploy-preview-1710--boosted.netlify.app/docs/content/tables/#with-icons-or-thumbnails

### Checklist

- [ ] I have added tests to cover my changes


julien-deramond:
Was just looking at the overall modifications to check what's the topic of this PR. The rendering seems broken here:
Screenshot 2022-12-16 at 07 55 51
The checkboxes are "glued" to the left hand side of the table.

julien-deramond:
There is a specific rule somewhere for the docs that can maybe cause this issue only in the docs:
```
.bd-content>.table th:first-child,
.bd-content>.table td:first-child,
.bd-content>.table-responsive .table th:first-child,
.bd-content>.table-responsive .table td:first-child {
 padding-left:0;
}
```

You can try your modification outside of the docs context. Maybe that's just the extra-css for the docs that is the issue.

julien-deramond:
If you want to cheat a little bit, you could probably use a `.table-responsive-{breakpoint}` instead of `.table-responsive` so that it doesn't trigger the specific docs CSS 😇

hannahiss:
It seems that this rule: 
```
.bd-content {
  > .table,
  > .table-responsive .table {
    th,
    td {
      &:first-child {
        padding-left: 0;
      }
    }
  }
}
```
changes almost all tables over documentation, except the ones in Content>Tables section, because they are all wrapped inside  `<div class="bd-example">` that breaks the rule. By moving `.table-responsive` to the div above, I reactivated the rule on the rich content tables **because they are missing `<div class="bd-example">`** (which I corrected in PR #1674)


Example from [Back To Top > Sass options](https://boosted.orange.com/docs/5.2/components/back-to-top/#sass-options): 

With the css rule:
image

Without the css rule:
image

Isn't it strange that all our tables across documentation do not respect our design specifications ?

Tables from ODS design specs:
image

Why should we keep this strange rule ?

julien-deramond:
Since Boosted doc must respect our design guidelines as well. From what I understand, we can probably remove this rule entirely and have a slightly different display compared to Bootstrap.

julien-deramond:
There's a rendering regression (size of checkboxes and texts) between the new version:

Screenshot 2022-12-19 at 12 58 50

And the previous one:

Screenshot 2022-12-19 at 12 59 22

louismaximepiton:
> There's a rendering regression (size of checkboxes and texts) between the new version:
> 
> And the previous one:

The issue comes from the fact that the direct `.table` of `.bd-content` have their font-size reduced. Shall we changed it ? Otherwise a quick solution here is to wrap those with a `.bd-example`.

julien-deramond (approved):

PR #1614, opened 2022-11-14
docs: add design callouts in all the documentation when incompatibility with ODS

isabellechanclou:
### Related issues

Closes #1199
Closes #1118
Closes #1117

### Description

Everywhere needed in Boosted Docs, a callout message has been added.

### Motivation & Context

Callout messages added to warn Boosted users that such component or component variant should not be used because incompatible with the orange design system specifications.

### Types of change

- New feature (non-breaking change which adds functionality)

### Live previews

* https://deploy-preview-1614--boosted.netlify.app/

### Checklist

- [x] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)
- [x] My code follows the code style of the project _(using `npm run lint`)_
- [x] My change introduces changes to the documentation
- [ ] My change introduces changes to the migration guide
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1713, opened 2022-12-16
Design callouts for merging in main

isabellechanclou:
This PR is used to embed only a part of https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1614 so that we can release it in a 2-steps approach.


julien-deramond (approved):
LGTM because I'm the one who included the commits from the other PR 😇 I've only kept those I was comfortable with.

julien-deramond:
@MewenLeHo and @louismaximepiton, I've assigned you the review of this PR. The principle is to flag content that you're not sure about. The goal here is to include as a first-step only things we are comfortable with. If some callouts need more discussion, it'll be removed from this PR and tackled in the main one.

site/content/docs/5.2/components/button-group.md
```diff
@@ -40,6 +40,12 @@ These classes can also be added to groups of links, as an alternative to the [`.
 
 ## Mixed styles
 
+{{< ods-incompatibility-alert >}}
+This variant **using and mixing colored button backgrounds** should not be used because it does not respect the Orange Design System specifications. In button groups, you should only use the button variant that using `.btn .btn-secondary`.
```
louismaximepiton, line 44:
```suggestion
This variant **using and mixing colored button backgrounds** should not be used because it does not respect the Orange Design System specifications. In button groups, you should only use the button variant that use `.btn .btn-secondary`.
```

site/content/docs/5.2/components/carousel.md
```diff
@@ -144,6 +144,12 @@ This `button` must immediately follow your carousel and have the custom `data-bs
 
 Add captions to your slides easily with the `.carousel-caption` element within any `.carousel-item`. They can be easily hidden on smaller viewports, as shown below, with optional [display utilities]({{< docsref "/utilities/display" >}}). We hide them initially with `.d-none` and bring them back on medium-sized devices with `.d-md-block`.
 
+{{< ods-incompatibility-alert >}}
+**Captions** should not be used because they do not respect the Orange Design System specifications.
```
louismaximepiton, line 148: Should we add an example showing how to add text to carousel ?

site/content/docs/5.2/components/dropdowns.md
```diff
@@ -774,6 +804,12 @@ Put a form within a dropdown menu, or make it into a dropdown menu, and use [mar
 
 Use `data-bs-offset` or `data-bs-reference` to change the location of the dropdown.
```
louismaximepiton, line 805: Just leaving it here but `data-bs-offset="0,0"` is totally correct

site/content/docs/5.2/components/dropdowns.md
```diff
@@ -806,6 +842,10 @@ Use `data-bs-offset` or `data-bs-reference` to change the location of the dropdo
 
 By default, the dropdown menu is closed when clicking inside or outside the dropdown menu. You can use the `autoClose` option to change this behavior of the dropdown.
 
+{{< ods-incompatibility-alert >}}
+The 3 last auto close behavior variants should not be used because featuring a single selection, they do no respect the Orange Design System specifications. They can be used only if featuring multiple selections.
```
louismaximepiton, line 846:
```suggestion
The 3 last auto close behavior variants should not be used because featuring a single selection, they do no respect the Orange Design System specifications. They should be used only with multiple selections.

File: site/content/docs/5.2/forms/input-group.md [hunk 5]
```diff
@@ -245,6 +275,12 @@ Input groups include support for custom selects and custom file inputs. Browser
 
 ### Custom select
 
+{{< ods-incompatibility-alert >}}
+These **3 first** form variants should not be used because they do not respect the Orange Design System specifications.
+
+Please refer to the [Forms](https://system.design.orange.com/0c1af118d/p/88ab5b-forms/b/599459) guidelines and to the [Pages](https://system.design.orange.com/0c1af118d/p/20500e-form/b/16bb53) form examples on the Orange Design System website.
```
louismaximepiton, line 281: Can't we redirect on Boosted select ?

site/content/docs/5.2/forms/input-group.md
```diff
@@ -289,6 +325,12 @@ Input groups include support for custom selects and custom file inputs. Browser
 
 ### Custom file input
 
+{{< ods-incompatibility-alert >}}
+These form variants should not be used because they do not respect the Orange Design System specifications.
+
+Please refer to the [Forms](https://system.design.orange.com/0c1af118d/p/88ab5b-forms/b/599459) guidelines and to the [Pages](https://system.design.orange.com/0c1af118d/p/20500e-form/b/16bb53) form examples on the Orange Design System website.
```
louismaximepiton, line 331: Same for Boosted file inputs ?

louismaximepiton (changes requested):
Few things to tackle before merging, however, no blocking change since we already agreed on most of them.

site/content/docs/5.2/components/button-group.md
julien-deramond:
```suggestion
This variant **using and mixing colored button backgrounds** should not be used because it does not respect the Orange Design System specifications. In button groups, you should only use the button variant that uses `.btn .btn-secondary`.
```

site/content/docs/5.2/components/dropdowns.md
julien-deramond: OK dropped it. Added a comment in the other PR

site/content/docs/5.2/forms/validation.md
```diff
@@ -111,6 +111,12 @@ Custom feedback styles apply custom colors, borders, focus styles, and backgroun
 
 ## Browser defaults
 
+{{< ods-incompatibility-alert >}}
+This form variant, **with browser default validation feedback messages**, should not be used because it does not respect the Orange Design System specifications.
+
+Instead, please refer to our Boosted [Custom styles]({{< docsref "/forms/validation#custom-styles" >}}) section. You can also refer to the [Forms](https://system.design.orange.com/0c1af118d/p/88ab5b-forms/b/599459) guidelines and to the [Pages](https://system.design.orange.com/0c1af118d/p/20500e-form/b/16bb53) form examples on the Orange Design System website.
+{{< /ods-incompatibility-alert >}}
+
 Not interested in custom validation feedback messages or writing JavaScript to change form behaviors? All good, you can use the browser defaults. Try submitting the form below. Depending on your browser and OS, you'll see a slightly different style of feedback.
```
MewenLeHo, line 120: Still think this one is a bit confusing: 'should not be used [...]' and just after 'All good, you can use [it]'

julien-deramond: OK dropping it from this PR

MewenLeHo (approved):
Ok for me!  🚀

site/content/docs/5.2/forms/input-group.md
isabellechanclou: Why did you set this question as resolved? Is the question still valid or did you decided to ignore it at the end? It's just to know if I modify the callout message by adding a link towards Boosted Select page.

site/content/docs/5.2/forms/input-group.md
isabellechanclou: Here, do you want to add a link towards Boosted Select page too? Here, unless I missed something, I don't think it's necessary to do that.

site/content/docs/5.2/forms/input-group.md
julien-deramond: The issue here is that we don't explain why these variants don't respect ODS specifications. Is it a problem of layout? Why the third one works and not the fourth one. Honestly, even by knowing the project pretty well, I don't understand why for the latter.

site/content/docs/5.2/forms/validation.md
isabellechanclou: @MewenLeHo: any suggestion to fix that?
Possibilities we identified:
1- place this callout just after the "Browser defaults" title to make the user stop reading what is below the callout. This one was agreed on.
2- modify the text "Not interested in custom [...] feedback."

Other possibility:
3- place this callout after the text "Not interested in custom [...] feedback." and just above the illustration ?

louismaximepiton: We should maybe try the third solution since it doesn't break the Bootstrap text ?

isabellechanclou: Let's go for this 3rd solution. I'm doing it in PR #1614.

PR #1714, opened 2022-12-16
feat(range): provide an example to display the value

julien-deramond:
### Description

This PR provides a new example in the input range documentation to show how to display the value as a label for better usability.

Regarding a11y, here is how I defined the HTML structure and attributes:
* ~~Avoided a double `<label>` and chose a `<span>`~~
* Tested with VoiceOver, the value was already announced so I chose to `aria-hidden="true"` it to avoid a double announce for the same value.
* Finally used what's provided by a11y expert described in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1714#issuecomment-1377055559

- [x] Check the rendering compared to the [ODS design guidelines](https://system.design.orange.com/0c1af118d/p/18fe7f-slider/b/014486)
- [x] Auto-check HTML markup for a11y (will need a real a11y review after that)
- [x] Enhance documentation (split HTML code and JS?)

### Types of change

### Live previews

* https://deploy-preview-1714--boosted.netlify.app/docs/5.3/forms/range/#usability

### Checklist

- (NA) I have added tests to cover my changes


Aniort:
after some checkings, i think that we should use an output tag, as follow:
```
<div class="d-flex w-100 justify-content-between">
  <label for="customRange4" class="form-label">Example range</label>
  <output id="customRange4ValueText" for="customRange4" class="fw-bold"></span>
</div>
<input type="range" class="form-range" min="0" max="100" step="1" id="customRange4" onchange="updateLabelValue()">

```
it's seems to work well with screen readers

julien-deramond:
> after some checkings, i think that we should use an output tag, as follow:
> 
> ```
> <div class="d-flex w-100 justify-content-between">
>   <label for="customRange4" class="form-label">Example range</label>
>   <output id="customRange4ValueText" for="customRange4" class="fw-bold"></span>
> </div>
> <input type="range" class="form-range" min="0" max="100" step="1" id="customRange4" onchange="updateLabelValue()">
> ```
> 
> it's seems to work well with screen readers

It confirms what we talked about with @louismaximepiton in private. Thanks for the feedback @Aniort. I'll make some changes to go with it!

julien-deramond:
@Aniort, this PR's ready for final a11y review based on https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1714#issuecomment-1377055559. Should be quick to double-check 😄

Aniort:
to avoid a double vocalization with both screen readers Jaws & NVDA please add ariahidden to output tag: 
 <output id="customRange4ValueText" class="fw-bold" for="customRange4" aria-hidden="true">xx</output>
and it'll be ok !

site/content/docs/5.3/forms/range.md
```diff
@@ -50,6 +50,28 @@ By default, range inputs "snap" to integer values. To change this, you can speci
...
+For better usability, it is recommended most of the time to display the current selected value.
+
+This requires extra JavaScript code.
+
+{{< example >}}
+<div class="d-flex w-100 justify-content-between">
+  <label for="customRange4" class="form-label">Example range</label>
+  <output id="customRange4ValueText" for="customRange4" class="fw-bold" aria-hidden="true"></output>
+</div>
+<input type="range" class="form-range" min="0" max="100" step="1" id="customRange4" onchange="updateLabelValue()">
```
louismaximepiton, line 64: Should we have `onchange` or maybe just add a listener on it (since we introduce Js anyway)?

julien-deramond: Sorry but I don't understand the comment 🤔 
You want to change the `onchange` by something else?

julien-deramond: Interesting idea! I'll try to dig a bit to check if we can do easily something more reusable with a loop and id management.

julien-deramond: @louismaximepiton What do you think about the following?

```html
<!-- To test with another use case in the range page -->
<div class="d-flex w-100 justify-content-between mt-5">
  <label for="customRange5" class="form-label">Example range</label>
  <output id="customRange5ValueText" for="customRange5" class="fw-bold" aria-hidden="true"></output>
</div>
<input type="range" class="form-range" min="0" max="100" step="1" id="customRange5" onchange="updateLabelValue()">

<script>
  function updateLabelValue() {
    if (this.id === undefined) {
      return
    }

    document.getElementById(this.id + "ValueText").innerHTML = document.getElementById(this.id).value;
  }

  window.addEventListener('load', function () {
    Array.from(document.getElementsByClassName('form-range')).forEach(function (el) {
      if (document.getElementById(el.id + "ValueText")) {
        el.addEventListener('input', updateLabelValue)
        updateLabelValue.call(el)
      }
    })
  })
</script>
```

julien-deramond: > Remove the onchange call in the html but seems to be the case with your js

I'm dumb, forgot to remove the `onchange` in the HTML. That's why I had double calls and needed to add the `if (this.id === undefined) {` 🤦 

> Remove the dependency to a specific id value

Like your approach by forcing the `output` usage.

Just added a comment on top of the `<script>` to explain that it can be modified depending on the needs

So fixed via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1714/commits/735a0a9836732acb6234226233d2ad68e8906817

site/content/docs/5.3/forms/range.md
```diff
@@ -50,6 +50,37 @@ By default, range inputs "snap" to integer values. To change this, you can speci
...
+## Usability
+
+For better usability, it is recommended most of the time to display the current selected value.
+
+This requires extra JavaScript code.
+
+{{< example >}}
+<div class="d-flex w-100 justify-content-between">
+  <label for="customRange4" class="form-label">Example range</label>
+  <output id="customRange4ValueText" for="customRange4" class="fw-bold" aria-hidden="true"></output>
```
louismaximepiton, line 62:
```suggestion
  <output for="customRange4" class="fw-bold" aria-hidden="true"></output>
```

louismaximepiton (approved):
One thing and we're good

PR #1825, opened 2023-02-08
Fix quantity selector SELECTOR_INPUT_GROUP in JS

MewenLeHo:
### Description

Bug reported because of a too imprecise js selector.
Found a better solution to fix it without any change in the markup to prevent any possible problems for people already using this component.

### Types of change

### Live previews

[https://deploy-preview-1825--boosted.netlify.app/docs/5.3/forms/quantity-selector/](https://deploy-preview-1825--boosted.netlify.app/docs/5.3/forms/quantity-selector/)

#### Contribution

#### Accessibility

#### Design

#### Development

- [x] I have added JavaScript unit tests to cover my changes

#### Documentation

- (NA) My change introduces changes to the documentation and/or I have updated the documentation accordingly

- (NA) My change introduces changes to the migration guide
- [x] Manually run [BrowserStack tests](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/actions/workflows/browserstack.yml)
- [x] Manually test browser compatibility with BrowserStack (Chrome >= 60, Firefox >= 60 (+ ESR), Edge, Safari >= 12, iOS Safari, Chrome & Firefox on Android)
- [ ] Code review
- (NA) Design review
- (NA) A11y review

#### After the merge

### DIFF
--- a/js/src/quantity-selector.js
+++ b/js/src/quantity-selector.js
@@ -140,7 +140,7 @@ EventHandler.on(document, EVENT_CLICK_DATA_API, SELECTOR_STEP_UP_BUTTON, Quantit
 EventHandler.on(document, EVENT_CLICK_DATA_API, SELECTOR_STEP_DOWN_BUTTON, QuantitySelector.StepDown)
 
 EventHandler.on(window, EVENT_LOAD_DATA_API, () => {
-  for (const el of SelectorEngine.find(SELECTOR_QUANTITY_SELECTORP)) {
+  for (const el of SelectorEngine.find(SELECTOR_QUANTITY_SELECTOR)) {
     QuantitySelector.getOrCreateInstance(el).ValueOnLoad(el)
   }
 })


MewenLeHo:
Another possible solution maybe closer to what we have done morerecently:
- adding a `data-bs-step="quantity-selector` to the `.input-group` element
- use `'[data-bs-step="quantity-selector"]'` as a value for `const SELECTOR_INPUT_GROUP`

js/src/quantity-selector.js
```diff
@@ -26,7 +26,7 @@ const EVENT_CLICK_DATA_API = `click${EVENT_KEY}${DATA_API_KEY}`
 const SELECTOR_STEP_UP_BUTTON = '[data-bs-step="up"]'
 const SELECTOR_STEP_DOWN_BUTTON = '[data-bs-step="down"]'
 const SELECTOR_COUNTER_INPUT = '[data-bs-step="counter"]'
-const SELECTOR_INPUT_GROUP = '.input-group'
+const SELECTOR_INPUT_GROUP = '.input-group.quantity-selector'
```
louismaximepiton, line 29: Should we rename `SELECTOR_INPUT_GROUP` to `SELECTOR_QUANTITY_SELECTOR`?

louismaximepiton (commented):
> Another possible solution maybe closer to what we have done recently:
    - adding a data-bs-step="quantity-selector to the .input-group element
    - use '[data-bs-step="quantity-selector"]' as a value for const SELECTOR_INPUT_GROUP
But it will introduce in change in the markup so...

IMHO, we should keep the markup as-is since it doesn't bring much to change.

js/src/quantity-selector.js
MewenLeHo: Why not, could be a tiny bit more clear.

louismaximepiton (approved):
LGTM 🚀

julien-deramond:
Did you manage to reproduce the issue locally? If so, could you please provide some info to create a reproducable env?

MewenLeHo:
> Did you manage to reproduce the issue locally (IDK with an Angular project for example)? If so, could you please provide some info to create a reproducable env?

No I didn't create a new project. Instead I just udes the 'Get started' example given by Angular and I tweaked it a bit.

When using `<script src="https://cdn.jsdelivr.net/npm/boosted@5.2.3/dist/js/boosted.bundle.min.js" integrity="sha384-MANW37RG4MpFWPMCcNZBnvSobOkBpIGlbBkEzTtD4FbbOzJXbW8TddND1ak2lfsB" crossorigin="anonymous"></script>`:
[https://stackblitz.com/edit/angular-asyifs?file=src%2Findex.html](https://stackblitz.com/edit/angular-asyifs?file=src%2Findex.html)

There is indeed the foollowing error message in the console: 
```
Uncaught TypeError: Cannot read properties of null (reading 'getAttribute')
    at In.ValueOnLoad (quantity-selector.js:47:30)
    at quantity-selector.js:144:46
    at i (event-handler.js:98:15)
```

And when using the fix (i.e. by using `<script src="https://deploy-preview-1825-boosted.netlify.app/docs/5.3/dist/js/boosted.bundle.min.js"></script>`: [https://stackblitz.com/edit/angular-asyifs?file=src%2Findex.html](https://stackblitz.com/edit/angular-asyifs?file=src%2Findex.html) then the error message is gone.

But in my opinion the problem is not related to Angular. For me it was more likely related to the not precise enough selector used inside the JS onload event.

julien-deramond:
> But in my opinion the problem is not related to Angular. For me it was more likely related to the not precise enough selector used inside the JS onload event.

Yep I agree. Just wanted to be sure to fix it also in this specific use case which was the original issue.
Thanks for the repo steps, I'll double-check while reviewing the PR.

julien-deramond (approved):
Tested it locally. LGTM 🚀

PR #1870, opened 2023-03-01
Migration: small fix

louismaximepiton:

### Related issues

NA

### Description

Group carousel changes

### Types of change


julien-deramond (approved):
Good catch!

PR #1904, opened 2023-03-15
Cards: Some a11y fixes

louismaximepiton:

### Related issues

NA

### Description

Removing useless `role` and inverting `<a>`s and `<button>`s seems more logical from usability point of view..

### Motivation & Context

Some accessibility issues were spotted.

### Types of change

### Live previews

- <https://deploy-preview-1904--boosted.netlify.app/docs/examples/cards>
- <https://deploy-preview-1904--boosted.netlify.app/docs/examples/cards-rtl>

#### Contribution

#### Accessibility

#### Design

#### Development

- (NA) I have added JavaScript unit tests to cover my changes
- (NA) I have added SCSS unit tests to cover my changes

#### Documentation

- (NA) My change introduces changes to the migration guide
- [ ] Code review
- [ ] Design review
- [ ] A11y review

#### After the merge


site/content/docs/5.3/examples/cards-rtl/index.html
```diff
@@ -19,6 +19,16 @@ <h1 class="visually-hidden">أمثلة على البطاقة</h1>
     </symbol>
   </svg>
 
+  <div class="alert alert-info mb-0" role="alert">
+    <span class="alert-icon"><span class="visually-hidden">Information</span></span>
```
MewenLeHo, line 23:
```suggestion
    <span class="alert-icon"><span class="visually-hidden">Accessibility information</span></span>
```

site/content/docs/5.3/examples/cards-rtl/index.html
```diff
@@ -19,6 +19,16 @@ <h1 class="visually-hidden">أمثلة على البطاقة</h1>
...
+    <div>
+      <h4 class="alert-heading">Buttons on the last card of each line</h4>
+      <p>Please make sure to use an <code>`a`</code> for buttons if they link to another page.</p>
```
MewenLeHo, line 26: Or if you don't like first comment:

```suggestion
<p>Please make sure to use an <code>`a`</code> for buttons if they link to another page for better accessibility.</p>
```

MewenLeHo, line 26: Should we add a part about the correct HTML element to use if users don't want to make a link to another page? Or do you will they will know it implicitly?

louismaximepiton: I don't think it's necessary since it's only replacing the `button` by `a href="..."`

MewenLeHo (commented):
Ok for the changes on the example page.
Juste one last question: could it be useful to add this warning in the doc too, just like we usually do, and not only on the examples?

louismaximepiton:
> Just one last question: could it be useful to add this warning in the doc too, just like we usually do, and not only on the examples?

I don't know, it might be useful on certain pages and less on other. Maybe a general callout inside `accessibility.md` or `getting-started.md`, might do the trick ?

site/content/docs/5.3/examples/cards-rtl/index.html
```diff
@@ -19,6 +19,16 @@ <h1 class="visually-hidden">أمثلة على البطاقة</h1>
     </symbol>
   </svg>
 
+  <div class="alert alert-info mb-0" role="alert">
```
julien-deramond, line 22: TBH I'm not a big fan of this callout. We tend to avoid having callouts in example pages since it's not really the documentation. I still don't like too much the one in title bars.
Moreover, the difference between a button and a button-link is already explained in the docs. IMO we don't have to explain it everywhere everytime.
Thoughts here? Do you find it mandatory?

louismaximepiton: IMHO, it might be a bit too much indeed since it seems to be basic web. 
I've set it because of a11y team revendications but I'd agree to remove it on my side. Then maybe it's a bit more tricky here, what would you think about inverting the favorite and share to buttons and the actual button to a link ? Might feel a bit better in here.

Note to myself: Not a big fan using `role="alert"` on a static element that don't need much attention.

MewenLeHo:
> > Just one last question: could it be useful to add this warning in the doc too, just like we usually do, and not only on the examples?
> 
> I don't know, it might be useful on certain pages and less on other. Maybe a general callout inside `accessibility.md` or `getting-started.md`, might do the trick ?

Unfortunately I think that almost no one will have a look on these two pages so...
I honestly think that we should not add any callout because all developers must know the difference between a `link` and a `button`. But we have one to explain it in `dropdowns.md` so it may be usefull after all. 🤔

MewenLeHo (approved):
LGTM 🚀

site/layouts/shortcodes/card.html
```diff
@@ -37,27 +37,27 @@ <h3 class="card-title{{ if and $metadata (not $caption) }} mb-3{{ end }}">{{ if
...
           </svg>
           <span class="visually-hidden">{{ if $rtl }}يشارك{{ else }}Share{{ end }}</span>
-        </a>
+        </button>
       </li>
     </ul>
     {{- end -}}
     {{- if $action -}}
-    <button class="btn btn-secondary{{ if eq $background "dark" }} btn-inverse{{ end }}" aria-label="{{ if $rtl }}وصف الزر{{ else }}Button description{{ end }}" title="{{ if $rtl }}وصف الزر{{ else }}Button description{{ end }}">{{ if $rtl }}زر{{ else }}Button{{ end }}</button>
+    <a href="#" class="btn btn-secondary{{ if eq $background "dark" }} btn-inverse{{ end }}" aria-label="{{ if $rtl }}وصف الزر{{ else }}Button description{{ end }}" title="{{ if $rtl }}وصف الزر{{ else }}Button description{{ end }}">{{ if $rtl }}زر{{ else }}Button{{ end }}</a>
```
julien-deramond, line 60: If we consider that most of the time we will have links instead of actions, shouldn't we use "Link description" as labels?

julien-deramond (changes requested):
A note must be added in the migration guide to inform that the example has changed (and maybe other places - I don't remember where the shortcode is used).

site/layouts/shortcodes/card.html
```diff
@@ -37,27 +37,27 @@ <h3 class="card-title{{ if and $metadata (not $caption) }} mb-3{{ end }}">{{ if
     <p class="card-text fw-bold text-{{ if eq $background "dark" }}light{{ else }}muted{{ end }}">{{ if $rtl }}البيانات الوصفية{{ else }}Meta data{{ end }}</p>
     {{- end -}}
     {{- if $buttons -}}
-    <ul class="nav flex-row gap-3">
+    <ul class="nav flex-row gap-3 lh-1">
       <li>
-        <a href="#">
-          <svg xmlns="http://www.w3.org/2000/svg" width="1.25rem" height="1.25rem" role="img" focusable="false" aria-hidden="true">
+        <button class="btn btn-icon btn-link btn-inverse p-0 border-0">
```
julien-deramond, line 42: `.btn-inverse` shouldn't be there only when `{{ if eq $background "dark" }}`?

site/layouts/shortcodes/card.html
```diff
@@ -37,27 +37,27 @@ <h3 class="card-title{{ if and $metadata (not $caption) }} mb-3{{ end }}">{{ if
...
             <use xlink:href="#heart"/>
           </svg>
           <span class="visually-hidden">{{ if $rtl }}مفضل{{ else }}Favorite{{ end }}</span>
-        </a>
+        </button>
       </li>
       <li>
-        <a href="#">
-          <svg xmlns="http://www.w3.org/2000/svg" width="1.25rem" height="1.25rem" role="img" focusable="false" aria-hidden="true">
+        <button class="btn btn-icon btn-link btn-inverse p-0 border-0">
```
julien-deramond, line 50: `.btn-inverse` shouldn't be there only when `{{ if eq $background "dark" }}`?

julien-deramond:
Changed few things in the migration guide via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1904/commits/69d3a318a4c2534c3d9c459bd7099fa5a5a87511:
- Moved the new content in the "Examples" section since this modification only impacts the examples
- Wording to precise a bit the modified things
- Dropped the `.text-warning-emphasis` and `.bg-warning-subtle` that doesn't work in dark mode

julien-deramond (approved):

PR #1938, opened 2023-04-04
Merge from 74c6f2b to ???

julien-deramond:
- [x] https://github.com/twbs/bootstrap/commit/74c6f2bcfb1ca3391c2134e4e3425409b1c11939

PR #1940, opened 2023-04-04
Merge from 74c6f2b to ac57661 (part 2)

julien-deramond:
Missing commit after https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1938


dist/css/boosted.css
```diff
@@ -2132,7 +2132,7 @@ h2, .h2,
   box-shadow: inset 0 0 0 9999px var(--bs-table-accent-bg);
 }
 .table > thead > tr {
-  border-bottom-width: calc(var(--bs-border-width) * 0.5 * 2);
+  border-bottom-width: calc(calc(var(--bs-border-width) * 0.5) * 2);
```
louismaximepiton, line 2135: We have double calcs here, don't know if this is mandatory or not

louismaximepiton (approved):
LGTM

dist/css/boosted.css
julien-deramond: Comes from `calc(#{$table-border-width} * 2)`, I'd say it's mandatory in the usage in `_tables.scss` because you can't know whether `$table-border-width` is already a `calc` or not. But in the end CSS file, agree that it's useless. IDK if it can be simplified by a plugin, tool, etc. 🤷

PR #1986, opened 2023-04-13
Prototyping new headings values

julien-deramond:
> **Warning**
> This PR is a prototype around new Typography tokens

The idea here is to try to make match [Typography tokens](https://system.design.orange.com/0c1af118d/p/68dbd8-typography/b/953f29) and HTML headings:

* `<h1>` being "Title L"
* `<h2>` being "Headline L"
* `<h3>` being "Headline M"
* `<h4>` being "Title M"
* `<h5>` being "Headline S"
* `<h6>` being "Title S"

Not handled in the PR yet, but the `.display-*` classes would have the same values as headings.

### Live previews

- https://deploy-preview-1986--boosted.netlify.app/docs/5.3/examples/cards/

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -148,27 +148,56 @@ hr {
 
 h1 {
   @extend %heading;
-  @include font-size($h4-font-size);
-  line-height: $h4-line-height;
+  @include font-size($h1-font-size);
+  // TODO: line-height
 
   /* rtl:remove */
-  letter-spacing: $h4-spacing;
+  // TODO: letter-spacing
 }
 
 h2 {
   @extend %heading;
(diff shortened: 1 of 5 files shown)


julien-deramond:
OK so the first test shows an issue because it doesn't work on mobile (down sm breakpoint) because the order is different:

julien-deramond:
Note: [Typography Tokens](https://system.design.orange.com/0c1af118d/p/68dbd8-typography/b/953f29) values have changed today. "Headline L" and "Title L" now have the same value for Desktop (50px) so the idea developed in this PR won't work.

julien-deramond:
Closing it. Keeping the branch in case of we need to work on it later.

PR #1993, opened 2023-04-17
Chore merge main from 4f24144 to ??

julien-deramond:
- [x] https://github.com/twbs/bootstrap/commit/4f24144a31996339daa5e4fc972f9644f12d07ca


louismaximepiton (approved):
Maybe we need to write somewhere to use oxipng on images (in the wiki maybe?)
~~For the images that don't pass through this, maybe try to pass them inside https://iloveimg.com compression system before?~~ Seems that it transform those to jpg that we don't want, so let's keep them as-is.

PR #2032, opened 2023-05-12
Docs > Content > Tables: Add design callout messages

isabellechanclou:
 
 ### Related issues
 
 #1614 
 
 ### Description
 
 After a design review of all the table variants, it came out that 2 design callout messages needed to be added before variants (accented tables and nesting) that should not be used according to the Orange design system.
 
 ### Motivation & Context
 
 Change required to prevent Boosted user from using in their app component variants that are not suitable in the Orange Design System.
 
 ### Types of change
 
 ### Live previews
 
 - <https://deploy-preview-2032--boosted.netlify.app/docs/5.3/content/tables/#striped-columns>
  - <https://deploy-preview-2032--boosted.netlify.app/docs/5.3/content/tables/#nesting>
  
 ### Checklist
 
 
 
 
 #### Contribution
 
 #### Accessibility
 
 #### Design
 
 #### Development
 
 - [ ] I have added JavaScript unit tests to cover my changes
 - [ ] I have added SCSS unit tests to cover my changes
 
 #### Documentation
 
 - [ ] My new component is added in Storybook
 - [ ] My new component is compatible with RTL
 - [x] Code review
 - [ ] Design review
 - [ ] A11y review
 
 #### After the merge
 

site/content/docs/5.3/content/tables.md
```diff
@@ -80,6 +80,10 @@ Use contextual class to color tables, table rows or individual cells.
 
 ## Accented tables
 
+{{< design-callout-alert info >}}
+It is possible to use striped rows **or** striped columns in your tables, but you **can't use both in a same table**.
```
MewenLeHo, line 84: Should we explain why? 'because it does not respect the Orange Design System specifications' for instance?

julien-deramond: We are not showing examples where we use both in the same table. Is this callout really needed?

isabellechanclou: > Should we explain why? 'because it does not respect the Orange Design System specifications' for instance?

Good point Mewen. Indeed, it would be better to write: "It is possible to use striped rows **or** striped columns in your tables, but you **can't use both in a same table** because  it does not respect the Orange Design System specifications."

isabellechanclou: > We are not showing examples where we use both in the same table. Is this callout really needed?

We thought this design callout message would be necessary to avoid users using both horizontal and vertical striped at the same time, knowing that in the following section "Striped columns", there is an example that could be misleading: it shows 2 striped columns + a striped line (but not the entire line).

hannahiss: > there is an example that could be misleading: it shows 2 striped columns + a striped line (but not the entire line).
Yes, this example is really confusing, it is just striped columns, but in the third line, 2 cells are merged with a `colspan="2"`. (example from Bootstrap)

julien-deramond: IMHO it's obvious that one shouldn't use both at the same time. I vote for not adding this callout. Moreover, the doc doesn't show that they can be used at the same time, it only shows the rendering when there are merged cells.

isabellechanclou: It this `it only shows the rendering when there are merged cells.` that can be confusing to me and motivated my adding the design callout.

julien-deramond (approved):
Still not a big fan of this callout, but the majority prevails. Let's ship it.

Issue #999, opened 2021-12-22
ODS > Promotions


julien-deramond:
## Requirement

**For** a Developer 
**Who** is creating an Orange Website
**The** Promotions Integration
**is an** automated component
**that** proposes the web site's Orange Promotions like Hero & Product Banners
**unlike** the default bootstrap framework
**Boosted** will provide this ready to use component compatible with the Orange Design and the accessibility requirements

## Description

Users can build a form into the website with all the following:
**Hero Banners**
* Sizes adaptation (Large, Medium & Small Desktop / Tablet Landscape & Portrait / Mobile Landscape & Portrait )
* Banner with white vertical panel
* Banner with OBS gray vertical panel
* Banner with black vertical panel
* Banner with white vertical panel
* Banner with OBS gray vertical panel

**Product & Services Banners**
*  Sizes adaptation (Large, Medium & Small Desktop / Tablet Landscape & Portrait / Mobile Landscape & Portrait )
* Banner with Title, 4 icons, labels & descriptions on black
* Banner with Title, 4 icons, labels & descriptions on white
* Banner with Title, 4 icons, labels & descriptions on gray
* Banner with Title, 4 icons & labels on black
* Banner with Title, 4 icons & labels on white
* Banner with Title, 4 icons & labels on gray
* Banner with Title, 6 icons & labels on black
* Banner with Title, 6 icons & labels on white
* Banner with Title, 6 cons & labels on gray
* Banner with 4 icons & labels on black
* Banner with 4 icons & labels on white
* Banner with 4 icons & labels on gray
* Banner with 6 icons & labels on black
* Banner with 6 icons & labels on white
* Banner with 6 icons & labels on gray

## References

Orange Design System: [Promotions
](https://system.design.orange.com/0c1af118d/p/925608-promotions)

## Implementation

- [x] #1000
- [x] #1001

PR #666, opened 2021-05-11
feat(tags)

vogloblinsky:
PR for this feature https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/552

All states and use-cases are implemented.

Review and discussion open ;)

PR #2035, opened 2023-05-15
Fix a11y problems on star rating component

### LINKED ISSUE #1764: Enhance star ratings with readonly/disabled states
Our Star rating component has a disabled version that can be found here: https://boosted.orange.com/docs/forms/checks-radios/#disabled-2.

The associated description begins with "Make star ratings look inactive inside or outside a form [...]".

IMO star rating should be enhanded to propose two separate states:
* `readonly` would have the same rendering as what we show in https://boosted.orange.com/docs/forms/checks-radios/#disabled-2
* `disabled` would show the orange stars as dark gray stars

A readonly state is different than a disabled state.

There's also in this spirit something to check with the readonly state because inputs in Bootstrap/Boosted are focusable in readonly state contrary to disabled state. In readonly, when it's focused, the value could be read by a11y tools as well.

### LINKED ISSUE #2005: A11y problems on star rating component
### Prerequisites

### Describe the issue

hello,
2 a11yf issues on star rating comp. :
- inactive stars must have a sufficient contrast ratio cause they provide a piece of information ('total number of stars):
- there's a title attribute in labels, we need to add also an aria-label attribute with the same content
that's all folks !

### Reduced test cases

https://boosted.orange.com/docs/5.2/forms/checks-radios/#star-rating

Windows

Chrome, Firefox

5.2

MewenLeHo:
### Related issues

Closes #2005 
Closes #1764 

### Description

Fix a11y problems on star rating component.

### Motivation & Context

Improve a11y.

### Types of change

### Live previews

- <https://deploy-preview-2035--boosted.netlify.app/docs/5.3/forms/checks-radios/#star-rating>
- <https://deploy-preview-2035--boosted.netlify.app/docs/5.3/forms/checks-radios/#readonly>
- <https://deploy-preview-2035--boosted.netlify.app/docs/5.3/forms/checks-radios/#disabled-2>

#### Contribution

#### Accessibility

#### Design

- (NA) My change is compatible with responsive display

#### Development

- (NA) I have added JavaScript unit tests to cover my changes
- (NA) I have added SCSS unit tests to cover my changes

#### Documentation

- [x] Manually run [BrowserStack tests](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/actions/workflows/browserstack.yml)
- [x] Manually test browser compatibility with BrowserStack (Chrome >= 60, Firefox >= 60 (+ ESR), Edge, Safari >= 12, iOS Safari, Chrome & Firefox on Android)
- [ ] Code review
- [ ] Design review
- [ ] A11y review

#### After the merge


MewenLeHo:
To do list:

- [ ] Add missing `aria-label` attribute matching the `title` attribute
- [ ] Change background-colors for inactive stars to ensure sufficient contrast ratio
- [ ] Update migration guide

MewenLeHo:
Coming from Spec Meeting - May, 16:

- Selected option: stop using solid stars for inactive stars and replace them with shape stars	
- Selected colors: on light background: `#666` / on dark background: `#999`

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -422,45 +422,65 @@ Add `.star-rating-dark` to the `.star-rating` for a dark variant.
...
+  <input type="radio" id="good5" name="rating" value="4" class="visually-hidden">
+  <label for="good5" title="Good"></label>
 
-  <input type="radio" id="excellent4" name="rating" value="5" class="visually-hidden">
-  <label for="excellent4" title="Excellent"></label>
+  <input type="radio" id="excellent5" name="rating" value="5" class="visually-hidden">
+  <label for="excellent5" title="Excellent"></label>
 </fieldset>
-<span class="visually-hidden">rated 3 stars out of 5</span>
+<p class="visually-hidden">Results relevance: rated 3 stars out of 5</p>
```
hannahiss, line 483: Why use this more complicated method? 
Isn't it ok to remove `aria-hidden="true"` and let screen reader vocalize the radio buttons?
@Aniort an opinion on this?

site/content/docs/5.3/migration.md
```diff
@@ -154,6 +154,12 @@ Learn more by reading the new [color modes documentation]({{< docsref "/customiz
 
 - <span class="badge bg-success">New</span> Our range documentation now provides an example which displays the current selected value for better usability, but with some extra JavaScript code.
 
+- <span class="badge bg-success">New</span> Star rating has been slightly changed for accessibility purpose; an aria-label attribute matching the title attribute was added and color contrast of inactive stars has been updated. Please reflect this modification into your websites. Our documentation also includes a disabled and a readonly state for this form element.
```
hannahiss, line 157: to adapt if there is any changes in code...

hannahiss (changes requested):
Some small notes and some questions about accessibility with screen reader...

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -368,19 +368,19 @@ Star rating system is built on top of radios. Simply add `.star-rating` to a `<f
     <legend class="visually-hidden">Results relevance</legend>
 
     <input type="radio" id="terrible" name="rating" value="1" class="visually-hidden">
-    <label for="terrible" title="Terrible"></label>
+    <label for="terrible" title="Terrible" aria-label="Terrible"></label>
```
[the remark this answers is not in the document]
Aniort: in fact, here aria-label is not the best solution, your suggestion is better :
<label for="bad" title="Bad"><span class="visually-hidden">bad</span></label>
@MewenLeHo  it's time to correct ;)

site/content/docs/5.3/forms/checks-radios.md
Aniort: @hannah, we have to take time during a Web sp$ec review so i can explain to you why we choose that method!

hannahiss: @Aniort ok thanks :-)

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -422,45 +422,65 @@ Add `.star-rating-dark` to the `.star-rating` for a dark variant.
...
+{{< example >}}
+<div class="star-rating">
+  <p class="visually-hidden">Results relevance: rated 3 stars out of 5</p>
+
+    <div aria-hidden="true">
+      <span id="terrible4" class="checked"></span>
+      <span id="bad4" class="checked"></span>
+      <span id="mixed4" class="checked"></span>
+      <span id="good4"></span>
+      <span id="excellent4"></span>
```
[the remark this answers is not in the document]
MewenLeHo: They are not necessary at all.
I kept them because it was a bit weird to have example 5 with id `xx4` but it's probably not a big deal.

scss/mixins/_star-rating.scss
```diff
@@ -1,5 +1,9 @@
 // scss-docs-start form-star-rating-mixin
-@mixin form-star-rating($color) {
-  background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 25 25'><path fill='#{$color}' d='m12.5 2.9 2.379 7.35 7.721-.007-6.25 4.536 2.392 7.346-6.242-4.547-6.242 4.547 2.392-7.346-6.25-4.536 7.721.007Z'/></svg>"));
+@mixin form-star-rating($color, $background-color) {
```
hannahiss, line 2: Maybe rename `$color` to `$border-color` for better understanding? Same below

hannahiss (changes requested):
Fix to do in migration guide 

+ I've been using Chrome with a big font size and stars size do not increase, and I can see a display problem (while it is ok on main branch): 
image

LM suggested maybe it comes from svg: it used to be  `viewBox='0 0 25 25'` and its now `width='25' height='25'`

MewenLeHo:
> 3. I've been using Chrome with a big font size and stars size do not increase, and I can see a display problem (while it is ok on main branch):
>    image
> 
> LM suggested maybe it comes from svg: it used to be `viewBox='0 0 25 25'` and its now `width='25' height='25'`


Not able to reproduce.
Could you please provide more informations @hannahiss?

hannahiss (approved):
Maybe rename attributes of form-star-rating mixin to $inner-color and $outer-color for better understanding, but OK otherwise 🚀

scss/mixins/_star-rating.scss
MewenLeHo: I gave it a second thought and I came with the idea that `$path-color` and `$fill-color` could be better than `$outer-color` and `$inner-color`. I think it's more explicit and easier to understand.

Any opinion @hannahiss?

hannahiss: I realize that I did not answer. One or another seems ok, it's better than `$color`/`$background-color` :-)

Aniort (approved):
good to me, OK

scss/forms/_star-rating.scss
```diff
@@ -3,31 +3,30 @@
...
 
     &:not(:first-of-type){
       margin-left: $form-star-margin-between;
     }
   }
 
-  > input:checked ~ label {
-    @include form-star-rating($gray-500);
+  > input:checked ~ label,
+  span {
```
louismaximepiton, line 23:
```suggestion
  .checked ~ span {
```
If we want to keep consistency between form and readonly version I think the markup should be:
```html
<div class="star-rating">
  <p class="visually-hidden">Star rating: rated 3 out of 5</p>
  <div aria-hidden="true">
    <span></span>
    <span></span>
    <span class="checked"></span>
    <span></span>
    <span></span>
  </div>
</div>
```
If this is commited and the one below, it might help using both markup anyway.

scss/forms/_star-rating.scss
```diff
@@ -3,31 +3,30 @@
...
+    @include form-star-rating($gray-700, $white);
   }
 
-  > input:checked + label {
-    @include form-star-rating($accessible-orange);
+  > input:checked + label,
+  .checked {
```
louismaximepiton, line 28:
```suggestion
  span.checked {
```
if the above one is commited

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -422,45 +422,65 @@ Add `.star-rating-dark` to the `.star-rating` for a dark variant.
...
+<div class="star-rating">
+  <p class="visually-hidden">Star rating: rated 3 out of 5</p>
+
+    <div aria-hidden="true">
+      <span class="checked"></span>
+      <span class="checked"></span>
+      <span class="checked"></span>
+      <span></span>
+      <span></span>
+    </div>
```
louismaximepiton, line 455: There's a shift right there and above.
```suggestion
  </div>
```

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -422,45 +422,65 @@ Add `.star-rating-dark` to the `.star-rating` for a dark variant.
...
+
```
louismaximepiton, line 456: To remove imo

scss/forms/_star-rating.scss
```diff
@@ -37,44 +36,85 @@
...
+    @include form-star-rating-sm($black, $black);
+  }
+
+  > input:hover ~ label {
+    @include form-star-rating-sm($gray-700, $white);
+  }
+
+  > input:hover + label {
+    @include form-star-rating-sm($black, $black);
+  }
```
louismaximepiton, line 119: Do we need all of those rules ?
Maybe yes for the maintainability.
Otherwise we might be able to remove the rules with filled stars.

scss/forms/_star-rating.scss
```diff
@@ -37,44 +36,85 @@
...
 
 .star-rating-dark {
   > label {
-    @include form-star-rating($brand-orange);
+    @include form-star-rating($brand-orange, $brand-orange);
   }
 
   > input:checked ~ label {
-    @include form-star-rating($gray-700);
+    @include form-star-rating($gray-500, $black);
```
louismaximepiton, line 74:
```suggestion
    @include form-star-rating($gray-500, transparent);
```
```suggestion
    @include form-star-rating($gray-500, none);
```
I don't know if this should happen/work but might be better on gray backgrounds (if yes, spread it to the whole file with empty background stars)

louismaximepiton (changes requested):
FYI:
- The colors aren't good when using `.star-rating.star-rating-dark.star-rating-sm`.
- The Readonly version seems to have the cursor while hovering the div which is kind of weird here.
- The disabled version looks weird to me when using the dark version
- The disabled version looks weird to me since the last star is filled but not checked

Franco-Riccitelli:
For the disabled state, the stars should remain in outline and change to a different grey. 

On white: #CCCCCC
On black: #666666

I don't think it's necessary to have a "filled" disabled state colour as the component is inactive.
Apart from that it all looks good.

Screenshot 2023-08-01 at 10 57 23 am

scss/forms/_star-rating.scss
MewenLeHo: Not sure about this one, it will just cause a 'selector-no-qualifying-type' error 🤔

scss/forms/_star-rating.scss
MewenLeHo: I removed as many common rules as possible: rules about diabled state or the common set of rules for label and span at the beggining.

It may be better to keep it like it is since we now have multiple set of colors depending of size variant, light/dark mode and state. It may be the only way to keep enough flexibility inside this component.

scss/forms/_star-rating.scss
louismaximepiton: Ah you're right the rule above had more effect than expected. Actually wondering if this line is still needed

MewenLeHo:
> For the disabled state, the stars should remain in outline and change to a different grey.
> 
> On white: #CCCCCC On black: #666666
> 
> I don't think it's necessary to have a "filled" disabled state colour as the component is inactive. Apart from that it all looks good.
> 
> Screenshot 2023-08-01 at 10 57 23 am

Thanks for the review @Franco-Riccitelli 

But for me there is no reason to prevent a disabled star to be filled. For instance you may see the actual quotation but the component is disabled because you are not logged in. Or you don't have the right to change it. Or you changed the score but you remove a mandatory information so it become disabled. In fact I see many reasons why it can be possible to have checked stars in a disabled star rating component.

This component is working like radio buttons and checkboxes so why should we reset it when it becomes disabled?

If you are ok, I will need to know what color(s) you prefer for checked stars in a disabled star rating component please 😉

Franco-Riccitelli:
> > For the disabled state, the stars should remain in outline and change to a different grey.
> > On white: #CCCCCC On black: #666666
> > I don't think it's necessary to have a "filled" disabled state colour as the component is inactive. Apart from that it all looks good.
> > Screenshot 2023-08-01 at 10 57 23 am
> 
> Thanks for the review @Franco-Riccitelli
> 
> But for me there is no reason to prevent a disabled star to be filled. For instance you may see the actual quotation but the component is disabled because you are not logged in. Or you don't have the right to change it. Or you changed the score but you remove a mandatory information so it become disabled. In fact I see many reasons why it can be possible to have checked stars in a disabled star rating component.
> 
> This component is working like radio buttons and checkboxes so why should we reset it when it becomes disabled?
> 
> If you are ok, I will need to know what color(s) you prefer for checked stars in a disabled star rating component please 😉

Ok, then I suggest we use the same grey colour used for the outlines in the disabled state for the fill colour. Thanks.

Screenshot 2023-08-04 at 11 13 29 am

scss/mixins/_star-rating.scss
```diff
@@ -1,5 +1,9 @@
 // scss-docs-start form-star-rating-mixin
-@mixin form-star-rating($color) {
-  background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 25 25'><path fill='#{$color}' d='m12.5 2.9 2.379 7.35 7.721-.007-6.25 4.536 2.392 7.346-6.242-4.547-6.242 4.547 2.392-7.346-6.25-4.536 7.721.007Z'/></svg>"));
+@mixin form-star-rating($path-color, $fill-color) {
+  background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 25 25'><path fill='#{$fill-color}' fill-rule='evenodd' stroke='#{$path-color}' d='m12.5 4.523 2.016 6.227 6.542-.005-5.296 3.843 2.027 6.224L12.5 16.96l-5.289 3.852 2.027-6.224-5.296-3.843 6.542.005L12.5 4.523Z'/></svg>"));
+}
+
+@mixin form-star-rating-sm($path-color, $fill-color) {
+  background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'><path fill='#{$fill-color}' fill-rule='evenodd' stroke='#{$path-color}' d='M10 3.943 11.54 8.7l4.998-.004-4.046 2.936 1.548 4.755L10 13.444l-4.04 2.943 1.548-4.755-4.046-2.936L8.46 8.7 10 3.943Z'/></svg>"));
```
louismaximepiton, line 7:
```suggestion
  background-image: escape-svg(url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'><path fill='#{$fill-color}' stroke='#{$path-color}' d='M10 3.943 11.54 8.7l4.998-.004-4.046 2.936 1.548 4.755L10 13.444l-4.04 2.943 1.548-4.755-4.046-2.936L8.46 8.7 10 3.943Z'/></svg>"));
```

Trying to lighten a bit the bundle here

site/content/docs/5.3/forms/checks-radios.md
```diff
@@ -422,45 +422,64 @@ Add `.star-rating-dark` to the `.star-rating` for a dark variant.
...
+    <label for="good3" title="Good"><span class="visually-hidden">Good</span></label>
 
     <input type="radio" id="excellent3" name="rating" value="5" class="visually-hidden">
-    <label for="excellent3" title="Excellent"></label>
+    <label for="excellent3" title="Excellent"><span class="visually-hidden">Excellent</span></label>
   </fieldset>
 </form>
 {{< /example >}}
 
+### Readonly
```
louismaximepiton, line 442: It feels weird to have it here since it's not the same behavior and not the same purpose. Do you think we should have a proper section for this component (I mean the whole star rating) ?

scss/forms/_star-rating.scss
```diff
@@ -37,44 +39,130 @@
...
+    @include form-star-rating($gray-700, transparent);
   }
 
   > input:hover + label {
-    @include form-star-rating($black);
+    @include form-star-rating($black, $black);
+  }
+
+  &.disabled,
+  &:disabled {
```
louismaximepiton, line 54: It doesn't seem to work for small variants (light and dark).

scss/forms/_star-rating.scss
```diff
@@ -3,31 +3,33 @@
...
-  > input:checked ~ label {
-    @include form-star-rating($gray-500);
+  > label {
+    cursor: pointer;
   }
 
-  > input:checked + label {
-    @include form-star-rating($accessible-orange);
+  > input:checked ~ label,
+  .checked ~ span {
```
louismaximepiton, line 26: Maybe we should expand the readonly to the small variant and dark variant as well ?

louismaximepiton (changes requested):
Few things more to tackle for the readonly and disabled variants

scss/forms/_star-rating.scss
MewenLeHo: Yes, we don't need it any more.

site/content/docs/5.3/forms/checks-radios.md
MewenLeHo: It could be a good thing yes since we now have normal/small versions + light/dark versions + disabled state + readonly state etc. It's probably too much for a simple section in the `Checks & radios` page.

In addition the `disabled` state is supposed to be used outside of a form, in a card for instance, so there is no reason to keep it in the `Forms` section although the rest of the star rating thing is sill a form element. 🤔 

Should we split it? (🤢)

scss/forms/_star-rating.scss
```diff
@@ -37,44 +35,175 @@
...
+      @include form-star-rating($gray-700, transparent);
+    }
+
+    > input:checked + label {
+      @include form-star-rating($gray-700, $gray-700);
+    }
+  }
+
+  &.star-rating-sm {
+
```
louismaximepiton, line 161: Blank line to remove

louismaximepiton (commented):
3 blank lines to remove and we're good!

scss/forms/_star-rating.scss
```diff
@@ -37,44 +39,130 @@
...
+      @include form-star-rating($gray-500, transparent);
+    }
+
+    > input:checked + label {
+      @include form-star-rating($gray-500, $gray-500);
+    }
+  }
+
+  &.star-rating-sm {
+
```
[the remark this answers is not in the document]
louismaximepiton: Still to remove I guess

louismaximepiton (approved):
🚀

MewenLeHo:
Will be resolved by #2184

PR #2092, opened 2023-06-19
Workflows: Add automation for github project to pass an approved PR directly in good column

louismaximepiton:
 
 ### Related issues
 
 NA
 
 ### Description
 
When a PR is approved while needing either a design review, or an a11y review or both, the label of the PR needs to be updated and the associated card should move in the `Need Design/A11y Review` column.
When a PR is approved while not needing any design or a11y review, the associated card should move in the `Need Lead Dev Review` column.

:warning: **After the PR is merged**, we might need to update a bit the project workflows to remove the code review approved behavior.
 
 ### Motivation & Context
 
 Fully automate the new board.
 
 ### Types of change
 
 
 
 

MewenLeHo (changes requested):
🤔

MewenLeHo (changes requested):
🦝

PR #1931, opened 2023-03-29
Pagination: responsive part round 2

louismaximepiton:
_Note: Please transform `- [ ]` into `- (NA)` in the description when things are not applicable_

### Related issues

#1370.

### Description

Add something to handle the mobile responsive part

### Types of change

- Bug fix (non-breaking which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Refactoring (non-breaking change)
- Breaking change (fix or feature that would change existing functionality)

### Live previews

- <https://deploy-preview-{your_pr_number}--boosted.netlify.app/>

#### Contribution

- [ ] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

#### Accessibility

- [ ] My change follows accessibility good practices; I have at least run axe

#### Design

- [ ] My change respects the design guidelines defined in [Orange Design System](https://system.design.orange.com/0c1af118d/p/8118d1-web)
- [ ] My change is compatible with responsive display

#### Development

- [ ] My change follows the [developer guide](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/wiki/Developer-guide)
- [ ] I have added JavaScript unit tests to cover my changes
- [ ] I have added SCSS unit tests to cover my changes

#### Documentation

- [ ] My change introduces changes to the documentation and/or I have updated the documentation accordingly

### Checklist (for Core Team only)

- [ ] My change introduces changes to the migration guide
- [ ] My new component is added in Storybook
- [ ] My new component is compatible with RTL
- [ ] Manually run [BrowserStack tests](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/actions/workflows/browserstack.yml)
- [ ] Manually test browser compatibility with BrowserStack (Chrome >= 60, Firefox >= 60 (+ ESR), Edge, Safari >= 12, iOS Safari, Chrome & Firefox on Android)
- [ ] Code review
- [ ] Design review
- [ ] A11y review

#### After the merge

- [ ] Manually launch [Percy tests](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/actions/workflows/percy.yml)

PR #2102, opened 2023-06-23
Fix pagination: make previous and next buttons accessible

### LINKED ISSUE #2008: Pagination: Make the previous and next buttons accessible
### Prerequisites

- [X] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed feature requests

### Proposal

Possible ways to better the content:
- Add the icon directly in the HTML (which is the solution to use at the end since it'd be breaking)
- Handle differently the aria-label with a `.visually-hidden`

Feel free to express some other way to discuss about.

### Motivation and context

After the review of #1931, it appears that the `<` and `>` aren't accessible since without CSS, the buttons aren't focusable.

MewenLeHo:
### Related issues

Closes #2008 

### Motivation & Context

Improve a11y of pagination component

### Types of change

### Live previews

#### Contribution

- [ ] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

#### Accessibility

- [ ] My change follows accessibility good practices; I have at least run axe

#### Design

- [ ] My change respects the design guidelines defined in [Orange Design System](https://system.design.orange.com/0c1af118d/p/8118d1-web)
- [ ] My change is compatible with responsive display

#### Development

- [ ] I have added JavaScript unit tests to cover my changes
- [ ] I have added SCSS unit tests to cover my changes

#### Documentation

- [ ] My change introduces changes to the documentation and/or I have updated the documentation accordingly

- [ ] My new component is added in Storybook
- [ ] My new component is compatible with RTL
- [ ] Code review
- [ ] Design review
- [ ] A11y review

#### After the merge

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -154,7 +154,7 @@
   &:first-child .page-link {
     margin-right: var(--#{$prefix}pagination-margin-x-first-last);
 
-    &:not(:empty) {
+    &:not(:empty) :not([data-bs-a11y]) {
       padding-right: var(--#{$prefix}pagination-padding-end);
     }
 
--- a/site/content/docs/5.3/components/pagination.md
+++ b/site/content/docs/5.3/components/pagination.md
@@ -30,7 +30,7 @@ Looking to use an icon or symbol in place of text for some pagination links? Be 
 {{< example >}}
 <nav aria-label="Page navigation example with icons">
   <ul class="pagination">
-    <li class="page-item"><a class="page-link" href="#" aria-label="Previous"></a></li>
+    <li class="page-item"><a class="page-link" href="#" aria-label="Previous"><span class="visually-hidden" aria-hidden="true" data-bs-a11y>Previous</span></a></li>
     <li class="page-item"><a class="page-link" href="#" aria-label="Next"></a></li>
   </ul>
 </nav>


MewenLeHo:
To do after prototype validation

- [ ] Make the change everywhere
- [ ] Update migration guide

scss/_pagination.scss
```diff
@@ -154,7 +154,7 @@
   &:first-child .page-link {
     margin-right: var(--#{$prefix}pagination-margin-x-first-last);
 
-    &:not(:empty) {
+    &:not(:empty) :not([data-bs-a11y]) {
```
MewenLeHo, line 157: I still need to figure what is the best way to target it without using something that users will not use 🤔

MewenLeHo, line 157: Resolve side effect:
screenshot-deploy-preview-2102--boosted netlify app-2023 06 23-14_56_39

scss/_pagination.scss
```diff
@@ -204,6 +204,18 @@
   // End mod
 }
 
+// Boosted mod: special style for pagination with icons only
+.icons-only {
```
MewenLeHo, line 208: Possible solution to add correct style after the a11y fix.
Not a big fan of it since it needs a change in the markup.

Any opinion @louismaximepiton? Could it work for your work on responsive pagination?

MewenLeHo: Possible alternatives:

- Using CSS parent slecto `:has` (not supported by Firefox: https://caniuse.com/?search=%3Ahas)
- Using Js to alter the parent according to the content of the link? (🤢)

louismaximepiton: As said in private, it doesn't interfere with the responsive pagination.

Concerning the solution, we won't have much more solution than the one proposed here. 

Can you provide a solution using `.has-label` or equivalent to better understand the impact for the users and which one we should use ?

MewenLeHo: Link to the requested prototype: https://codepen.io/MewenLeHo/pen/LYXWqrJ?editors=1100

First prototype (see Netlify preview):
Need to add a class to the `ul` and add this new class to style file.

Second prototype (see Codepen):
Need to add a class to the each `a` and make some changes to the actual style.

Please note that I used `.has-label` in Codepen but we need to find a new name for this class since `.has-label` was used in V4 for another purpose.

louismaximepiton: Thanks for the protoype but it' not what I had in mind.

I had in mind a `.has-label` on the `ul` as well but with the same purpose as in v4 so the class should be added when you want to display the `Previous` ans `Next` labels. I don't know if it's feasible.

An other solution could be to remove our actual behavior with the `:empty` rule. and replace the simple text by `<span class="me-3">Previous</span>` but it could introduce a small shift.

louismaximepiton: Viewing this a month after, I'm wondering if there isn't any way to modify `_pagination.scss` in order to have no change for end-users. I feel it's feasible and I'm wondering if it's worth a try. Any thoughts on this ? Have you tried anything in this way ?

louismaximepiton: Tried it this way https://codepen.io/louismaximepiton/pen/zYMXqzp, do you think it's worth a try ?

MewenLeHo: @louismaximepiton: I made the change according to your prototype. Seems interresting because there is no need to change the markup (except the small change to add an accessible name for the icons _via_ the `visually-hidden` span around each icon).

I am currently looking for any potential side effect.

scss/_pagination.scss
```diff
@@ -43,6 +43,7 @@
   position: relative;
   box-sizing: content-box; // Boosted mod
   display: flex; // Boosted mod
+  gap: 10px; // Boosted mod
```
louismaximepiton, line 46: Maybe we should add a variable for this one ?

scss/_pagination.scss
```diff
@@ -56,6 +57,7 @@
   // Boosted mod: embed icon as mask-image
   @include button-icon(
     var(--#{$prefix}pagination-icon),
+    .625rem,
```
louismaximepiton, line 60: Same here, we should add a variable to better control the component

scss/_variables.scss
```diff
@@ -1616,7 +1616,7 @@ $pagination-transition:             $transition-focus !default;
 // Boosted mod
 $pagination-padding-end:            1.125rem !default;
 $pagination-icon:                   var(--#{$prefix}chevron-icon) !default;
-$pagination-icon-size:              subtract($spacer * 2, calc(var(--#{$prefix}border-width) * 2)) !default; // stylelint-disable-line function-disallowed-list
+$pagination-icon-size:              .875rem !default; // Boosted mod: instead of `subtract($spacer * 2, calc(var(--#{$prefix}border-width) * 2))`
```
louismaximepiton, line 1619:
```suggestion
$pagination-icon-size:              1.125rem !default;
```

scss/_pagination.scss
```diff
@@ -166,9 +164,6 @@
   }
 
   &:last-child .page-link {
-    &:not(:empty) {
-      padding-left: var(--#{$prefix}pagination-padding-end);
-    }
```
louismaximepiton, line 167: Remove blank line

scss/_pagination.scss
```diff
@@ -204,7 +199,6 @@
   // End mod
 }
 
-
```
louismaximepiton, line 207: Add this one if this one is on Bs side

scss/_pagination.scss
```diff
@@ -154,10 +156,6 @@
   &:first-child .page-link {
     margin-right: var(--#{$prefix}pagination-margin-x-first-last);
 
-    &:not(:empty) {
-      padding-right: var(--#{$prefix}pagination-padding-end);
```
louismaximepiton, line 158: Do we still need this variable ?

scss/_variables.scss
```diff
@@ -1577,8 +1577,8 @@ $dropdown-dark-header-color:        $white !default; // Boosted mod
 // Pagination
 
 // scss-docs-start pagination-variables
-$pagination-padding-y:              null !default;
-$pagination-padding-x:              null !default;
+$pagination-padding-y:              .625rem !default; // Boosted mod: instead of `null`
```
louismaximepiton, line 1580:
```suggestion
$pagination-padding-y:              .5625rem !default; // Boosted mod: instead of `null`
```

scss/_pagination.scss
```diff
@@ -154,10 +156,6 @@
   &:first-child .page-link {
```
louismaximepiton, line 156: Should we limit the width to `40px or 2.5rem` of `.page-item:not(:first-child):not(:last-child)` ? Big numbers seems weird atm.

site/content/docs/5.3/components/pagination.md
```diff
@@ -30,8 +30,8 @@ Looking to use an icon or symbol in place of text for some pagination links? Be
 {{< example >}}
 <nav aria-label="Page navigation example with icons">
   <ul class="pagination">
-    <li class="page-item"><a class="page-link" href="#" aria-label="Previous"></a></li>
-    <li class="page-item"><a class="page-link" href="#" aria-label="Next"></a></li>
+    <li class="page-item"><a class="page-link" href="#"><span class="visually-hidden">Previous</span></a></li>
+    <li class="page-item"><a class="page-link" href="#"><span class="visually-hidden">Next</span></a></li>
```
louismaximepiton, line 34: Change all occurence in the doc as well!
Do we add a migration note about this ?

louismaximepiton (changes requested):
Some few things to tackle before seing it live

site/content/docs/5.3/components/pagination.md
MewenLeHo: Note sure for the migration note since this PR is no more than an 'simple' a11y fix for https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/1931 🤔

scss/_pagination.scss
MewenLeHo: No, it has been deprecated.

scss/_variables.scss
```diff
@@ -1650,6 +1654,8 @@ $pagination-dark-active-item-border-color:  $brand-orange !default;
 $pagination-dark-disabled-bg:               $black !default;
 $pagination-dark-disabled-color:            $gray-700 !default;
 $pagination-dark-disabled-border-color:     $gray-700 !default;
+
+$pagination-item-max-width:                 2.5rem !default;
```
MewenLeHo, line 1658: Needed or not needed? Or needed only on mobile? It may depend of what you have done for the responsive pagination.

louismaximepiton: I think we should take care about this kind of thing once one PR is merged. So let's keep it like that

Two things about this: Maybe group with others before dark variant variables + replace value by `$spacer * 2` could be easier to understand (if not you can resolve this) ?

scss/_variables.scss
```diff
@@ -1614,16 +1614,20 @@ $pagination-transition:             $transition-focus !default;
...
 $pagination-icon-height:            subtract(1rem, 1px) !default;
+$pagination-gap:                    10px !default;
 
 $pagination-active-item-bg:         $primary !default;
 $pagination-active-item-color:      color-contrast($pagination-active-item-bg) !default;
 $pagination-active-item-border:     $pagination-active-item-bg !default;
 // End mod
+// fusv-disable
+$pagination-padding-end:            1.125rem !default; // Deprecated in Boosted 5.3.0 for responsive pagination
+// fusv-enable
```
louismaximepiton, line 1630: I think it's still a Boosted mod here.

scss/_variables.scss
```diff
@@ -1614,16 +1614,20 @@ $pagination-transition:             $transition-focus !default;
 // Boosted mod: no $pagination-border-radius-lg
 
 // Boosted mod
-$pagination-padding-end:            1.125rem !default;
 $pagination-icon:                   var(--#{$prefix}chevron-icon) !default;
-$pagination-icon-size:              subtract($spacer * 2, calc(var(--#{$prefix}border-width) * 2)) !default; // stylelint-disable-line function-disallowed-list
+$pagination-icon-min-width:         .625rem !default;
+$pagination-icon-size:              1.125rem !default;
```
louismaximepiton, line 1619: These two should be named almost the same since their goal is the same. Maybe something like `$pagination-icon-efficient-width/height`? It add some changes to variables but it feel more handy to me. What would think about a change like this ?

scss/_variables.scss
```diff
@@ -1614,16 +1614,20 @@ $pagination-transition:             $transition-focus !default;
...
 $pagination-icon-width:             add(.5rem, 1px) !default;
 $pagination-icon-height:            subtract(1rem, 1px) !default;
+$pagination-gap:                    10px !default;
```
louismaximepiton, line 1622: Do you think any other value would make sense here ? like `13px or 15px`

scss/_pagination.scss
```diff
@@ -182,6 +178,10 @@
     margin-left: var(--#{$prefix}pagination-margin-x-first-last);
   }
 
+  &:not(:first-child):not(:last-child) {
+    max-width: var(--#{$prefix}pagination-item-max-width);
```
louismaximepiton, line 182: Since all the styles seems to be applied on `.page-link`, I think we should do so for this rule, meaning change `width` or `max-width`. (`$pagination-item-max-width: 10px` or the width of the icon).

louismaximepiton (commented):
Few comments more, but the rendering looks great! Great work already

site/content/docs/5.3/components/pagination.md
louismaximepiton: Maybe just a quick note to remind people that they should change for better a11y ?

scss/_variables.scss
MewenLeHo: Something like `$pagination-icon-hitbox-width/height`? Or `$pagination-icon-box-width/height` maybe?
Not sure which one sounds more like a Bootstrap's name?

louismaximepiton: I don't have any idea on how to call those. Let's wait for the last review to decide ?

scss/_variables.scss
```diff
@@ -1614,15 +1614,21 @@ $pagination-transition:             $transition-focus !default;
...
 
 $pagination-active-item-bg:         $primary !default;
 $pagination-active-item-color:      color-contrast($pagination-active-item-bg) !default;
 $pagination-active-item-border:     $pagination-active-item-bg !default;
+
+$pagination-item-max-width:         $spacer * 2 !default;
```
louismaximepiton, line 1628:
```suggestion
$pagination-item-max-width:         $spacer * .5 !default;
```
Big numbers look weird atm

louismaximepiton (commented):
One small thing to tackle and let's send it to reviews.

scss/_variables.scss
MewenLeHo: Yes, it will be a job for the led dev review 😉 

**Simple reminder:** changing the name of `$pagination-icon-size` will need to be properly documented in our migration guide since it's an already existing variable.

louismaximepiton (approved):
🚀

Aniort:
ok, fine

Aniort:
ok, with the "title", i think, it's usable :)

Franco-Riccitelli:
The pagination buttons with the icon only look good.
The space between the buttons is ok for both variants.
For the pagination buttons with the icons and labels, please follow the spacing guide attached. Thanks
[Next_Previous_Spacing.zip](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/files/12417874/Next_Previous_Spacing.zip)

scss/_pagination.scss
```diff
@@ -1,10 +1,12 @@
 .pagination {
   // scss-docs-start pagination-css-vars
   --#{$prefix}pagination-padding-x: #{$pagination-padding-x};
+  --#{$prefix}pagination-padding-inline-x: #{$pagination-padding-inline-x};
```
louismaximepiton, line 4:
```suggestion
  --#{$prefix}pagination-label-padding-inline-x: #{$pagination-label-padding-inline-x}; // Boosted mod
```
I'd prefer to have an explicit name for this variable. Not this one specifically.

scss/_pagination.scss
```diff
@@ -149,15 +154,15 @@
       background-color: var(--#{$prefix}pagination-active-item-bg);
       border-color: var(--#{$prefix}pagination-active-item-border-color);
     }
+
+    &:not([title]) {
```
louismaximepiton, line 158: The whole selector might lead to an issue (if the `.page-item` is `.active` as well, it may not happen but if we can anticipate). IMO, the final selector should be `.page-item:first-child .page-link:not([title]), .page-item:last-child .page-link:not([title])`. 
However, I'm fine forcing a11y on this component so the `:not([title])` seems fair to me.

scss/_pagination.scss
```diff
@@ -149,15 +154,15 @@
...
+      padding-inline: var(--#{$prefix}pagination-padding-inline-x);
```
louismaximepiton, line 159: Might be a bit too soon for this https://caniuse.com/mdn-css_properties_padding-inline (but we use gap sometimes so idk). I think we should stick to `--bs-pagination-padding-x` here

scss/_pagination.scss
```diff
@@ -1,10 +1,12 @@
 .pagination {
   // scss-docs-start pagination-css-vars
   --#{$prefix}pagination-padding-x: #{$pagination-padding-x};
+  --#{$prefix}pagination-label-padding-inline-x: #{$pagination-label-padding-inline-x};
```
louismaximepiton, line 4:
```suggestion
  --#{$prefix}pagination-label-padding-x: #{$pagination-label-padding-x}; // Boosted mod
```

Just wondering if we should keep the `inline` or not.

MewenLeHo: I would say no, especially since we removed the `padding-inline` property.

scss/_pagination.scss
```diff
@@ -1,10 +1,12 @@
 .pagination {
   // scss-docs-start pagination-css-vars
   --#{$prefix}pagination-padding-x: #{$pagination-padding-x};
+  --#{$prefix}pagination-label-padding-x: #{$pagination-label-padding-x};
```
louismaximepiton, line 4:
```suggestion
  --#{$prefix}pagination-label-padding-x: #{$pagination-label-padding-x}; // Boosted mod
```

Franco-Riccitelli:
All looks good to me.

scss/_variables.scss
```diff
@@ -1614,15 +1615,21 @@ $pagination-transition:             $transition-focus !default;
...
 $pagination-icon-height:            subtract(1rem, 1px) !default;
+$pagination-gap:                    5px !default;
 
 $pagination-active-item-bg:         $primary !default;
 $pagination-active-item-color:      color-contrast($pagination-active-item-bg) !default;
 $pagination-active-item-border:     $pagination-active-item-bg !default;
+
+$pagination-item-max-width:         $spacer * .5 !default;
+// fusv-disable
+$pagination-padding-end:            1.125rem !default; // Deprecated in Boosted 5.3.0 for responsive pagination
```
louismaximepiton, line 1631:
```suggestion
$pagination-padding-end:            1.125rem !default; // Deprecated in Boosted 5.3.2 for responsive pagination
```

louismaximepiton (approved):
👌

PR #2126, opened 2023-07-04
fix(docs): update broken DSM links

julien-deramond:
### Related issues
 
 This PR fixes some broken DSM links.
 
 ### Live previews
 
 - <https://deploy-preview-2126--boosted.netlify.app/>


MewenLeHo (approved):
Good catch 👍 

I think it's coming from the design description we made for the design team: `progress indicators` were then splitted into `progress bars` and `spinners` in the DSM, exactly like that we have in our doc.

So we may have broken links in the incoming callouts for the `spinners` part too.

PR #2206, opened 2023-08-29
Workflow: fix broken review in progress workflow

louismaximepiton:

### Related issues

NA

### Description

Changes to make the workflow apply the condition.

### Motivation & Context

The workflow triggers when it shouldn't.

### Types of change


julien-deramond (approved):
LGTM! Let's try this!

PR #2237, opened 2023-09-12
Add Solaris icons for light/dark mode

julien-deramond:
### Description

This PR changes the light/dark icons used in the theme selector to use Solaris icons:
- UI Light Mode
- UI Dark Mode

UI Auto Mode Solaris icon is not yet available and will be updated later on.

The rendering is the following:

Screenshot 2023-09-14 at 12 21 11

### Technical

_Note: icon sizes have been specified by @franco-riccitelli_

- Icons (sun, moon, auto, and tick) size: 20px x 20px (in the toggle button and in the dropdown menu items)
  - @franco-riccitelli This is treated as a specific rendering within the documentation and so not as an official variant of the dropdowns containing icons. It could maybe be a new variant within the DSM and Boosted
- Caret size: 10px x 5px
  - This size is the same as in our current dropdowns so I suggest creating a specific issue to tackle it within the whole framework (the current size is a bit too large)
- This PR changes the `#check2` SVG used in the examples page as it wasn't the right one before (it was Bootstrap's)
- This PR includes the fix from https://github.com/twbs/bootstrap/pull/39177 right now so that we can see the tick in example pages
- The rendering in light/dark mode is a bit broken on the examples page but this will be treated outside of this PR because probably linked to the basic rendering of other components in Boosted.

### Live previews

- https://deploy-preview-2237--boosted.netlify.app/
- https://deploy-preview-2237--boosted.netlify.app/docs/5.3/examples/stickers/ (for example) for the other theme selector which is treated separately in our code


MewenLeHo (approved):
New icons look indeed smaller but are still ok from an a11y point of view so ok for me.

julien-deramond:
* Icons are prolly too small. Let's try to force their size + the height of the button.
* Change the tick icon as well to use the Solaris one

julien-deramond:
Design workshop with @franco-riccitelli:
- Icons when the selector is closed: 20px x 20px
- Icons in theme selector dropdown items: 20px x 20px (even the tick icon)
- Double check the caret size: it should be 10x x 5px
- Force the button height to 40px

MewenLeHo (approved):
Looks good to me and a11y is ok too.

PR #2276, opened 2023-10-03
Dark mode: Local navigation

hannahiss:
### Description

Local navigation in dark mode, by using existing css vars:
- `$local-nav-hover-bg` = --bs-secondary-bg // from - `gray300 to gray900 under lg point
- `$local-nav-active-color` = --bs-link-hover-color // from accessible-orange to brand-orange
- `$local-nav-border-color` = --bs-border-color-translucent // from gray-500 to gray-700

⚠️ 3 additional vars are added:
- `--bs-hover-color` to switch `$local-nav-hover-color` from black to brand-orange 
- `--bs-active-bg` to switch `$local-nav-active-bg` from gray-400 to gray-700
- `--bs-disabled-color` to switch `$form-star-rating-disabled-color`, `$nav-link-disabled-color` and `$navbar-light-disabled-color` from gray-500 to gray-700 => ⚠️ **used in local-nav, navbars, navs and star-rating**

### Links

- https://deploy-preview-2276--boosted.netlify.app/docs/5.3/components/local-navigation/
- https://deploy-preview-2276--boosted.netlify.app/docs/5.3/dark-mode/#stepped-process

PR #2278, opened 2023-10-03
Dark mode: form color

louismaximepiton:
### Description

Form color in dark mode, by using existing and new Sass vars :
| Sass var | Previous value | New value |
| :------- | :--------------: | :----------: |
| `$form-color-border-color` | - | `var(--#{$prefix}emphasis-color)` |
| `$form-color-hover-bg-color` | - | `var(--#{$prefix}emphasis-color)` |
| `$form-color-disabled-bg-color` | - | `$input-bg` |
| `$form-color-disabled-border-color` | `$gray-500` | `var(--#{$prefix}disabled-color)` |
| `$form-color-disabled-background-swatch` | `brightness(0) invert(1) brightness(.8)` | `var(--#{$prefix}disabled-filter)` |

:warning: In the same line as #2276, introduced a new CSS var: `--bs-disabled-filter` (associated to `$disabled-filter` and `$disabled-filter-dark`. It renders the same value as `--bs-disabled-color`). Use it on icons or whatever to have the same rendering.

### Links

- https://deploy-preview-2278--boosted.netlify.app/docs/5.3/forms/form-control#color
- https://deploy-preview-2278--boosted.netlify.app/docs/5.3/dark-mode/#color


hannahiss (approved):
Juste 2 little observation:
- Just to be sure, I supposed it is due to browser, but in Chrome we have a border #777 around the purple, and not in FF
- In Chrome, when the page is globally in light mode, when opening the picker, it has contextually a dark or light bg, depending on the component's theme. But when the page is globally in dark mode, picker always sticks to dark background. (supposed its due to the browser's implementation :-) )

Otherwise it looks good! If you approve both observations, I think you can merge!

louismaximepiton:
> * Just to be sure, I supposed it is due to browser, but in Chrome we have a border `#777` around the purple, and not in FF

Good catch! it should have come from Bootstrap merge commits we must have missed it. I've added it with https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2278/commits/b19be8cdc64221b2d8b57102db28f7e75b7b0b79.

> * In Chrome, when the page is globally in light mode, when opening the picker, it has contextually a dark or light bg, depending on the component's theme. But when the page is globally in dark mode, picker always sticks to dark background. (supposed its due to the browser's implementation :-) )

I wasn't aware of such a behavior. I think this should be tackled Bootstrap side with `color-scheme: light;` inside the `:root, [data-bs-theme=light]` selector. Any opinion on this ?

julien-deramond (approved):

PR #2290, opened 2023-10-09
Dark mode: Spinners

julien-deramond:
New Sass vars:
- `$spinner-color`

⚠️ Added temporary `data-bs-theme` for buttons containing spinners. This should go if we modify `color:` automatically with `[data-bs-theme]`.

### Live previews
- https://deploy-preview-2290--boosted.netlify.app/docs/5.3/components/spinners/
- https://deploy-preview-2290--boosted.netlify.app/docs/5.3/dark-mode/#spinners

### DIFF
--- a/scss/_spinners.scss
+++ b/scss/_spinners.scss
@@ -7,6 +7,7 @@
   display: inline-block;
   width: var(--#{$prefix}spinner-width);
   height: var(--#{$prefix}spinner-height);
+  color: #{$spinner-color}; // Boosted mod
   vertical-align: var(--#{$prefix}spinner-vertical-align);
   // stylelint-disable-next-line property-disallowed-list
   border-radius: 50%;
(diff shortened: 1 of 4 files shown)


louismaximepiton (commented):
Just wondering if we should recheck this once we merge text utilities since the orange used here is `#f16e00` everywhere.
There's an issue [here](https://deploy-preview-2290--boosted.netlify.app/docs/5.3/components/spinners/#buttons), do you think we should care about this right know ?

julien-deramond:
I can add a sub-task to re-check that for text utilities and buttons yes. Right now, there's nothing else I can do without breaking other stuff.

louismaximepiton (approved):
I'm fine with the changes, the actual solution for spinners inside buttons might only work for light theme documentation once buttons will be handled but let's try it this way and document to recheck once finished.

PR #2293, opened 2023-10-09
Dark mode: Buttons

louismaximepiton:
### Description

The PR might be working, the description is WIP

### Links

- https://deploy-preview-2293--boosted.netlify.app/docs/5.3/components/buttons
- https://deploy-preview-2293--boosted.netlify.app/docs/5.3/dark-mode/#buttons

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -16,6 +16,7 @@
   --#{$prefix}btn-border-width: #{$btn-border-width};
   --#{$prefix}btn-border-color: transparent;
   --#{$prefix}btn-border-radius: #{$btn-border-radius};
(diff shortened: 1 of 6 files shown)


julien-deramond (approved):
Neat! Let's merge this first version so that it unblocks other comments.

I'll the following comments within the description of the main PR:
- `.btn-primary` is now transparent when active; can be an issue against images
- `.btn-info` is now transparent when hovered; can be an issue against images
- Mention the impact of the new `button-variant` mixin in the migration guide
- Not a big fan of using `$red` instead `$danger` because the color doesn't change. Do we need to introduce new Sass variables for fixed functional colors?
- Do we need to drop `btn-outline-secondary`?
- Double check later the `&:not(.dropdown-toggle-split) {` use case in `_dropdown.scss`

PR #2300, opened 2023-10-10
feat(buttons): reintroduce colored versions from Bootstrap of buttons and outline buttons

julien-deramond:
### Description

This PR reintroduces the colored versions of buttons and outlined buttons coming from Bootstrap instead of hiding them or overwriting their value.

First of all, it adds a little bit of extra CSS in the bundle due to the new outline classes. However, it's necessary for compatibility with Bootstrap when folks use external plugins, libraries, etc. based on Bootstrap with Boosted style.

For regular buttons, the changes are minimal since we modify only the rendering of `.btn-warning` and `.btn-info` so that they use ODS functional colors. They are not compatible with ODS design guidelines rules so a design callout has been added to mention that to the users.
The migration guide includes something about it since the rendering of these buttons will change whenever folks have wrongly used the concepts with `.btn-warning` and `.btn-info`; this can happen if they have used them only based on the rendering and not the meaning.

Please note that I tried to get back to the code from Bootstrap to create the "regular" buttons, but the rules are too precise and not generic enough to be able to do that (event if some success, danger, info, warning buttons rely on the same parameters).

For the outlined buttons, we already had only the secondary one.
Technically, I've used the same technique as Bootstrap by looping over the `$theme-colors` and calling the newly created `button-outline-variant` mixin (with extra parameters to differentiate the active from the hover state).
A special rendering has been introduced for outline secondary and dark buttons when rendered as dark variants.
The new variants and the new mixin are documented in the migration guide, and a design callout has been added to mention that only the secondary variant can be used on their website.

- [ ] ⚠️ **Before merging!** Drop the dark variants examples used for the review in Buttons and Outlined Buttons

### Types of change

### Live previews

- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/components/buttons/#variants
- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/components/buttons/#outline-buttons
- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/components/buttons/#sass-mixins
- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/migration

Some impacts where I changed the classes used or not depending on the use cases:
- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/components/button-group/#mixed-styles
- https://deploy-preview-2300--boosted.netlify.app/docs/5.3/components/button-group/#button-toolbar
- https://deploy-preview-2300--boosted.netlify.app (in private mode for the cookies management notif)

### DIFF
--- a/site/content/docs/5.3/migration.md
+++ b/site/content/docs/5.3/migration.md
@@ -9,6 +9,21 @@ aliases:
 toc: true
 ---
 
+## v5.3.3
+
+<hr class="mb-4">
+
+Boosted v5.3.3 has landed also including specific Boosted content as usual.
+
+If you need more details about the changes, please refer to the [v5.3.3 release](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/releases/tag/v5.3.3).
+
+### Components
+
+- **Buttons**
+  - <span class="badge bg-warning">Warning</span> Warning and info buttons are now rendered in yellow and blue colors. Please check that it doesn't break your design.
+  - All versions of outline buttons are now displayed in the documentation for Bootstrap compatibility reasons even if you can only use the secondary variant.
+    - A new ` button-outline-variant` Sass mixin is now available.
+
 ## v5.3.2
 
 <hr class="mb-4">
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -26,11 +26,11 @@
     },
     {
       "path": "./dist/css/boosted.css",
-      "maxSize": "43.75 kB"
+      "maxSize": "44.0 kB"
     },
     {
       "path": "./dist/css/boosted.min.css",
-      "maxSize": "40.75 kB"
+      "maxSize": "41.0 kB"
     },
     {
       "path": "./dist/js/boosted.bundle.js",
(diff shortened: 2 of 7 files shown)


site/content/docs/5.3/migration.md
```diff
@@ -9,6 +9,21 @@ aliases:
...
+Boosted v5.3.3 has landed also including specific Boosted content as usual.
+
+If you need more details about the changes, please refer to the [v5.3.3 release](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/releases/tag/v5.3.3).
+
+### Components
+
+- **Buttons**
+  - <span class="badge bg-warning">Warning</span> Warning and info buttons are now rendered in yellow and blue colors. Please check that it doesn't break your design.
+  - All versions of outline buttons are now displayed in the documentation for Bootstrap compatibility reasons even if you can only use the secondary variant.
+    - A new ` button-outline-variant` Sass mixin is now available.
```
hannahiss, line 25:
```suggestion
    - A new `button-outline-variant` Sass mixin is now available.
```

julien-deramond: This one is closed

PR #2223, opened 2023-09-11
Dark mode

julien-deramond:
## Description

_Grab a cup of coffee or tea, and take your time._

This PR finally introduces the dark mode in Boosted. It's a huge PR that has been developed over the course of 3 months. It's the result of a lot of prototyping, discussions, and iterations with the ODS web core team, but also with the Bootstrap team.

The entire solution relies on the `data-bs-theme` attribute that has been introduced in Bootstrap 5.3.0. It allows to switch between light and dark mode, and to switch between different themes (light, dark, etc.) in the same page.

### Live previews

- https://deploy-preview-2223--boosted.netlify.app
- In order to be able to check the different renderings in light and dark mode, contextual or not, this PR adds an hidden [`/dark-mode` page to the documentation](https://deploy-preview-2223--boosted.netlify.app/docs/5.3/dark-mode).

### Usage

The exact usage of the dark mode for Orange websites and web apps is not defined precisely yet. It's a work in progress. The goal of this PR is to provide the technical solution to implement the dark mode in Boosted and to provide a first version of the documentation to help developers to use it. Here are the remaining tasks to do to define the usage of the dark mode at Orange:

- [ ] Define in the DSM what is the dark mode and why it's important for Orange.
- [ ] Define the usage of the dark mode at Orange. Must it be present on all websites, web apps, etc.? If not, in which cases should it be used?
  - [ ] Update the DSM to explain this rule
  - [ ] When the rule is defined, update the Boosted documentation to explain this rule, or at least link to the DSM

### Switching from light to dark mode

The switch from light to dark mode is not defined precisely yet. It's a work in progress. The goal of this PR is to provide the technical solution to implement the dark mode in Boosted and to provide a first version of the documentation to help developers to use it. Here are the remaining tasks to do to define the switch from light to dark mode at Orange:

- [ ] Define how to switch from light to dark mode in web interfaces. Is it always the same selector? Always in the header? Always a dropdown? Always with the light/dark/auto options?
  - [ ] Update the DSM to define this rule
  - [ ] When the rule is defined, update the Boosted documentation to explain this rule, or at least link to the DSM
  - [ ] Do we insert this switch to all Boosted/UI Kit/DSM navbars and examples?
  - [ ] Do we create a reusable component for this switch?
  - [ ] Do we need to update [the new Navbar mode selector example page we created](https://deploy-preview-2223--boosted.netlify.app/docs/5.3/examples/navbar-mode-selector/)?

[...]

PR #2306, opened 2023-10-12
fix(unit tests): drop `.btn-secondary` from carousel play/pause buttons

julien-deramond:
### Description

This PR suggests changing the markup used to render the play/pause buttons in carousels' unit tests so that they use the same one as in https://boosted.orange.com/docs/5.3/components/carousel/#with-pauseplay-button:

```html
<button type="button" class="btn btn-icon carousel-control-play-pause pause" data-bs-target="#carouselExamplePause" data-bs-play-text="Play Carousel" data-bs-pause-text="Pause Carousel" title="Pause Carousel">
  <span class="visually-hidden">Pause Carousel</span>
</button>
```

It basically means to drop the use of `.btn-secondary`.

_Note: it'll help with the work in progress in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2223_

### Motivation & Context

Consistency between the carousels' markups.

### Types of change

### Live previews

N/A

### DIFF
--- a/js/tests/unit/carousel.spec.js
+++ b/js/tests/unit/carousel.spec.js
@@ -540,7 +540,7 @@ describe('Carousel', () => {
     it('should take care of element either passed as a CSS selector or DOM element (Play/Pause button)', () => {
       fixtureEl.innerHTML = [
         '<div id="myCarousel" class="carousel"></div>',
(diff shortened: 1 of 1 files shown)


louismaximepiton (approved):
It's not exactly the carousel's actual markup since the pause button is contained inside the carousel but I'm fine with the changes since it passes tests

PR #2309, opened 2023-10-12
Dark mode: Toggle buttons

louismaximepiton:
### Description Superseeded by #2365.

⚠️ Big refactoring of this component. Might be done on main before dark mode so I can split this PR to have a smaller diff?

⚠️ Impacts on user for toggle buttons: they don't need a `.btn-*` class anymore on the toggles.

⚠️ Impact on default `.btn` rendering (for example: `View on GitHub` button).

### Links

- https://deploy-preview-2309--boosted.netlify.app/docs/5.3/forms/checks-radios/#toggle-buttons
- https://deploy-preview-2309--boosted.netlify.app/docs/5.3/dark-mode/#toggle-buttons
- https://deploy-preview-2309--boosted.netlify.app/docs/5.3/components/button-group/#checkbox-and-radio-button-groups


scss/forms/_form-check.scss
```diff
@@ -246,144 +246,15 @@
...
-  + .btn.active {
-    color: color-contrast($primary);
-    background-color: $primary;
-    // Remove CSS gradients if they're enabled
-    background-image: if($enable-gradients, none, null);
-    border-color: $primary;
-  }
-  // End mod
-
-  &[disabled],
```
julien-deramond, line 329: It's not possible to keep this rule as Bootstrap so that we can keep also `$form-check-btn-check-disabled-opacity`?

scss/_buttons.scss
```diff
@@ -72,6 +86,7 @@
     // Boosted mod: no &:focus-visible
   }
 
+  .btn-check:disabled + &,
```
julien-deramond, line 89: Please double-check the Boosted mods. Seems to me that some are missing in this file.

julien-deramond:
There seems to be an issue with the hover of the first one that should have a border:

julien-deramond:
There's an issue with the active state of link buttons:

Screenshot 2023-11-09 at 07 06 01

PR #2313, opened 2023-10-13
Auto color and bg-color with data-bs-theme

julien-deramond:
### Description

This PR tries a new approach with an automatic set of `color:` and `background-color:` linked to `[data-bs-theme]` attribute.

So when your create a `<div data-bs-theme="dark">` the content will be automatically rendered with a dark background and a white text in Boosted.

Based on this modification, this PR also sets the Sass and/or CSS variables of components handling color and/or bg color to `null` by default, or `inherit`, and also sometimes sets the background to transparent.

#### More info 

We decided to make big components (components with multiple layers) to inherit the color because it seemed more logical. Some components escape this logic such as spinners. It's because it's not really the color that is redefined but the use of `currentColor`. 
All this is due to a technical issue we have when the theme is set directly on the component.

If the component sets `color` at the top level, it overrides the `color` coming from the `data-bs-theme` so if it's set to `null` or `inherit`, it will inherit the color above and avoid `data-bs-theme` `color`. On multiple-layer components it's easier to handle since we only need to avoid declaring `color` at the top level of this component.

Weird behavior for titles. For example `h1` -> Redefines color but `.h1` -> inherits when the data theme is placed on them directly.

For basic elements, form elements and input labels, we decided to keep the behavior as-is. Most likely the theme won't be applied to "basic" elements themselves. And the users could override this behavior if they are using this edge case.

#### Note

- We got rid of `text-body` classes in dark mode page for basic elements introduced by https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2355 to really test the impact of this PR in this context.

<details>
<summary>We chose not to modify <code>color-mode()</code> mixin, here's why...</summary>

### Not covered in this PR

Based on this modification, it would be possible to do the following in components:

```scss
@include color-mode(light, true) {
  --#{$prefix}accordion-btn-icon: #{escape-svg($accordion-button-icon)};
  --#{$prefix}accordion-btn-active-icon: #{escape-svg($accordion-button-active-icon)};
}

@if $enable-dark-mode {
  @include color-mode(dark, true) {
    --#{$prefix}accordion-btn-icon: #{escape-svg($accordion-button-icon-dark)};
    --#{$prefix}accordion-btn-active-icon: #{escape-svg($accordion-button-active-icon-dark)};
  }
}
```

instead of the current in Bootstrap which doesn't work when a light accordion is declared within a dark container: 

```scss
@if $enable-dark-mode {
  @include color-mode(dark) {
    .accordion-button::after {
      --#{$prefix}accordion-btn-icon: #{escape-svg($accordion-button-icon-dark)};
      --#{$prefix}accordion-btn-active-icon: #{escape-svg($accordion-button-active-icon-dark)};
    }
  }
}
```
</details>

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1146,7 +1146,7 @@ $input-focus-width:                     $input-btn-focus-width !default;
 $input-focus-box-shadow:                none !default; // Boosted mod
 
 $input-placeholder-color:               var(--#{$prefix}placeholder-color) !default; // Boosted mod: instead of `var(--#{$prefix}secondary-color)`
-$input-plaintext-color:                 null !default;
+$input-plaintext-color:                 null !default; // Boosted mod: instead of `var(--#{$prefix}body-color)`
 
 // Boosted mod: no $input-height-border
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -799,7 +799,7 @@ $headings-font-family:        null !default;
 $headings-font-style:         null !default;
 $headings-font-weight:        700 !default;
 $headings-line-height:        $h6-line-height !default;
-$headings-color:              $black !default; // Boosted mod: instead of `inherit`
+$headings-color:              inherit !default;
 // scss-docs-end headings-variables
 
 // scss-docs-start display-headings
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1436,7 +1436,7 @@ $nav-link-padding-y:                $spacer * .5 !default;
 $nav-link-padding-x:                $spacer !default;
 $nav-link-font-size:                null !default;
 $nav-link-font-weight:              $font-weight-bold !default;
-$nav-link-color:                    var(--#{$prefix}link-color) !default;
+$nav-link-color:                    inherit !default; // Boosted mod: instead of `var(--#{$prefix}link-color)`
 $nav-link-hover-color:              var(--#{$prefix}link-hover-color) !default;
 $nav-link-transition:               null !default; // Boosted mod
 $nav-link-disabled-color:           var(--#{$prefix}disabled-color) !default; // Boosted mod: instead of `var(--#{$prefix}secondary-color)`
@@ -1640,7 +1640,7 @@ $pagination-padding-x:              null !default; // Boosted mod: instead of `.
 
 $pagination-font-size:              $font-size-base !default;
 
-$pagination-color:                  var(--#{$prefix}link-color) !default;
+$pagination-color:                  inherit !default; // Boosted mod: instead of `var(--#{$prefix}link-color)`
 $pagination-bg:                     transparent !default; // Boosted mod: instead of `var(--#{$prefix}body-bg)`
 $pagination-border-radius:          var(--#{$prefix}border-radius) !default;
 $pagination-border-width:           var(--#{$prefix}border-width) !default;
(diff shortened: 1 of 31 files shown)


scss/mixins/_color-mode.scss
```diff
@@ -13,9 +13,18 @@
       }
     }
   } @else {
-    [data-bs-theme="#{$mode}"] {
-      @content;
+    // Boosted mod
+    @if $mode == "light" {
```
louismaximepiton, line 17: Should we introduce the concept of "default theme" that could be useful for users?
Or maybe use the parameter `$root` ? Because it works for us but maybe not for every use case

louismaximepiton (commented):
I'm strongly in favor of this change but here are some questions that have to be raised: 
- [ ] What if people use it already, does it have impacts ?
- [ ] We might need some explanations on how to use it in `_color-modes.md` (i.e. no selector inside or might be breaking + explain why, ...)
- [ ] Seems to work not that good for many of our transparent components atm

scss/mixins/_color-mode.scss
```diff
@@ -13,9 +13,18 @@
       }
     }
   } @else {
-    [data-bs-theme="#{$mode}"] {
-      @content;
+    // Boosted mod
```
julien-deramond, line 16: Check if it is really useful for this PR. If not yet, let's drop it and wait for https://github.com/twbs/bootstrap/pull/39295

julien-deramond:
There's an issue in the /dark-mode page for the badge component where there's a background color for "Example heading" and the shopping basket icon (see https://deploy-preview-2313--boosted.netlify.app/docs/5.3/dark-mode/#badge):

Screenshot 2023-11-29 at 07 17 20

`data-bs-theme` should probably be on the component directly and not on the `<h6>` or on the `<a>` directly.

julien-deramond:
There's an issue with the "Dark/Light theme" on component for modals in https://deploy-preview-2313--boosted.netlify.app/docs/5.3/dark-mode/#modal where the background is solid:

Screenshot 2023-11-29 at 07 25 20

julien-deramond:
I think the behavior shown in https://deploy-preview-2313--boosted.netlify.app/docs/5.3/dark-mode/#contents is acceptable even some use cases are weird such as the highlight. But there's also no real usage for that, IMO it doesn't make any sense to do that IRL:

Screenshot 2023-11-29 at 07 32 39

julien-deramond:
In https://deploy-preview-2313--boosted.netlify.app/docs/5.3/dark-mode/#helper, the label shouldn't have a `data-bs-theme`

scss/_variables.scss
```diff
@@ -1146,7 +1146,7 @@ $input-focus-width:                     $input-btn-focus-width !default;
 $input-focus-box-shadow:                none !default; // Boosted mod
 
 $input-placeholder-color:               var(--#{$prefix}placeholder-color) !default; // Boosted mod: instead of `var(--#{$prefix}secondary-color)`
-$input-plaintext-color:                 var(--#{$prefix}body-color) !default;
+$input-plaintext-color:                 null !default;
```
julien-deramond, line 1149:
```suggestion
$input-plaintext-color:                 null !default; // Boosted mod: instead of `var(--#{$prefix}body-color)`
```

julien-deramond:
I haven't checked it yet but since "Placeholders" in /dark-mode page PR has been merged, maybe there are some `.text-body` classes to remove too.

julien-deramond:
Some `.text-body`s could probably be removed:
- `<a href="{{.link}}" class="stretched-link text-decoration-none h4 text-body" aria-label="{{.description}}" title="{{.description}}">{{.name}}</a>` in `/site/content/docs/5.3/guidelines/_index.md`

PR #2325, opened 2023-10-18
Dark mode: Popovers

louismaximepiton:
### Description

Modals in dark mode, by using existing Sass vars :

| Sass var | Previous value | New value |
| :------- | :--------------: | :----------: |
| `$popover-header-color` | `$headings-color` | `var(--#{$prefix}heading-color)` |
| `$headings-color` | `inherit` | `$black` |

⚠️ Need a `data-bs-theme="body"` somehow ?
⚠️ Change in `$headings-color` and probably need to check everything that depends from that (`var(--#{$prefix}heading-color)`, etc...)

### Links

- https://deploy-preview-2325--boosted.netlify.app/docs/5.3/components/popovers
- https://deploy-preview-2325--boosted.netlify.app/docs/5.3/dark-mode/#popovers

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -796,7 +796,7 @@ $headings-font-family:        null !default;
 $headings-font-style:         null !default;
 $headings-font-weight:        700 !default;
 $headings-line-height:        $h6-line-height !default;
-$headings-color:              $black !default;
+$headings-color:              $black !default; // Boosted mod: instead of `inherit`
 // scss-docs-end headings-variables
 
 // scss-docs-start display-headings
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -1815,7 +1815,7 @@ $popover-box-shadow:                var(--#{$prefix}box-shadow) !default;
 $popover-header-font-size:          $font-size-lg !default; // Boosted mod: instead of `$font-size-base`
 $popover-header-line-height:        1.11 !default; // Boosted mod
 $popover-header-bg:                 $popover-bg !default; // Boosted mod: instead of `var(--#{$prefix}secondary-bg)`
-$popover-header-color:              $headings-color !default;
+$popover-header-color:              var(--#{$prefix}body-color) !default; // Boosted mod: instead of `$headings-color`
```
julien-deramond, line 1818: Are you sure we don't want to keep either `$headings-color` or either create a new reusable `--bs-headings-color` so that we don't lose that layer of semantics?

julien-deramond (approved):

PR #2340, opened 2023-10-26
feat(back to top): new `$back-to-top-icon-margin-left` Sass var

julien-deramond:
### Description

This PR introduces in new Sass variable for the back-to-top component: `$back-to-top-icon-margin-left`.

### Types of change

### Live previews


julien-deramond:
@louismaximepiton Are we still needing this variable? It seems to be used only in the old use case which is not anymore in the documentation:

```html
<nav aria-label="Label inside back to top example" class="back-to-top position-static">
  <a href="#top" class="back-to-top-link position-static btn btn-secondary px-3">Back to top</a>
</nav>
```

Maybe we can drop this variable and the corresponding rule

scss/_variables.scss
```diff
@@ -2275,6 +2275,9 @@ $back-to-top-title-offset-right: add(100%, var(--#{$prefix}border-width)) !defau
 $back-to-top-title-padding:      subtract($btn-padding-y, 1px) $btn-padding-x add($btn-padding-y, 1px) !default;
 $back-to-top-title-bg-color:     $white !default;
 $back-to-top-icon:               var(--#{$prefix}chevron-icon) !default;
+// fusv-disable
+$back-to-top-icon-margin-left:   $spacer * .5 !default; // Deprecated in v5.3.3
+// fusv-enable
```
louismaximepiton, line 2280: We don't need to introduce it 😄

louismaximepiton (approved):

julien-deramond:
Note: remove "Even if the rendering could still work, it is recommended to only use the versions presented in the documentation." from the migration guide.

PR #2341, opened 2023-10-26
fix(form-control-color): adjust size when valid feedback

julien-deramond:
### Description

Spotted in https://deploy-preview-2302--boosted.netlify.app/docs/5.3/dark-mode/#form-validation while working on the dark mode, the rendering of the form control color when the feedback is valid is odd:

Screenshot 2023-10-26 at 08 18 12

The purple area should be square like for the other states.

This PR fixes it to have this unified rendering whatever the states:

Screenshot 2023-10-26 at 12 11 56

While we were working on this issue, we spotted another issue that is linked to the right-padding of the valid icon which was too important. With this PR, this padding has the same value (for all form controls) as the top/bottom padding.
Only the left padding of the input is bigger, but this is probably by design.

Some issues have been created based on the development of this one:
- https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/2342

### Types of change

### Live previews

- https://deploy-preview-2341--boosted.netlify.app/docs/5.3/forms/validation/

### DIFF
--- a/scss/mixins/_forms.scss
+++ b/scss/mixins/_forms.scss
@@ -30,7 +30,7 @@
           padding-right: $input-height-inner;
           background-image: escape-svg($icon);
           background-repeat: no-repeat;
-          background-position: right subtract($input-height-inner-quarter, 2px) top subtract($input-height-inner-quarter, 2px);
+          background-position: right subtract($input-height-inner-quarter, $input-border-width) top subtract($input-height-inner-quarter, $input-border-width);
           background-size: $input-height-inner-half $input-height-inner-half;
         }
       }
@@ -49,7 +49,7 @@
 
       .form-control-color {
         @include form-validation-state-selector($state) {
-          width: calc(#{$form-color-width} + #{$input-height-inner-half} + #{$input-height-inner-quarter} - 2px); // stylelint-disable-line function-disallowed-list
+          width: calc(#{$form-color-width} + #{$input-height-inner-half} + #{$input-height-inner-quarter} - #{$input-border-width}); // stylelint-disable-line function-disallowed-list
         }
       }
     }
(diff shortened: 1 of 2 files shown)


scss/mixins/_forms.scss
```diff
@@ -49,7 +49,7 @@
 
       .form-control-color {
         @include form-validation-state-selector($state) {
-          width: add($form-color-width, $input-height-inner);
+          width: calc($form-color-width + $form-feedback-icon-size + .8125rem); // stylelint-disable-line function-disallowed-list
```
julien-deramond, line 52: I haven't found any logical calculation to avoid having `.8125rem` and use a Sass variable instead. If you have any ideas, don't hesitate.

scss/mixins/_forms.scss
```diff
@@ -49,7 +49,7 @@
 
       .form-control-color {
         @include form-validation-state-selector($state) {
-          width: add($form-color-width, $input-height-inner);
+          width: calc(#{$form-color-width} + #{$form-feedback-icon-size} + .8125rem); // stylelint-disable-line function-disallowed-list
```
louismaximepiton, line 52: I think the icon isn't well centered. I found 2 solutions:
- First one uses `calc(#{$form-color-width} + #{$form-feedback-icon-size} + 1.0625rem)` and sets the padding-right to `1.25em + 1.25rem`
- The second one uses `calc(#{$form-color-width} + #{$form-feedback-icon-size} + .5rem)` and sets `background-position: right $input-height-inner-quarter top subtract($input-height-inner-quarter, 2px);`

I don't know which one's better here.

louismaximepiton (approved):

scss/mixins/_forms.scss
julien-deramond: We worked together for a better version. Still not perfect, but due to previous code.

PR #2352, opened 2023-11-03
Dark mode: Tooltips

louismaximepiton:
### Description

Tooltips in dark mode

### Links

- https://deploy-preview-2352--boosted.netlify.app/docs/5.3/components/tooltips
- https://deploy-preview-2352--boosted.netlify.app/docs/5.3/dark-mode/#tooltips


julien-deramond (approved):
Arrow border color rendering is not great in dark mode. The value is correct but it looks like it's grayish... IDK if we can find a workaround for that.

PR #2365, opened 2023-11-09
Dark mode: fix toggle-buttons

louismaximepiton:
- New `.btn-toggle` specific Boosted class

### Live preview

- https://deploy-preview-2365--boosted.netlify.app/docs/5.3/dark-mode/#toggle-buttons


julien-deramond:
Some elements have a black color when the dark theme is set to the container in light mode instead of white:

Screenshot 2023-11-17 at 14 44 04

scss/_variables.scss
```diff
@@ -1204,7 +1204,7 @@ $form-check-input-indeterminate-bg-image:       url("data:image/svg+xml,<svg xml
 
 $form-check-input-disabled-opacity:        null !default;
 $form-check-label-disabled-opacity:        null !default;
-$form-check-btn-check-disabled-opacity:    null !default;
+// Boosted mod: no $form-check-btn-check-disabled-opacity
```
julien-deramond, line 1207: Can't we keep this Sass var and apply it even if `null` just to ensure compatibility with Bootstrap like it's done for `$form-check-label-disabled-opacity` for instance?

scss/_buttons.scss
```diff
@@ -58,8 +58,7 @@
 
   // Boosted mod: no .btn-check:focus-visible + &
 
-  // Boosted mod: custom styles for .btn-check
-  // Boosted mod: simpler rule
+  // Boosted mod: selectors are slightlty different to handle `.btn-check:checked`
```
louismaximepiton, line 61:
```suggestion
  // Boosted mod: selectors are slightly different to handle `.btn-check:checked`
```

scss/_buttons.scss
```diff
@@ -73,7 +72,9 @@
     // Boosted mod: no box shadow modification for the active state
     // Boosted mod: no &:focus-visible
   }
+  // End mod
 
+  // Boosted mod: selectors are slightlty different to handle `.btn-check`
```
louismaximepiton, line 77:
```suggestion
  // Boosted mod: selectors are slightly different to handle `.btn-check`
```

PR #2367, opened 2023-11-10
Dark mode: Download app example

louismaximepiton:
### Description

⚠️ If `--bs-border-color-translucent` is mapped to `.text-tertiary`, `.bg-tertiary`, `.border-tertiary`, we could replace `background-color` CSS specific rule here of `.custom-vr`
⚠️ Might not need the `.text-body` once #2313 is merged.

### Links

- https://deploy-preview-2367--boosted.netlify.app/docs/5.3/examples/download-app


site/content/docs/5.3/examples/download-app/index.html
```diff
@@ -286,7 +286,7 @@ <h2 class="h1 mb-0 mb-md-5 text-primary text-center" id="download-notice">How to
                 <span class="d-inline-block fw-bold mb-3">Register with your Google store account to our <a href="https://groups.google.com/g/odsapp" target="_blank" rel="noopener">Beta program list</a></span>
                 <a href="https://groups.google.com/g/odsapp" target="_blank" rel="noopener" class="btn btn-outline-secondary mb-md-3">Register</a>
               </li>
-              <li class="vr mx-auto my-4 bg-light custom-vr"></li>
+              <li class="vr mx-auto my-4 custom-vr"></li><!-- Need an extra bg-* to handle this use case ? -->
```
julien-deramond, line 289:
```suggestion
              <li class="vr mx-auto my-4 custom-vr"></li>
```

> Need an extra bg-* to handle this use case?

We'll see that. It depends on how we are handling `.border-tertiary` + `.text-tertiary`. If the color is the same as `--bs-border-color-translucent`, we might be able to use `.bg-tertiary` here for instance.

julien-deramond:
In terms of pure rendering, these icons should be black:

Screenshot 2023-11-16 at 07 46 27

PR #2369, opened 2023-11-10
Add color mode on Masonry example

louismaximepiton:
### Description

⚠️ Might be merged once `data-bs-theme="inverted"` is introduced.

### Links

- https://deploy-preview-2369--boosted.netlify.app/docs/5.3/examples/masonry

### DIFF
--- a/site/content/docs/5.3/examples/masonry/index.html
+++ b/site/content/docs/5.3/examples/masonry/index.html
@@ -57,8 +57,8 @@ aliases:
       </div>
     </div>
     <div class="col-sm-6 col-lg-4 mb-4">
-      <div class="card text-bg-secondary border-secondary text-center p-3">
-        <figure class="mb-0" data-bs-theme="inverted">
+      <div class="card bg-black border-dark text-center p-3" data-bs-theme="dark">
+        <figure class="mb-0">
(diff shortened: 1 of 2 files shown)


site/content/docs/5.3/examples/masonry/index.html
```diff
@@ -57,12 +57,12 @@ <h5 class="card-title">Card title</h5>
       </div>
     </div>
     <div class="col-sm-6 col-lg-4 mb-4">
```
julien-deramond, line 59: Based on the current rules that we have, there's no such cases as `.text-bg-secondary` available for cards in the DSM. So I would suggest something like this (to rework if new rules happen):

```html
<div class="col-sm-6 col-lg-4 mb-4">
  <div class="card bg-black border-dark text-center p-3" data-bs-theme="dark">
    <figure class="mb-0">
      <blockquote class="blockquote">
        <p>A well-known quote, contained in a blockquote element.</p>
      </blockquote>
      <figcaption class="blockquote-footer mb-0">
        Someone famous in <cite title="Source Title">Source Title</cite>
      </figcaption>
    </figure>
  </div>
</div>
```

Thoughts?

PR #2374, opened 2023-11-14
Docs: Homepage tweaks

louismaximepiton:
### Description

Proposals to get closer to Bootstrap. One of them should happen anyway in dark mode branch.

### Links

- https://deploy-preview-2374--boosted.netlify.app/


julien-deramond (approved):
For the "Quickly customize components", even if the rendering is not compatible with the DSM, I think this is the best approach because we want to show that it's customizable, not compatible. LGTM!

PR #2375, opened 2023-11-14
Dark mode: Make docs callouts more highlighted

louismaximepiton:
### Description

Proposal to make docs callouts more highlighted than before in light and dark mode.

### Links

- https://deploy-preview-2375--boosted.netlify.app/docs/5.3/docsref


julien-deramond (approved):
The callout is way more emphasized this way! Neat!
Color contrast stays AA even with code texts so LGTM

PR #2376, opened 2023-11-14
Dark mode: Change syntax colors

louismaximepiton:
### Description

Proposal to make docs code highlights with differents colors in dark mode.

### Links

- https://deploy-preview-2376--boosted.netlify.app/docs/5.3/docsref


julien-deramond:
In your original proposal, the background of the area was darker isn't it?

PR #2377, opened 2023-11-14
Remove useless CSS since dropdown have been changed

louismaximepiton:
### Description

Remove some useless CSS due to our new implementation of dropdowns.
Make sure the docs header works as expected.

### Links

- https://deploy-preview-2377--boosted.netlify.app/

### DIFF
--- a/site/assets/scss/_navbar.scss
+++ b/site/assets/scss/_navbar.scss
@@ -13,20 +13,10 @@
 
   // Boosted mod: no .navbar-brand
 
-  // Boosted mod: Needed since the dropdowns define their own color (could be removed once they are tackled)
-  .nav-link {
-    &:hover {
-      color: var(--bs-link-hover-color);
-    }
-
-    &.active {
-      color: var(--bs-link-hover-color);
-    }
-
-    .bi {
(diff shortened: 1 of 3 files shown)


site/assets/scss/_navbar.scss
```diff
@@ -13,20 +13,10 @@
...
-
-    &.active {
-      color: var(--bs-link-hover-color);
-    }
-
-    .bi {
-      width: var(--bd-navbar-dropdown-item-icon-size);
-      height: var(--bd-navbar-dropdown-item-icon-size);
-    }
+  // Boosted mod: Needed since our icons are larger in the navbar
```
julien-deramond, line 16: Not sure about this whole change for the icons. We checked with the designers the rendering of the icons within the dropdown which is changed here. Now we have really big icons:

Screenshot 2023-11-15 at 08 00 48

Maybe let's wait for the design specifications of this theme selector before changing the actual rendering.

louismaximepiton: I think that they are the same size as before, it should be transparent removal

julien-deramond (approved):

PR #2378, opened 2023-11-14
Dark mode: Proposals to handle dark mode code highlights

louismaximepiton:
### Description

Proposal to make docs code highlights less highlighted in dark mode. Adding some small changes in the borders as well.
Refactor Ratio examples to better fit dark mode. This change should be done anyway.

### Links

- https://deploy-preview-2378--boosted.netlify.app/docs/5.3/docsref
- https://deploy-preview-2378--boosted.netlify.app/docs/5.3/helpers/ratios

### DIFF
--- a/site/assets/scss/_component-examples.scss
+++ b/site/assets/scss/_component-examples.scss
@@ -178,7 +178,7 @@
     font-weight: $font-weight-bold; // Boosted mod
     color: var(--bs-secondary-color);
     background-color: var(--bs-secondary-bg); // Boosted mod: instead of `var(--bs-tertiary-bg)`
-    border: var(--bs-border-width) solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-color)`
+    border: var(--bs-border-width) solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-width) solid var(--bs-border-color)`
 
     > div {
       display: flex;
--- a/site/assets/scss/_component-examples.scss
+++ b/site/assets/scss/_component-examples.scss
@@ -4,7 +4,7 @@
 
 .bd-code-snippet {
   margin: 0 ($bd-gutter-x * -.5) 1rem;
-  border: solid $gray-400; // Boosted mod: instead of `var(--bs-border-color)`
+  border: solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-color)`
   border-width: 1px 0;
 
   @include media-breakpoint-up(md) {
@@ -21,7 +21,7 @@
   position: relative;
   padding: var(--bd-example-padding);
   margin: 0 ($bd-gutter-x * -.5) 1rem;
-  border: solid $gray-400; // Boosted mod: instead of `var(--bs-border-color)`
+  border: solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-color)`
   border-width: 1px 0;
   @include clearfix();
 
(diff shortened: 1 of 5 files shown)


site/assets/scss/_component-examples.scss
```diff
@@ -176,9 +176,9 @@
     display: inline-block;
     width: 9rem; // Boosted mod
     font-weight: $font-weight-bold; // Boosted mod
-    color: $gray-800; // Boosted mod: instead of `var(--bs-secondary-color)`
-    background-color: $gray-400; // Boosted mod: instead of `var(--bs-tertiary-bg)`
-    border: var(--bs-border-width) solid $gray-500; // Boosted mod
+    color: var(--bs-secondary-color);
+    background-color: var(--bs-secondary-bg); // Boosted mod: instead of `var(--bs-tertiary-bg)`
+    border: var(--bs-border-width) solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-color)`
```
julien-deramond, line 181:
```suggestion
    border: var(--bs-border-width) solid var(--bs-border-color-translucent); // Boosted mod: instead of `var(--bs-border-width) solid var(--bs-border-color)`
```

site/assets/scss/_component-examples.scss
```diff
@@ -314,7 +314,7 @@
   min-height: 16.625rem; // Boosted mod: instead of `15rem`
 
   > div {
-    color: var(--bs-body-bg);
+    color: var(--bs-highlight-color);
```
julien-deramond, line 317:
```suggestion
    color: var(--bs-highlight-color); // Boosted mod: instead of `var(--bs-body-bg)`
```

site/assets/scss/_component-examples.scss
```diff
@@ -439,6 +439,6 @@
 // scss-docs-end sticker-fs-xl
 
 .border-gray-400 {
```
julien-deramond, line 441: You should get rid of this specific Boosted class or rename it. It doesn't exist in Bootstrap, so it's not logical to have something named `.border-gray-400` which doesn't apply `$gray-400`. Either `.border-color-translucent`, either something with context to avoid changing the name each time the value changes, either directly a `.border-*` from the framework which has `--bs-border-color-translucent` as a value.

site/assets/scss/_variables.scss
```diff
@@ -39,7 +39,7 @@ $bd-callout-variants: info, warning, danger !default;
   // Boosted mod: no --bd-sidebar-link-bg
   // Boosted mod: no --bd-callout-link
   // Boosted mod: no --bd-callout-code-color
-  --bd-pre-bg: #{adjust-color($gray-900, $lightness: -2.5%)}; // stylelint-disable-line scss/at-function-named-arguments
```
julien-deramond, line 42: `var(--bd-pre-bg)` is still used two times in this branch in __component-examples.scss_.

julien-deramond (changes requested):

site/assets/scss/_component-examples.scss
louismaximepiton: We don't have such a class in Boosted right now, `.border-tertiary` was handling this variable. Should we introduce it for the docs only in a first time ?

site/assets/scss/_variables.scss
louismaximepiton: Yes, but we don't need to overwrite it in dark mode. The light mode definition is already fine to me. However, it changes the background-color of the code examples. Are you fine with this change ?

site/assets/scss/_component-examples.scss
julien-deramond: I think this would be the best approach for now

julien-deramond: We agreed on changing it to `.border-tertiary`

julien-deramond (approved):

PR #2379, opened 2023-11-14
Docs: Micro anim on docs guidelines page

louismaximepiton:
### Description

Add a micro animation as there was before on the DSM.
Anyway the color change might be done on dark mode branch.

### Previews

- https://deploy-preview-2379--boosted.netlify.app/docs/5.3/guidelines

### DIFF
--- a/site/assets/scss/_boosted.scss
+++ b/site/assets/scss/_boosted.scss
@@ -70,8 +70,8 @@ body {
       width: 100%;
       height: 0;
       content: "";
-      background-color: var(--bs-link-hover-color);
-      transition: height ease-out .2s; // stylelint-disable-line property-disallowed-list
+      background-color: $accessible-orange;
+      @include transition(height ease-out .2s);
     }
 
     &:hover::before {
--- a/site/assets/scss/_boosted.scss
+++ b/site/assets/scss/_boosted.scss
@@ -62,13 +62,20 @@ body {
 .ods-guidelines {
   background-color: var(--bs-secondary-bg);
 
-  .card a:hover::before {
-    position: absolute;
-    bottom: 0;
-    left: 0;
-    width: 100%;
-    height: 4px;
-    content: "";
-    background-color: $accessible-orange;
+  .card a{
+    &::before {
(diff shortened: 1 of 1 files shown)


julien-deramond:
Can you please check with the designers that it's something they want. We generally don't have animations; not that they are not wanted, but rather because we don't have any rules about animations anywhere.
We should keep the previous color in this PR because we're not sure this custom property will be actually used even in the dark mode branch. Let's keep something consistent in the _main_ branch.

site/assets/scss/_boosted.scss
```diff
@@ -62,13 +62,20 @@ body {
...
+  .card a{
+    &::before {
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      width: 100%;
+      height: 0;
+      content: "";
+      background-color: var(--bs-link-hover-color);
+      transition: height ease-out .2s; // stylelint-disable-line property-disallowed-list
```
julien-deramond, line 74: Don't you want to use our `transition` mixin here for a11y?

julien-deramond:
This micro animation has been validated by the designers. I let you check the comments in order to finalize this PR and merge it @louismaximepiton.

site/assets/scss/_boosted.scss
```diff
@@ -62,13 +62,20 @@ body {
...
-    background-color: $accessible-orange;
+  .card a{
+    &::before {
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      width: 100%;
+      height: 0;
+      content: "";
+      background-color: $accessible-orange;
```
julien-deramond, line 73: Note: change it to `background-color: var(--bs-link-hover-color);` in _main-jd-dark-mode_ branch when rebased

site/assets/scss/_boosted.scss
julien-deramond: Done via f0379a10b9a30c09d10bd33ec6684c4110edc303

julien-deramond (approved):
LGTM!

PR #2380, opened 2023-11-14
Dark mode: Toasts

louismaximepiton:
### Description

Tweaked a bit the documentation to be consistent.

Remaining question from the dark mode PR description:

> Need to have `data-bs-theme="inverted"` for `.*bg-secondary` (toasts color schemes)? 

### Previews

- https://deploy-preview-2380--boosted.netlify.app/docs/5.3/components/toasts
- https://deploy-preview-2380--boosted.netlify.app/docs/5.3/dark-mode/#toasts

### DIFF
--- a/site/content/docs/5.3/components/toasts.md
+++ b/site/content/docs/5.3/components/toasts.md
@@ -95,7 +95,7 @@ We use the following JavaScript to trigger our live toast demo:
 
 Toasts are slightly translucent to blend in with what's below them.
 
-{{< example class="bg-dark" stackblitz_add_js="true" >}}
+{{< example class="bg-secondary" stackblitz_add_js="true" >}}
 <div class="toast" role="alert" aria-live="assertive" aria-atomic="true">
   <div class="toast-header">
     {{< placeholder width="20" height="20" background="#ff7900" class="me-2" text="false" title="false" >}}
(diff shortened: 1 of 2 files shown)


site/content/docs/5.3/components/toasts.md
```diff
@@ -172,10 +172,12 @@ Alternatively, you can also add additional controls and components to toasts.
 
 ### Color schemes
 
-Building on the above example, you can create different toast color schemes with our [color]({{< docsref "/utilities/colors" >}}) and [background]({{< docsref "/utilities/background" >}}) utilities. Here we've added `.text-bg-secondary` to the `.toast`. For a crisp edge, we remove the default border with `.border-0`.
+Building on the above example, you can create different toast color schemes with our [color]({{< docsref "/utilities/colors" >}}) and [background]({{< docsref "/utilities/background" >}}) utilities. Here we've added `.text-bg-dark` to the `.toast`. For a crisp edge, we remove the default border with `.border-0`.
+
+{{< callout-deprecated-dark-variants "toast" >}}
```
julien-deramond, line 177: This wasn't really a dark variant but I can understand why you're adding this here. However, in this case, we shouldn't need to use `.text-bg-dark` isn't it?

site/content/docs/5.3/components/toasts.md
```diff
@@ -95,7 +95,7 @@ We use the following JavaScript to trigger our live toast demo:
 
 Toasts are slightly translucent to blend in with what's below them.
 
-{{< example class="bg-dark" stackblitz_add_js="true" >}}
+{{< example class="bg-secondary" stackblitz_add_js="true" >}}
```
julien-deramond, line 98: We shouldn't use `.bg-secondary` here as it switches off to a white background in dark mode which doesn't exist on our side. To still show the feature, `.bg-body` doesn't work?

julien-deramond (changes requested):
Few things to finalize before merging

site/content/docs/5.3/components/toasts.md
louismaximepiton: We shouldn't but since we don't have the `data-bs-theme="inverted"` or something similar, we won't be able to use `.text-bg-secondary` when it's including other components. so to keep the same spirit at least in light theme, I went this way.

site/content/docs/5.3/components/toasts.md
louismaximepiton: The goal here is to show that the toast-body is transparent, I feel that with `.bg-secondary` we really see it well.

julien-deramond (approved):

PR #2406, opened 2023-12-13
Dark mode: Local nav and nav underline fix

louismaximepiton:
### Description

This PR safely removes `--bs-nav-link-active-color`.

Following design private conversation: 
>  regarding the Local navigation Selected state Label, the label should change to White on the dark mode version with the orange underline. This would make it consistent with the Light mode version.
> 
> the Hover states should be the same on both the light and dark mode versions. The hover state should use an orange coloured label.
> 
> the same changes should be applied to the Tabs Underline component selected and hover states in light and dark modes.

The artboards have been updated.

I tweaked a bit the associated CSS to follow these new rules. (Less Generated CSS)

Should be transparent for users (apart for the look).

### Live previews

- https://deploy-preview-2406--boosted.netlify.app/docs/5.3/dark-mode/#local-navigation
- https://deploy-preview-2406--boosted.netlify.app/docs/5.3/dark-mode/#navs--tabs

### DIFF
--- a/scss/_local-navigation.scss
+++ b/scss/_local-navigation.scss
@@ -20,7 +20,7 @@
     width: 100%;
     padding: subtract(var(--#{$prefix}local-nav-padding-y), var(--#{$prefix}local-nav-border-width)) 0 var(--#{$prefix}local-nav-padding-y);
     font-weight: $font-weight-bold;
-    color: var(--#{$prefix}local-nav-hover-color);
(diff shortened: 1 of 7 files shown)


julien-deramond:
Discussed with the designers, we need to use `#141414` as background color

julien-deramond (approved):
LGTM!

PR #2411, opened 2023-12-13
Dark mode: Docs about background utility

louismaximepiton:
### Description

This PR tackles the following sub-task defined in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2223 description:

> **Background**
> Doc: explain the rules between  `.bg-primary|success|etc.` and `.bg-supporting-*` (how to use them)

Texts might need to be rewrote.

#### Live preview

- https://deploy-preview-2411--boosted.netlify.app/docs/utilities/background

### DIFF
--- a/site/content/docs/5.3/utilities/background.md
+++ b/site/content/docs/5.3/utilities/background.md
@@ -1,7 +1,7 @@
 ---
 layout: docs
 title: Background
-description: Convey meaning through `background-color` and add decoration with gradients.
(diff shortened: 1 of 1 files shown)


julien-deramond:
We might explain that `.bg-body-tertiary` shouldn't be used; `#fafafa` does not exist in the color palette.

julien-deramond (approved):

PR #2417, opened 2023-12-14
Dark mode: Dropdown fix design

louismaximepiton:
### Description

Following the desin updates for the dark mode.
Also remove the focus state of the active state that is no longer relevant.

### Live previews

- https://deploy-preview-2417--boosted.netlify.app/dark-mode/#dropdowns
- https://deploy-preview-2417--boosted.netlify.app/docs/components/dropdowns


julien-deramond (approved):
Passed the design review. LGTM in terms of code.
The active visible focus rendering is better this way. We should do the same thing for list groups IMO.

PR #2427, opened 2023-12-19
Dark mode: drop bottom-right theme selector for Download page and Form examples

julien-deramond:
### Description

This PR drops the bottom-right theme selector for the download app and forms page examples. These examples now contain a theme selector at the right place in the header.

### Motivation

This is done this way to avoid users misunderstanding where to put the theme selector on a real Orange page.

### Live previews

- https://deploy-preview-2427--boosted.netlify.app/examples/navbars/ (bottom-right theme selector)
- https://deploy-preview-2427--boosted.netlify.app/examples/download-app/ (NO bottom-right theme selector)
- https://deploy-preview-2427--boosted.netlify.app/examples/form/ (NO bottom-right theme selector)


site/layouts/_default/examples.html
```diff
@@ -95,6 +95,7 @@
     {{- end }}
   </head>
   <body{{ with .Page.Params.body_class }} class="{{ . }}"{{ end }}>
+    {{ if ne .Page.Params.theme_selector false }}
     <svg xmlns="http://www.w3.org/2000/svg" style="display: none;">
```
louismaximepiton, line 99: Maybe we could keep this svg in each example since we introduce the theme selector anyway ? If yes, we could remove newly added svgs from the examples.

louismaximepiton (changes requested):
One thing to discuss before approval

site/layouts/_default/examples.html
julien-deramond: Done via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2427/commits/14de9ca1dd40a0f6372bcc6e5679fb683bb677c4

louismaximepiton (approved):

PR #2432, opened 2023-12-20
Navbar: Handle in a proper way our dropdowns inside documentation navbar

louismaximepiton:
### Description

We had an issue when we wanted to explore an overcharged navbar with some dropdowns for example. 
We implement a little different behavior than Bootstrap since we have much complex navbars.
Reorganize docs navbar to have something more branded navbar (more spacing).

### Live previews

- https://deploy-preview-2432--boosted.netlify.app/docs/


scss/_navbar.scss
```diff
@@ -456,7 +456,8 @@
         }
         // End mod
 
-        .navbar-nav-scroll {
+        .navbar-nav-scroll,
+        &.navbar-nav-scroll {
```
julien-deramond, line 460: Since we're not sure yet about a generic solution, isn't it safer to use `.overflow-visible` in `site/layouts/partials/docs-navbar.html`, don't modify this rule, and don't add a new explanation in `site/content/docs/5.3/components/orange-navbar.md` that might not work in all cases?

site/assets/scss/_navbar.scss
```diff
@@ -51,8 +51,23 @@
...
+      @include media-breakpoint-up(md) {
+        line-height: 1.875rem;
+      }
+    }
   }
   // End mod
 }
 
+// Boosted mod
+.main-header {
```
julien-deramond, line 67: The dropdown menu is not visible in desktop mode, how about we change it to the following instead? Does it fit what you had in mind?

```scss
@include media-breakpoint-down(lg) {
  .main-header {
    max-height: 100vh;
    overflow: hidden auto;
  }
}
```

site/layouts/partials/docs-versions.html
```diff
@@ -10,11 +10,11 @@
...
 {{- end }}
 
-<li class="nav-item dropdown d-flex">
-  <a href="#" class="nav-link dropdown-toggle gap-1 mx-0 bg-transparent" data-bs-toggle="dropdown" aria-expanded="false" data-bs-display="static">
-    <span class="d-lg-none" aria-hidden="true">Boosted</span><span class="visually-hidden">Boosted&nbsp;</span> v{{ .Site.Params.docs_version }} <span class="visually-hidden">(switch to other versions)</span>
-  </a>
-  <ul class="dropdown-menu dropdown-menu-end dropdown-menu-dark position-absolute">
+<li class="nav-item dropdown">
+  <button class="nav-link nav-icon dropdown-toggle" data-bs-toggle="dropdown" aria-expanded="false" data-bs-display="static">
+    <span class="d-lg-none" aria-hidden="true">Boosted&nbsp;</span><span class="visually-hidden">Boosted&nbsp;</span> v{{ .Site.Params.docs_version }} <span class="visually-hidden">(switch to other versions)</span>
```
julien-deramond, line 15:
```suggestion
    <span class="d-lg-none" aria-hidden="true">Boosted&nbsp;</span><span class="visually-hidden">Boosted&nbsp;</span>v{{ .Site.Params.docs_version }} <span class="visually-hidden">(switch to other versions)</span>
```

julien-deramond (approved):
LGTM, we can't do better right now.

Let's create an issue too regarding the fact that we had to add a `.bd-header` instead of using `.bd-navbar` as it is done on the Bootstrap side.

PR #2435, opened 2023-12-21
Dark mode: fix list group documentation

julien-deramond:
### Description

This PR fixes the list group documentation where `.list-group-item-Success` kind of classes were used instead of `list-group-item-success` due to the changes we made in the YML files.

### Live preview

- https://deploy-preview-2435--boosted.netlify.app/docs/5.3/components/list-group/#variants


louismaximepiton (approved):
Good catch!

PR #2531, opened 2024-04-02
Chore merge main from 2ba7dae to fb4629f

julien-deramond:
- [x] https://github.com/twbs/bootstrap/commit/2ba7dae3c080c5fac21f4d7f1663c8f524fefb26
- [x] https://github.com/twbs/bootstrap/commit/4093ed32ee6390223a885a676d0a3086257b8789
  - [x] Modify `boosted-guidelines.png` so that the rendering is the same as in [Bootstrap documentation](https://twbs-bootstrap.netlify.app/) at all breakpoints
- [x] ~https://github.com/twbs/bootstrap/commit/ba73ed6906226b9b6be37b01b91bc1a473417b28~: N/A since we don't have the same kind of images for Parcel, Vite, and webpack guides
- [x] https://github.com/twbs/bootstrap/commit/fb4629f82ee768bc52ccb90adf4e35176451e66a


site/content/docs/5.3/examples/_index.md
```diff
@@ -6,88 +6,3 @@ aliases:
   - "/examples/"
   - "/docs/examples/"
 ---
-
-{{< callout info >}}
```
julien-deramond, line 10: IMO, we can drop this since we never had any feedback and the issue is still in the backlog (we won't lose the information).

louismaximepiton (approved):
👌

PR #2573, opened 2024-06-25
[OUDS] feat: tokens architecture and opacity utilities

### LINKED ISSUE #2575: [OUDS] Tokens: Opacity
⚠️ **On hold** until https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/2597

---

## Description

The aim of this issue is to study and implement the opacity tokens taking into account the cohesive multi-platform approach, the customization by libraries inheriting OUDS Web, but also what we're using for OUDS Web: Boosted, Bootstrap, etc.
- Define the possibilities
- Define the constraints
- Define the impacts
- Implement the tokens (with a future Style Dictionary usage in mind)

Reminder: there will be in the end 3 layers of tokens:
- **Raw primitive values: Insertion of brand foundations.**
  - These tokens will be customizable by libraries inheriting OUDS Web.
  - These tokens cannot be used by developers using OUDS Web.
- **Semantic applications: Depends on raw primitive values.**
  - Mapping between these tokens and the raw primitive values will be customizable by libraries inheriting OUDS Web.
  - These tokens can be used by developers using OUDS Web.
- **Component applications: Depends on semantic applications.**:
  - Mapping between these tokens and the semantic tokens should rarely be customized by libraries inheriting OUDS Web.
  - These tokens can be used by developers using OUDS Web.

### Raw primitive values

|  Token name | Token value |
|---|---|
| opacity-0  | 0  |
| opacity-100  | 0.04  |
| opacity-200  | 0.08  |
| opacity-300  | 0.16  |
| opacity-400  | 0.24  |
| opacity-500  | 0.32  |
| opacity-600  | 0.48  |
| opacity-700  | 0.64  |
| opacity-800  | 0.88  |
| opacity-900  | 1  |

### Semantic applications

|  Token name | Token value |
|---|---|
| opacity-transparent | opacity-0 |
| opacity-weaker | opacity-100 |
| opacity-weak | opacity-300 |
| opacity-medium | opacity-500 |
| opacity-emphasis | opacity-700 |
| opacity-opaque | opacity-900 |

### Component applications

⚠️ This is just an example, it doesn't need to be implemented for now, it will be when the modals will be tackled

|  Token name | Token value |
|---|---|
| overlays-modal-backdrop-opacity | opacity-emphasis |

---

## Study

- [x] Study the architecture of these tokens, and define the technical details

### Technical details

For now, these tokens must not be applied to other existing tokens in `_variables.scss` (except if necessary to generate some opacity utilities).

#### Utilities level

It has a direct impact on the [Utilities > Opacity](https://boosted.orange.com/docs/5.3/utilities/opacity/) that must provide new utilities instead of the Boosted ones.

Before: `opacity-0`, `opacity-25`, etc.
Now: `opacity-transparent`, `opacity-weak`, etc.

Note that Bootstrap utilities must remain in the library.

The documentation must be modified for this utility.

#### Migration guide

Migration guide must be updated for:
- Boosted migration
- OUDS Web: new features

##### Impact on other utilities

Not to tackle right now: Impact on other utilities based on the opacity-{value} spirit:

- `.link-underline-opacity-{value}`
- `.bg-opacity-{value}`
- `.border-opacity-{value}`
- `.text-opacity-{value}`

### LINKED ISSUE #2578: [OUDS] Add a new `$enable-bootstrap-specificities`
⚠️ Improve the description!

### Description

- Find a better name
- Add a new `$enable-bootstrap-specificities` (default to `false`)
  - Add Sass testing with `$enable-bootstrap-specificities` = true to be sure that it always compile 
- Make it work for utilities
- Apply it on existing Bootstrap stuff that wasn't used before already by Boosted?
  - Integrate floating labels?
  - Some utilities, etc.
- Reintegrate https://boosted.orange.com/docs/5.3/customize/options/#content (a part of it at least)
- Create a ouds-web (by default / not Bootstrap), and a ouds-web-bootstrap (with Bootstrap)

### LINKED ISSUE #2597: [OUDS] Common cross-libraries token architecture
### Description

This issue aims to define and implement a common cross-libraries token architecture for the tech libraries: Android, iOS, Flutter, and web.

Design tokens are defined on the design side via Figma, and Figma's capability to handle and define tokens is more limited compared to what we can do on the development side. So there will be a slight discrepancy between the design and the tech side.

In essence, Figma’s token mechanism requires designers to pre-create a set of tokens that may not be inherently useful for the core part of our tech libraries.

On the technical side, it is important to define what is available by default:
* for developers using the Orange theme (default)
* what can be overridden for country-specific themes, and how overrides will function for B-Brand themes

This must be considered at different levels of the token stack, including primitive raw values, semantic tokens, and component tokens.

Our objective is to establish a common cross-library architecture to ensure uniformity in terminology and implementation.

> [!NOTE]  
> Closing this issue will unlock all the "Tokens: *" issues in this repository

> [!IMPORTANT]  
> The possibility to add semantic tokens from outside is a topic we'd like to deep dive in. However, creating prototypes for that now, is considered as too complex as we don't have yet libraries handling tokens. The safest approach is to limit the scope of this study.

Based on the study done on the Android side in https://github.com/Orange-OpenSource/ouds-android/issues/34, in terms of customization, our libraries will need to handle first only a light customization such as white-label themes.

Said differently, our libraries will define only the raw/semantics/components tokens needed for the global core Orange theme. And only these tokens would be updatable. It also means that our tech libraries won't contain at first the concepts coming from B-Brands that are not used in the core library. Basically, it's like what has been done in ODS, but with a new constraint: the theme should be extendable.
An extended theme can rewire semantic tokens with its own raw tokens and component tokens with semantic tokens. It defines its own raw tokens (often generated from Figma) and doesn't need to access the raw tokens of the parent theme.

It corresponds to the schema presented in the "Android architecture" section in https://github.com/Orange-OpenSource/ouds-android/issues/34.

#### Implementation

As a summary, for the implementation part:
- Implement the architecture defined in the "Technical details" section
- Create a fake component in order to test the usage
- Check that the documentation can access the semantic and components tokens
- Create a mock library using OUDS Web to customize the dummy component for a light-themed environment, like a white-label library.
- Create a theme extending the Orange theme which has its own raw tokens and can rewire Orange semantics and/or components tokens

### Study

- [x] Study the limitations and the way it could be done on Web

#### Technical details

> [!WARNING]  
> The following elements could need to be updated after the results of the Android, iOS, and Flutter corresponding studies. We'll try to make rather the same technical choices, and interfaces for the basic users, and the users having to create libraries for countries or B-Brands on top of OUDS Web.

##### Context

The current implementation regarding tokens can be summarized as follows:
- `_color-palette.scss` file contains the raw color values of the ODS color palette.
- `_variables.scss`, inherited from Bootstrap, contains the different levels of variables used in the framework, regardless of their place in the layers of tokens (raw, semantic, component).

##### Evolution

The goal is to prepare the field for the future integration of the tokens of our design system via the [Style Dictionary](https://amzn.github.io/style-dictionary/#/) tool, but also to be able to manage other libraries based on OUDS Web (for different themes, for example).

The idea, for now, is to create a `tokens` directory that will contain the following files:
- `_raw.scss` file with the raw tokens, the raw primitive values of the design system, such as hex colors, opacity values, etc.
- `_semantic.scss` file with the semantic tokens, the tokens that will bring some meaning to the raw tokens, such as primary color, secondary color, etc. They will depend on raw primitive values.
- `_component.scss` file with the component tokens, the tokens that will be used to build the components, such as button color, button background color, etc. They will depend on semantic tokens only.

These files will for now be filled out manually based on our design guidelines. The idea in the future is to generate these files automatically via the Style Dictionary tool, from the JSON file exported from Figma.

##### Actions to take

The aim of this task is to:
1. create this Sass architecture
2. update the documentation accordingly to explain how to use it (e.g. Sass customization)
3. explain the new feature in the migration guide
4. explain the impact of such feature in the migration guide from Boosted

julien-deramond:
### Raw and semantics tokens

In Bootstrap, the raw primitive values are in the `_variables.scss` file directly with different levels of semantics. To ease the customization, we introduced in Boosted a `_color-palette.scss` file containing the raw values of the colors.
We're introducing here a new directory (more generic this time), named `tokens` containing three types of tokens:
* `scss/tokens/_raw.scss`: raw primitive values
* `scss/tokens/_semantic.scss`: semantic applications depending on raw tokens
* `scss/tokens/_component.scss`: component applications depending on semantic tokens

The idea is that a B-Brand or a country customizing our library (to create another specific library) would only change the `scss/tokens/_raw.scss`, and replug the right tokens into the `scss/tokens/_semantic.scss`, and possibly into `scss/tokens/_component.scss`, to associate the right semantics.

We moved the `_color-palette.scss` from `_variables.scss` to `tokens/_raw.scss` as it represents raw tokens from ODS. This file will be removed whenever we implement the OUDS color tokens.

#### Opacity tokens

In order to put in place this new tokens' architecture, this PR also puts in place the opacity tokens as an example as defined in #2575:
* `scss/tokens/_raw.scss` contains the raw tokens from `$ouds-opacity-0` to `$ouds-opacity-900`, with values from 0 to 1
* `scss/tokens/_semantic.scss` contains the semantic tokens defined by Orange Unified Design System from "transparent" to "opaque".
*  `scss/tokens/_component.scss` is empty because we haven't yet components

Please note that all tokens are prefixed by `$ouds-` to differenciate them from Bootstrap tokens.

#### Opacity utilities

The semantic tokens are currently defined as maps, making them ready to be looped over for creating corresponding utilities.

So, from these semantic tokens, we'd like to propose to the developers to have access to utilities:
* `opacity-transparent`
* `opacity-weaker`
* `opacity-weak`
* `opacity-medium`
* `opacity-emphasis`
* `opacity-opaque`

These utilities will be the same, whatever the customization happening.

#### Opacity utilities from Bootstrap

Opacity utilities from Bootstrap must be kept as a plugin or library based on Bootstrap should work when our CSS is replacing Bootstrap one.

Given the fact that these utilities match exactly a value, and not a semantic concept, I don't think we could say that `opacity-50` should be modified to use `$opacity-600`, which is 0.48, as it could have unexpected effects when integrated in a Bootstrap-based library or plugin with our CSS file.

**However**, doing that will make our CSS bundle grow because we would have basically 2 times the almost same utilities.

The idea in this PR is to:
1. Keep that double set of almost the same utilities for the pre-built libraries available with the CDN
2. Add a customizable new Sass variable that would allow extracting unused Bootstrap stuff, utilities but not only. This variable is temporarily named `$enable-bootstrap-compatibility` and will be set to `false` by default.
  * Thanks to this new Sass variable, it's also possible to exclude some utilities from the bundle with `bootstrap-compatibility: true`

|  Branch  |  `$enable-bootstrap-specificities`  | `dist/css/boosted.css` size  |
|---|---|---|
| `main`  | N/A  |  360 122  |
|  `main-jd-uds-token-opacity-study`  |  `true`  | 360 417  |
|  `main-jd-uds-token-opacity-study`  |  `false` (default)  | 360 199  |

#### Impact on other utilities

It's not done in this PR, but some other utilities would be impacted as well as based on the `opacity-{value}` spirit:
- `.link-underline-opacity-{value}`
- `.bg-opacity-{value}`
- `.border-opacity-{value}`
- `.text-opacity-{value}`

### Documentation

The documentation will make appear the new utilities, and would keep the Bootstrap one semi-hidden, as it still could be used anyway with the full CSS build when `$enable-bootstrap-compatibility` is set to `true`.

#### Other cases

- In `api.md`, we probably can keep the opacity example as is since it explains rather easily how it works even if these utilities won't be anymore in the CSS bundle (when Bootstrap stuff is removed)

### Migration guide

We'll need to have a part explaining these new utilities for OUDS Web. But also from the Boosted users' perspective, to allow them to migrate to UDS Web.

For Boosted users migrating to our new version, we must explain:
* New utilities: `opacity-opaque` → `opacity-transparent`
* Deprecated utilities: `opacity-100` → `opacity-0`
  * What are the replacement values, the closest values, so that the users can research/replace easily

### Modal backdrop

Based on the values in the description, we already know what is the new value for the modal backdrop, so this PR also changes that, and link the old value to the new corresponding semantic Sass variable. 

## Live previews

- Opacity utilities: https://deploy-preview-2573--boosted.netlify.app/docs/0.0/utilities/opacity/
- Migration guide: https://deploy-preview-2573--boosted.netlify.app/docs/0.0/migration
- Migration from Boosted guide: https://deploy-preview-2573--boosted.netlify.app/docs/0.0/migration-from-boosted
- New `$enable-bootstrap-compatibility`: https://deploy-preview-2573--boosted.netlify.app/docs/0.0/customize/options/

Closes #2575

Closes #2597

Closes #2578


scss/_variables.scss
```diff
@@ -376,6 +377,18 @@ $light-border-subtle:     $light !default; // Boosted mod: instead of `$gray-200
 $dark-border-subtle:      $dark !default; // Boosted mod: instead of `$gray-500`
 // scss-docs-end theme-border-subtle-variables
 
+// Boosted mod
+// Semantic tokens
+
+// Opacity
```
julien-deramond, line 383: Feedback from @louismaximepiton 
We could use directly a map here to better handle adding/removing semantic concepts in B-Brands.

site/content/docs/5.3/utilities/opacity.md
```diff
@@ -9,7 +9,43 @@ aliases:
...
+```html
+<div class="opacity-opaque">...</div>
+<div class="opacity-emphasis">...</div>
+<div class="opacity-medium">...</div>
+<div class="opacity-weak">...</div>
+<div class="opacity-weaker">...</div>
+<div class="opacity-transparent">...</div>
+```
+
+<details style="border: 1px dashed black" class="p-2 mb-4">
```
julien-deramond, line 38: The rendering must be improved, use class, and maybe use shortcodes as well. It's just for now a suggestion of rendering.

site/content/docs/5.3/migration.md
```diff
@@ -9,6 +9,15 @@ aliases:
 toc: true
 ---
 
+## From Boosted
+
+* New `$enable-bootstrap-specificities` option set to `true` by default.
+  * Must be a global migration note as it will impact not only the opacity utilities
+* New opacity utilities: `.opacity-transparent`, etc.
+* Deprecated opacity utilities: `.opacity-0`, etc.
```
hannahiss, line 17: Just to be noted: you can't use the vocabulary "deprecated" since these utilities will never be removed

julien-deramond: Fixed via https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2573/commits/625c76148b1fa68ae0ed089e8ef55f5065f96999 for this temp version really quickly written (just to not forget it)

scss/_tokens.scss
```diff
@@ -0,0 +1,12 @@
+// Opacity
```
vprothais, line 1: Will this file be directly extracted from the design token management tools in the future ?
If not I would maybe call it core-tokens and set system/semantic tokens in another token file than variables.

scss/_variables.scss
vprothais: I confirm, having the values directly declared in utilities is not handy. I had to create a local copy of the utilities file in order to remove some values.

scss/_tokens.scss
julien-deramond: > Will this file be directly extracted from the design token management tools in the future ?

Yes it will, but we need to define if we split this file or not.

> If not I would maybe call it core-tokens

The name can be changed, no problem. I just created a fake file for now, waiting for a cool name for it :)

> and set system/semantic tokens in another token file than variables

Not entirely sure whether we could do that. We probably could, but it will also depend on how these system/semantic tokens will depend on each other.

vprothais: yeah, in fact I was thinking about b-brand and countries and how to make it simpler to know where I can change values. core-tokens should not be changed, and for variables, well it will be a mix of a lot of things, not mensionning utilities. It's still a little obscure for me on some point on how we will map values from the dictionnary to the format attented by Bootstrap. I was thinking at first as a sort of mapping cascade but it's hard to keep genericity.

julien-deramond: Yeah, totally agree with your point to clarify usage, but I'm not yet sure on how to find the right balance between files, especially at the semantic level, versus/with the `_variables.scss`. 

I think that in v6, Bootstrap should split this file, but since it's not the case yet, we need to find the best approach for basic users, and B-Brands.

Let's define together what's our starting point for these first PRs (to start with something), and let's challenge ourselves regularly, and this system, to enhance it along the way.
We'll be safe while being still in v0.x versions, so we'll be able to experiment 🔬

scss/_variables.scss
julien-deramond: Here's a separate commit that shows how it could be implemented based on this feedback: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2573/commits/b530b18c78155e71bab768550492dff1d22df624
Is this what you had in mind @louismaximepiton and @vprothais?

Note: it can be more difficult to generate from Style Dictionary, but would a problem for our future selves 😄

vprothais: yes, I find it better. If we could use only map in utilities it would be great

scss/_tokens.scss
julien-deramond: Based on this discussion and internal ones with the rest of the team, I tried out something with https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2573/commits/1fb9a33a3e45dd591a3d308bc7141dc0b6df4f87

scss/mixins/_utilities.scss
```diff
@@ -2,81 +2,71 @@
...
-    $values: zip($values, $values);
-  }
-
-  @each $key, $value in $values {
-    $properties: map-get($utility, property);
-
-    // Multiple properties are possible, for example with vertical or horizontal margins or paddings
-    @if type-of($properties) == "string" {
-      $properties: append((), $properties);
+  @if ($bootstrap-compatibility != true or $enable-bootstrap-compatibility == true) { // OUDS mod: filter Bootstrap specificities
```
julien-deramond, line 7: It's pretty obvious so we can probably drop the detail of the comment:
```suggestion
  @if ($bootstrap-compatibility != true or $enable-bootstrap-compatibility == true) { // OUDS mod
```

scss/ouds-web-reboot.scss
julien-deramond, line 1: These imports can be removed as they are not needed to compile

scss/ouds-web-utilities.scss
```diff
@@ -3,6 +3,9 @@
 
 // Configuration
 @import "functions";
+@import "tokens/raw";
+@import "tokens/semantic";
+@import "tokens/component";
```
julien-deramond, line 8: `@import "tokens/component";` is not needed to compile

scss/ouds-web-grid.scss
```diff
@@ -4,6 +4,9 @@
 $include-column-box-sizing: true !default;
 
 @import "functions";
+@import "tokens/raw";
+@import "tokens/semantic";
+@import "tokens/component";
```
julien-deramond, line 9: `@import "tokens/component";` is not needed to compile

site/assets/scss/docs.scss
julien-deramond, line 1: Among the 6 new imports, you can import only these:
```scss
@import "../../../scss/tokens/raw";
@import "../../../scss/tokens/semantic";
```

site/assets/scss/docs.scss
```diff
@@ -24,15 +24,22 @@
...
+@import "../../../scss/utilities";
 
 // fusv-disable
 $enable-grid-classes: false;
 $enable-cssgrid: true;
+$enable-bootstrap-compatibility: true;
 // fusv-enable
 @import "../../../scss/grid";
 
+
```
julien-deramond, line 42: Extra line to remove

site/content/docs/0.0/migration.md
```diff
@@ -1,14 +1,49 @@
...
+title: Migrating to recent versions
+description: Track and review changes to the OUDS Web source files, documentation, and components to help you migrate from previous versions.
 group: migration
 aliases:
   - "/migration/"
   - "/docs/migration/"
 toc: true
 ---
 
+## v0.0.2
```
julien-deramond, line 12: Next one will be v0.0.3. v0.0.2 is already released.

```suggestion
## v0.0.3
```

scss/ouds-web-utilities.scss
vprothais: Since the PR is for the token architecture and documentation, should'nt we have this imported even if the file is empty ? You think it should be added only when we have our first component ? Do we completely delete the tokens/component file if it is not used anywhere ?

package.json
```diff
@@ -41,7 +41,7 @@
     "css-rtl": "cross-env NODE_ENV=RTL postcss --config build/postcss.config.mjs --dir \"dist/css\" --ext \".rtl.css\" \"dist/css/*.css\" \"!dist/css/*.min.css\" \"!dist/css/*.rtl.css\"",
     "css-lint": "npm-run-all --aggregate-output --continue-on-error --parallel css-lint-*",
     "css-lint-stylelint": "stylelint \"**/*.{css,scss}\" --cache --cache-location .cache/.stylelintcache",
-    "css-lint-vars": "fusv scss/ site/assets/scss/",
+    "css-lint-vars": "fusv scss/ site/assets/scss/ --ignoreFiles \"**/tokens/_raw.scss\"",
```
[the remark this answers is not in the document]
vprothais: Like you said, in the end it would be simpler to exclude the file, but for now we can comment unused variables since the full publication process is not ready. I don't have strong opinion regarding that.

scss/ouds-web-utilities.scss
julien-deramond: I'd keep it for `ouds-web.scss` since it will be used in the end for sure. For the other `ouds-web-*.scss` files, it's not mentioned in the documentation, so I'd import only the necessary files.

package.json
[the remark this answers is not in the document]
julien-deramond: I would remove this then, as maybe we won't have anything to comment in the future if our Style Dictionary parser is smart enough.

scss/ouds-web-reboot.scss
louismaximepiton: Not quite sure for this one, we have the default link color etc, I tried something in the last commit.

scss/tokens/_raw.scss
julien-deramond, line 1: I'm wondering whether we shouldn't drop right away `_color-palette.scss` and temporarily store its content in `scss/tokens/_raw.scss`.

site/content/docs/0.0/migration-from-boosted.md
```diff
@@ -51,8 +51,17 @@ Technically, it means that you can get rid of the following things:
...
+## Utilities
+
+### Opacity
+
+- <span class="badge text-bg-danger">Breaking</span> `.opacity-0`, `.opacity-25`, `.opacity-50`, `.opacity-75` and `.opacity-100` have been removed from the default build. Please check the new [opacity values]({{< docsref "/utilities/opacity" >}}) directly in the documentation and adapt your websites to them. You can still have them using `$enable-bootstrap-compatibility`.
+- <span class="badge text-bg-success">New</span> Opacity utilities: `.opacity-transparent`, `.opacity-weaker`, `.opacity-weak`, `.opacity-medium`, `.opacity-emphasis` and `.opacity-opaque`.
+
 ## CSS and Sass variables
 
+- <span class="badge text-bg-success">New</span> `$enable-bootstrap-compatibility` option set to `false` by default. It allows you to have the Bootstrap utilities (using `*-0, *-1, *-50, ...` in addition to the semantic names like `transparent, weak, weaker, ...`).
```
julien-deramond, line 63: > It allows you to have the Bootstrap utilities (using `*-0, *-1, *-50, ...` in addition to the semantic names like `transparent, weak, weaker, ...`).

It's too specific to the opacity utilities. We should rather explain what it does, the generic approach.

site/data/sidebar.yml
```diff
@@ -31,10 +31,10 @@
   pages:
     - title: Overview
       draft: true
+    - title: Lib for Orange
```
julien-deramond, line 34: TBH, I'm not a big fan of this wording. Isn't "Library" enough? Or what do you think about "Custom Libraries"?

.storybook/preview-head.html
```diff
@@ -2,7 +2,7 @@
 <link href="https://cdn.jsdelivr.net" rel="preconnect" crossorigin="anonymous">
 
 <!-- OUDS Web CSS -->
-<link href="https://cdn.jsdelivr.net/npm/@ouds/web/dist/css/ouds-web.min.css" rel="stylesheet" crossorigin="anonymous">
+<link href="https://cdn.jsdelivr.net/npm/@ouds/web/dist/css/ouds-web-bootstrap.min.css" rel="stylesheet" crossorigin="anonymous">
```
julien-deramond, line 5: Do we need to rely on `ouds-web-bootstrap.min.css` for Storybook as we're for now only will use it for ZeroHeight to show specific OUDS content?

README.md
```diff
@@ -112,6 +120,7 @@ Within the download you'll find the following directories and files, logically g
 
 We provide compiled CSS and JS (`ouds-web.*`), as well as compiled and minified CSS and JS (`ouds-web.min.*`). [Source maps](https://web.dev/articles/source-maps) (`ouds-web.*.map`) are available for use with certain browsers' developer tools. Bundled JS files (`ouds-web.bundle.js` and minified `ouds-web.bundle.min.js`) include [Popper](https://popper.js.org/docs/v2/).
 
+OUDS Web also provides a compiled and minified CSS including Bootstrap compatibility (`ouds-web-bootstrap.*` and `ouds-web-bootstrap.min.*`).
```
julien-deramond, line 123:
```suggestion
We also provide a compiled and minified CSS (`ouds-web-bootstrap.*` and `ouds-web-bootstrap.min.*`) enforcing Bootstrap compatibility.
```

scss/tokens/_raw.scss
louismaximepiton: Not a big fan of this, I'd rather keep the `_raw.scss` file with the already accepted tokens. It shouldn't take long to remove the color-palette anyway

.storybook/preview-head.html
louismaximepiton: Yeah maybe we can use the default one for now and change it if needed after, I had in mind a use case that shouldn't happen.

scss/tokens/_raw.scss
julien-deramond: > but maybe we can change the import to import it in the raw file ?

Yep, if you prefer, you can do that. It's just so that we don't mention this file in the documentation, and can finalize the final Sass compilation without it.

scss/ouds-web-reboot.scss
julien-deramond: Ah yeah, you'll need them so that `_variables.scss` can compile since it embeds all kinds of tokens unfortunately. Nevermind then, let's resolve this comment.

site/layouts/partials/stylesheet.html
```diff
@@ -1,11 +1,11 @@
 {{ if hugo.IsProduction -}}
 {{ if eq .Page.Params.direction "rtl" -}}
-<link href="/docs/{{ .Site.Params.docs_version }}/dist/css/ouds-web.rtl.min.css" rel="stylesheet" {{ printf "integrity=%q" .Site.Params.cdn.css_rtl_hash | safeHTMLAttr }}>
```
julien-deramond, line 3: Note for future selves: we decided not to handle the `integrity` here in order to simplify the entire process, and to avoid creating new "hash" entries in `hugo.yml`.

PR #2593, opened 2024-07-01
[OUDS] Drop deprecated `.btn-close-white`

julien-deramond:
### Description

This PR drops the deprecated set of classes, CSS, and Sass variables around `.btn-close-white`. Indeed, this was kept in Boosted because deprecated, but OUDS Web doesn't need to start with deprecated classes in general (except if they come from Bootstrap, whose we're a fork of).

The migration guide from Boosted is updated to reflect this change → https://deploy-preview-2593--boosted.netlify.app/docs/0.1/migration-from-boosted/


louismaximepiton (approved):
🚀 Ok with the process

Issue #2611, opened 2024-07-10
[OUDS] [Tracking] Documentation content


julien-deramond:
⚠️ WIP: will need to be updated after https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2612

---

### Description

The content of the OUDS Web documentation has been cleaned up. However, given that we want iso-functional versions with Boosted/Bootstrap v5.3.3, it is essential to track what was previously included in both documentation libraries to determine whether any of this information should be reintegrated.

- [ ] Header (Boosted / Bootstrap)
  - [ ] Versioning dropdown (Boosted / ~Bootstrap~) 
  - [ ] Solaris icons entry (Boosted / ~Bootstrap~)
- [ ] Footer (Boosted / Bootstrap)
- [ ] Homepage
  - [ ] Jumbotron (Boosted / Bootstrap)
  - [ ] Get started any way you want (Boosted / Bootstrap)
  - [ ] Read our getting started guides (Boosted / Bootstrap)
  - [ ] Customize everything with Sass (Boosted / Bootstrap)
  - [ ] Build and extend in real-time with CSS variables (Boosted / Bootstrap)
  - [ ] Components, meet the Utility API (Boosted / Bootstrap)
  - [ ] Powerful JavaScript plugins without jQuery (Boosted / Bootstrap)
  - [ ] Orange Design System for web (Boosted / ~Bootstrap~)
- [ ] Dark mode page (Boosted / ~Bootstrap~)
- [ ] All versions (Boosted / Bootstrap)
- [ ] Design guidelines (Boosted / ~Bootstrap~)

---

```
WARN  Template _default/_markup/render-image.html is unused, source file /Users/ju/Orange-Boosted-Bootstrap/site/layouts/_default/_markup/render-image.html
WARN  Template partials/callouts/danger-async-methods.md is unused, source file /Users/ju/Orange-Boosted-Bootstrap/site/layouts/partials/callouts/danger-async-methods.md
WARN  Template partials/callouts/info-mediaqueries-breakpoints.md is unused, source file /Users/ju/Orange-Boosted-Bootstrap/site/layouts/partials/callouts/info-mediaqueries-breakpoints.md
WARN  Template partials/callouts/info-npm-starter.md is unused, source file /Users/ju/Orange-Boosted-Bootstrap/site/layouts/partials/callouts/info-npm-starter.md
[...]

PR #2612, opened 2024-07-10
[OUDS] Docs: empty pages content

julien-deramond:
## Description

This PR finalizes the clean-up of our OUDS Web main branch before starting the development:
- For pages that are mentioned in the side menu:
  - Remove the content of the Markdown files
  - Use the strike-through approach in the side menu to inform users that the content is not available
  - For each of these pages, a callout is displayed in the page to inform that the content is not yet available
  - Note: some easy pages in the About section are kept as they already work
- For the homepage, I kept only the content that is usable as is 
- The version dropdown has been simplified
- The "Design Guidelines" and "Solaris Icons" links have been removed
- The footer has been adapted with the available content
- The examples page is empty and all the examples have been removed
  - I've created a fake example (not mentioned in the examples page) just in order to keep the compilation/packaging system  

Please be aware that when some resources were not used anymore, they were dropped from the repository as well.

### OUDS Web and Bootstrap versions

Boosted and Bootstrap were using the exact same versions at the same time. Now we have to deal with different versions, and so, some Hugo params must be split:
* New `bootstrap_docs_version` parameter set to `5.3`
* New `bootstrap_current_version` parameter set to `5.3.3`

### System of URLs

This PR fixes the URLs issues and configuration and works when:
- [x] `npm run start`
- [x]  https://deploy-preview-2612--boosted.netlify.app/ 
- [x] `npm run docs`, then run `php -S localhost:8008` in `_site` directory
- [x] `npm run release`, `mv _site/ouds-web /tmp`, `git checkout gh-pages`, `rm -rf ouds-web/*`, `mv /tmp/ouds-web/* ouds-web/`, then run `php -S localhost:8008`

Instead of adding a `/ouds-web` everywhere in the code like tried in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2614/files, a `modify-site.sh` script does it only when we build the documentation targetting `gh-pages`.
It allows running the documentation locally and via Netlify as usual, and we'll only need to drop the usage of this script whenever OUDS Web will replace Boosted.

Please note that there are some edge cases that don't work for now until the new DNS is correctly set up, but should be automatically handled in a near future:
* Redirection like https://orange-opensource.github.io/Orange-Boosted-Bootstrap/ouds-web/docs have been configured to use the newly created DNS web.unified-design-system.orange.com that will redirect to https://orange-opensource.github.io/Orange-Boosted-Bootstrap/ouds-web
* CSS used by Storybook in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2612/files#diff-22bfa436a490448858048de4423952a0f8db74eb9667a997e5af2cf1d2c94a38 won't work until the new DNS is correctly set up

---

## After the merge

- [ ] Take into account https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2612#discussion_r1672188927
- [ ] Make `modify-docs.sh` script work on Windows too
- [ ] Generate patch versions (0.0.1, 0.0.2, etc.) with the real process to check that all download references and links work as expected (zip files, examples packages, npm packages, etc.)
- [ ] Fix `ReferenceError: boosted is not defined`, and handle the JS part correctly
- [ ] **We'll need to track everything that has been temporarily removed in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/2611 so that it can be reintroduced step by step for OUDS Web**

#### Storybook

The following URLs should work:
- [ ] Locally after a `npm run start`: https://localhost:9001/storybook
- [ ] In Netlify: https://deploy-preview-2612--boosted.netlify.app/storybook
- [ ] In `gh-pages`: https://orange-opensource.github.io/orange-Boosted-Bootstrap/ouds-web/storybook
- [ ] With the DNS: https://web.unified-design-system.orange.com/storybook


NOTICE.txt
julien-deramond, line 1: TODO: double-check remaining images at the end of the process

site/layouts/partials/docs-sidebar.html
```diff
@@ -34,7 +35,7 @@
           {{ if eq $group_slug "migrations" }}
             {{- $href = urls.JoinPath "/docs" $.Site.Params.docs_version $doc_slug "/" }}
           {{ end }}
-          <li><a href="{{ $href }}" class="bd-links-link d-inline-block{{ if $is_active }} active{{ end }}"{{ if $is_active }} aria-current="page"{{ end }}>{{ $doc.title }}</a></li>
+          <li><a href="{{ $href }}" class="bd-links-link d-inline-block{{ if $is_active }} active{{ end }}{{if $is_draft }} text-decoration-line-through{{ end }}"{{ if $is_active }} aria-current="page"{{ end }}>{{ $doc.title }}</a></li>
```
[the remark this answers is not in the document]
hannahiss: I think that `<del>` is not accurate, and `<s>`, why not... 
But a `visually-hidden` mention is a solution I often like because it always works 👍  So go for it!?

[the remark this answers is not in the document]
julien-deramond: Yep, let's start with a `visually-hidden` 👌

site/content/docs/0.0/examples/fake/index.html
```diff
@@ -0,0 +1,9 @@
+---
```
vprothais, line 1: **question(non-blocking):** Just for my knowledge, what is this file used for ? I don't see any "Fake" on the exampes page of the netlify published doc.

vprothais (approved):

.bundlewatch.config.json
```diff
@@ -10,7 +10,7 @@
     },
     {
       "path": "./dist/css/ouds-web-reboot.css",
-      "maxSize": "4.5 kB"
+      "maxSize": "4.75 kB"
```
louismaximepiton, line 13: Is that only because of the header ? That's unlucky

louismaximepiton (approved):
👌

site/content/docs/0.0/examples/fake/index.html
julien-deramond: In Boosted, you can [package and download the examples](https://boosted.orange.com/docs/5.3/examples/). The script in our `package.json` that handles this expects files in the examples directory, including at least one CSS file. To keep things simple, I created a temporary hidden example to avoid modifying the build/packaging process. This example is intentionally not displayed on the examples page.

Consider this a temporary hack, knowing that we will have at least the "Masonry" example in the future, so this fake example will be removed soon.

.bundlewatch.config.json
julien-deramond: Yes 🤦

site/content/docs/0.0/examples/fake/index.html
vprothais: ok thanks, I understand.

PR #2613, opened 2024-07-11
chore: drop IcoMoon images

julien-deramond:
### Description

This PR simply drops unused IcoMoon images and "icomoon" string from spell check as it's not mentioned anymore in our documentation.


louismaximepiton (approved):
👌

PR #2617, opened 2024-07-15
[OUDS] Docs(storybook): reintegrate Storybook run and build

julien-deramond:
### Description

This PR reintroduces Storybook configuration for running it and building it locally, via Netlify, and via GitHub Pages.

In order to get Netlify to build correctly the documentation, I had to change the way `modify-site.sh` is executed. Indeed, the `LC_ALL=C sh` env was working well on macOS but not on Windows (/cc @louismaximepiton), and not on a Unix env through Netlify. So, this .sh script has been transformed into a JS module (like the other build files).

The first commit reintroduces:
- The accordion documentation to create at least one automatic story
- The `docs.css` from Boosted documentation to avoid infinite loading caused by web.unified-design-system.orange.com not yet set up properly

This temp configuration allows verifying the following running contexts:
- [x] Locally: `npm run storybook`
- [x] Locally: `npm run storybook-build` + `cd _site/ouds-web/storybook ; php -S localhost:8008`
- [x] Netlify at https://deploy-preview-2617--boosted.netlify.app/storybook
- [x] GitHub Pages by copying `_site/ouds-web/storybook` temporarily locally in `gh-pages` branch

The second commit drops the temp modifications.


build/modify-site.mjs
```diff
@@ -0,0 +1,28 @@
...
+if (sh.test('-d', 'ouds-web')) {
+  sh.rm('-rf', 'ouds-web')
+}
+
+sh.mkdir('ouds-web')
+sh.mv('_site/*', 'ouds-web/')
+sh.mv('ouds-web', '_site/')
+
+for (const file of sh.find('_site').filter(filename => {
+  return filename.includes('.html') || filename.includes('.xml')
```
louismaximepiton, line 22: Why do we need this filter ?

package.json
```diff
@@ -96,7 +96,7 @@
     "watch-js-docs": "nodemon --watch site/assets/js/ --ext js --exec \"npm run js-lint\"",
     "storybook": "npm run storybook-generate && storybook dev -p 6006",
     "storybook-generate": "npm run docs && node stories/create-stories-from-doc.js",
-    "storybook-build": "npm run storybook-generate && storybook build -o ./_site/storybook",
+    "storybook-build": "npm run storybook-generate && storybook build -o ./_site/ouds-web/storybook",
     "storybook-build-netlify": "npm run storybook-generate && storybook build -o ./site/storybook"
```
louismaximepiton, line 100: Do we keep only one url for storybook on prod? I'd say no, but maybe you have something in mind?

louismaximepiton (changes requested):
It does work on my side with small adjustments

build/modify-site.mjs
julien-deramond: Because you can't `sed` a directory. The syntax was easier than checking that it's a file. At least the ones I've found.

package.json
julien-deramond: What do you mean by that?
For now, on Boosted, you only have boosted.orange.com/storybook (latest version) since it's only used by Zeroheight

louismaximepiton (approved):
👌

PR #2618, opened 2024-07-15
[OUDS] Fix(js): rename JS global value from 'ouds-web' to 'oudsWeb'

julien-deramond:
### Description

This PR renames the JS global value from 'ouds-web' to 'oudsWeb'.

In our documentation JavaScript files, we use the global value to access the OUDS Web API/SDK. The current value was 'ouds-web', which was not valid when called like this: `ouds-web.Tooltip`. This PR renames the global value to 'oudsWeb', which is a valid JavaScript variable name.

Moreover, this PR also updates the documentation to reflect the new global value by replacing `boosted` with `oudsWeb`.

Here are the checks that have been done:
- [x] going to the [Accordion page](https://deploy-preview-2618--boosted.netlify.app/docs/0.0/components/accordion/) and playing with the code copy button.
  - [x] ⚠️ the Accordion page content commit must be reverted before merging
- [x] rendering locally the JS visual tests
- [x] nothing to check for the ESM version that is not modified by this PR and does not contain "ouds-web" nor "boosted" strings


louismaximepiton (approved):
LGTM 🚀

Issue #2589, opened 2024-06-28
[OUDS] [TRACKING] Backlog to be functionally equivalent to Boosted


julien-deramond:
### Description

This tracking issue can be seen as a backlog of things to develop so that OUDS Web is functionally equivalent to Boosted in terms of features, documentation, packaging, and many other things.
Most of these elements have been removed temporarily when creating the branch for OUDS Web; they existed before in Boosted or Bootstrap, and will probably need to be reintegrated at some point. At least, if they don't, we must decide explicitly that they shouldn't be integrated.

### Ready to dev

- [ ] Make Algolia and search work #3299 
  - Might be an issue caused by `site/assets/js/search.js` with the `liveUrl` targeting `web.unified-design-system.orange.com`? 
- [ ] Determine what we do with Meteor define in `package.js`. Do we keep it? Do we need this `package.js` file?
- [ ] Take into account this comment: https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2647#pullrequestreview-2196537994

### Storybook

> [!WARNING]  
> Blocked by web.unified-design-system.orange.com not being properly set-up

- [ ] Error while rendering https://orange-opensource.github.io/Orange-Boosted-Bootstrap/ouds-web/storybook. Due to `.storybook/preview-head.html` configured to use `https://web.unified-design-system.orange.com/docs/0.0/assets/css/docs.css`. Should be fixed automatically.
- [x] Reintegrate `ods.web` parameter in `hugo.yml` with the new value

### Documentation

> [!NOTE]  
> As soon as `boosted-sprite.svg` must be reintegrated, mention it in the `NOTICE.txt`

- [x] Redirection of pages to double-check. (:warning: Blocked by web.unified-design-system.orange.com not being properly set-up)
- [ ] Add a page that explains how we are compatible with Bootstrap (maybe the README too). What version is compatible with what Bootstrap version. That the framework is compatible with Bootstrap. etc. 
- [ ] **Homepage**
  - [x] **(On hold: Question asked to the project, waiting for an answer)** Introduce a new OUDS logo replacing Boosted homemade logo? https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/980869e8a3338afa5ed8986c40b1b7dfdb916f4d/site/assets/scss/_boosted.scss#L20-L42
  - [ ] Jumbotron (compare Boosted / Bootstrap)
[...]

PR #2635, opened 2024-07-22
[OUDS] feat(css): Drop `form-star-rating()` mixin

julien-deramond:
### Related issues

Listed in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/2589

### Description

`form-star-rating()` mixin has been deprecated in Boosted v5.3.2. Since this library starts from scratch, there is no need to start with it. However, this must be mentioned in the migrating guide from Boosted.

Plus, the other elements that are deprecated and that use the `deprecate()` mixin are all from Bootstrap (because Boosted didn't deprecate really elements back then and deleted them directly): https://github.com/search?q=repo%3AOrange-OpenSource%2FOrange-Boosted-Bootstrap+%22%40include+deprecate%28%22&type=code
So, let's consider that `deprecate()` is a mixin dedicated to Bootstrap elements, as it will be easier to maintain.
If we need a deprecation mixin for OUDS Web elements, we will create a new one.

In order to make the Bootstrap deprecation mixin clearer, "Bootstrap" string has been added to the sentences.

### Types of change

### Live previews

- https://deploy-preview-2635--boosted.netlify.app/docs/0.0/migration-from-boosted/


louismaximepiton (approved):
LGTM 🚀

PR #2638, opened 2024-07-22
[OUDS] Css: drop deprecated Boosted variables and change deprecated comments

julien-deramond:
### Related issues

Listed in https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/2589

### Description

This PR handles the "deprecate" comments in JavaScript and Sass files.

The rules are the following:
- If the deprecation comment comes from Bootstrap, keep it and include the word "Bootstrap" in it
- If the deprecation comment comes from Boosted and relies on Bootstrap values, change it to OUDS Web, and keep the Boosted previous values for Boosted projects migrating to OUDS Web
- If the deprecation comment comes from Boosted only, the corresponding variable(s) can be removed from the framework

All removed Sass variables are listed in the Boosted migration guide.

### Types of change

### DIFF
--- a/site/content/docs/0.0/migration-from-boosted.md
+++ b/site/content/docs/0.0/migration-from-boosted.md
@@ -65,12 +65,12 @@ Technically, it means that you can get rid of the following things:
       <li><code>$btn-close-white-color</code></li>
       <li><code>$btn-close-white-disabled-color</code></li>
       <li><code>$btn-close-white-hover-color</code></li>
+      <li><code>$footer-nav-link-font-weight</code></li>
       <li><code>$form-star-focus-box-shadow</code></li>
       <li><code>$form-star-focus-color</code></li>
       <li><code>$form-star-focus-color-dark</code></li>
       <li><code>$form-star-focus-outline</code></li>
       <li><code>$form-star-focus-outline-dark</code></li>
-      <li><code>$footer-nav-link-font-weight</code></li>
       <li><code>$orange-filter</code></li>
       <li><code>$outline-offset</code></li>
       <li><code>$outline-width</code></li>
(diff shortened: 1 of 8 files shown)


site/content/docs/0.0/migration-from-boosted.md
```diff
@@ -56,25 +56,38 @@ Technically, it means that you can get rid of the following things:
...
-      <li><code>$btn-close-white-active-border-color</code></li>
+      <li><code>$btn-close-white-color</code></li>
       <li><code>$btn-close-white-disabled-color</code></li>
+      <li><code>$btn-close-white-hover-color</code></li>
+      <li><code>$form-star-focus-box-shadow</code></li>
+      <li><code>$form-star-focus-color</code></li>
+      <li><code>$form-star-focus-color-dark</code></li>
+      <li><code>$form-star-focus-outline</code></li>
+      <li><code>$form-star-focus-outline-dark</code></li>
+      <li><code>$footer-nav-link-font-weight</code></li>
```
louismaximepiton, line 73: Footer should come before form.

louismaximepiton (changes requested):
I think it lacks a bit of consistency between all the deprecating messages. I think it would be great to have the rules for those messages stored in the description and not only the rules for search & replace. For instance, we have `deprecated in Bootstrap *` but can we use it in OUDS since we have deprecation messages like `deprecated in OUDS`? Maybe a clear message like: `Not to be used in OUDS because it is deprecated in Bootstrap v*` or something similar everywhere would be safer. I also think that we could write the rules and keep the rewording when we will reintroduce things little by little.

One small change is needed and we could launch the PR imo.

site/content/docs/0.0/migration-from-boosted.md
julien-deramond: Good catch, fixed at https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/2638/commits/e5935cb42bdb282e9b2a664391b37ac632c20296

louismaximepiton (approved):

PR #2647, opened 2024-07-24
fix(docs): broken links to /blob/vx.y.z-ouds-web files

julien-deramond:
### Description

This PR fixes some broken links due to the fact that our tags have a "-ouds-web" suffix.

### Types of change

### Live previews


louismaximepiton (approved):
Good catch 👌 Do you think they well be removed somehow once the OUDS Web lib will be the main one ? If yes, maybe we should add a parameter to handle this rather than changing the link directly.

julien-deramond:
> Do you think they well be removed somehow once the OUDS Web lib will be the main one ?

The only moment when we would be able to remove the suffix, is when the major version of OUDS Web will be way over 6/7/8 to not block Boosted tags (that will follow Bootstrap tags).

PR #2678, opened 2024-08-05
docs(a11y): drop `focusable='false'` for SVGs

### LINKED ISSUE #2677: Get rid of `focusable="false"` for decorative `<svg>`s
### Description

Decorative SVGs are most of the time used like this: `<svg aria-hidden="true" focusable="false">` in Boosted.
`focusable="false"` was there for a better support for Internet Explorer.
IE being not supported anymore, can't we get rid of `focusable="false"` in those cases, knowing that it's already the case for Bootstrap (see https://github.com/twbs/bootstrap/commit/ac831aa4df8e817883c1510f74604ea67e26f179#diff-17bd0a43dcea923cbb44c877ce1462c41b5624ccafa66f29cb6d0946707f096c)

julien-deramond:
### Related issues

Closes #2677 

### Description

This PR drops `focusable="false"` for SVGs since we don't support Internet Explorer anymore.

### Types of change

- Accessibility enhancement (non-breaking change)

### Live previews

- <https://deploy-preview-2678--boosted.netlify.app/>

### DIFF
--- a/site/content/docs/5.3/components/badge.md
+++ b/site/content/docs/5.3/components/badge.md
@@ -67,7 +67,7 @@ Please refer to our [Boosted Navbars examples]({{< docsref "/examples/navbars" >
 
 {{< example >}}
 <a href="#" class="position-relative">
-  <svg width="2rem" height="2rem" fill="currentColor" aria-hidden="true" focusable="false" class="overflow-visible">
+  <svg width="2rem" height="2rem" fill="currentColor" aria-hidden="true" class="overflow-visible">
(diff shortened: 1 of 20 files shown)


vprothais (approved):

louismaximepiton (approved):
Since we don't support IE and that Edge implements `tabindex` since quite a long time, I'd remove all the `focusable` occurrences and if really needed we could add `tabindex="0"` but I don't think it's necessary. 
The remaining looks good to me.

PR #2683, opened 2024-08-07
Docs: fix sidebar spacing

louismaximepiton:
### Description

Fix the spacing under the migration link for a better design.

https://deploy-preview-2683--boosted.netlify.app/docs


julien-deramond:
I'm not sure to catch exactly how it's better. Could you please share some before/after screenshots?

louismaximepiton:
Yep sure, it's not much and maybe we can close the PR but here was the result before:
image

And here is the result with the PR:
image

Migration link is more centered right now

julien-deramond:
Thanks, the rendering is slightly different on macOS and is not centered anymore. As the rendering discrepancies are probably linked to the scrollbar rendering, let's close this PR.

Issue #2667, opened 2024-08-01
Accessibility audit


julien-deramond:
> [!NOTE]   
>Total issues: 42
> - 🟢 Done => 23 issues
> - 🟡 Ongoing (approved, almost ok) => 0 issue
> - 🟠 Ongoing (not reviewed nor approved yet) => 3 issues
> - 🔴 Not started (questions still pending) => 0 issue
> - 🔴🔴 Won't be done (because we can't fix it) => 16 issues (covering 5 topics)

### Description

An accessibility audit has been done on the v5.3.3 documentation Boosted website, the 30th of July 2024. Here are the results.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

These results are not exhaustive. They are given by way of example, to illustrate the anomalies. It is the responsibility of the project team to apply these corrective measures systematically to all similar cases.

#### Pages

- [Homepage](https://boosted.orange.com)
- [Docs](https://boosted.orange.com/docs)
- [Examples](https://boosted.orange.com/docs/5.3/examples)
- [Design guidelines](https://boosted.orange.com/docs/5.3/guidelines)
- [Overview](https://boosted.orange.com/docs/5.3/about/overview)
- [Brand](https://boosted.orange.com/docs/5.3/about/brand)
- [Translations](https://boosted.orange.com/docs/5.3/about/translations)
- [Cookies](https://boosted.orange.com/docs/5.3/about/cookies)

##### 🟢 [issue-0-1](https://la-va11ydette.orange.com/?list=wcag-web&lang=en#headingtestID-017)
- **Test**: Is the page title appropriate?
- **Recommendation**: https://a11y-guidelines.orange.com/fr/web/developper/contenu-textuel/#donner-un-titre-aux-pages
- **Summary**: Irrelevant page title
- **Description**: Each page should have a unique title, making it easy to understand its role.
The role "home page" should be integrated into the page title.
- **Solution**: Fill in the `<title>` element with keywords to help understand the page's role. Information can be ordered from the most specific to the least specific, or vice versa (no rules).
- **Technical solution**: `<title>Home - Orange Boosted....</title>`
- **Priority**: P1
- **PR**: #2698 

##### 🟢 [isssue-0-2](https://la-va11ydette.orange.com/?list=wcag-web&lang=en#headingtestID-013)
- **Test**: Do the images have appropriate text alternatives? (decorative `<img>` should have empty `alt`)
[...]

PR #2695, opened 2024-08-20
Fix a11y audit 7-1

louismaximepiton:
### Related issues

#2667 

### Description

- Rearrange the tarteaucitron modal structure.

### Motivation & Context

Read the related issue. The switch was too far away from the label.

### Types of change

### Live previews

- <https://deploy-preview-2696--boosted.netlify.app/docs/about/cookies>

### DIFF
--- a/site/assets/js/tac.js
+++ b/site/assets/js/tac.js
@@ -11,12 +11,12 @@
     const tac = document.getElementById('tarteaucitron')
 
     tac.querySelectorAll('.tarteaucitronAllow').forEach(button => {
-      button.classList.add('btn', 'btn-sm', 'btn-success', 'my-2', 'my-lg-0')
+      button.classList.add('btn', 'btn-sm', 'btn-success')
       button.innerHTML = tarteaucitron.lang.allowAll
     })
 
     tac.querySelectorAll('.tarteaucitronDeny').forEach(button => {
-      button.classList.add('btn', 'btn-sm', 'btn-danger', 'my-2', 'my-lg-0', 'ms-lg-2')
+      button.classList.add('btn', 'btn-sm', 'btn-danger', 'mt-1', 'mt-md-0', 'ms-md-3')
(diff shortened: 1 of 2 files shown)


hannahiss (approved):
LGTM as you say 😄

julien-deramond:
It was not seen in the accessibility audit, but with the new tarteaucitron version, there's a "Save" button. Don't you think there's an issue with the focus order, where the "Save" button is focused after the footer link? Can be tackled in another PR.
/cc @hannahiss

hannahiss:
> It was not seen in the accessibility audit, but with the new tarteaucitron version, there's a "Save" button. Don't you think there's an issue with the focus order, where the "Save" button is focused after the footer link? Can be tackled in another PR. /cc @hannahiss

Yes, the order is strange... I vote for another PR 😉

julien-deramond:
> > It was not seen in the accessibility audit, but with the new tarteaucitron version, there's a "Save" button. Don't you think there's an issue with the focus order, where the "Save" button is focused after the footer link? Can be tackled in another PR. /cc @hannahiss
> 
> Yes, the order is strange... I vote for another PR 😉

Added to the audit issue as a bonus task since it can be a problem during the next one.

julien-deramond (approved):

louismaximepiton:
> It was not seen in the accessibility audit, but with the new tarteaucitron version, there's a "Save" button. Don't you think there's an issue with the focus order, where the "Save" button is focused after the footer link? Can be tackled in another PR. /cc @hannahiss

It seems like moving the link after the Save button breaks, because of the tarteaucitron focus trap. If the link is moved after the button manually, you can't focus it anymore (even doing it manually). I'd say it's not easily feasible.

julien-deramond:
> It seems like moving the link after the Save button breaks, because of the tarteaucitron focus trap. If the link is moved after the button manually, you can't focus it anymore (even doing it manually). I'd say it's not easily feasible.

In your opinion, we abandon this modification?

PR #2699, opened 2024-08-20
Fix a11y audit 0-6: focus order for the copy button on code example

louismaximepiton:
### Related issues

#2667 0-6

### Description

- Fix the focus order on code examples.
- Could probably be done in a better way on Bootstrap directly using CSS only.

### Motivation & Context

Read the related issue. The focus was on the copy button before the code.

### Types of change

### Live previews

- <https://deploy-preview-2699--boosted.netlify.app/>
- <https://deploy-preview-2699--boosted.netlify.app/docsref>


hannahiss (approved):
Great focus order thx!

site/assets/js/partials/code-examples.js
```diff
@@ -20,9 +20,9 @@ export default () => {
 
   // Boosted mod: removed unused svg height, width and fill="currentColor"
   const btnHtml = [
-    '<div class="bd-code-snippet">',
-    '  <div class="bd-clipboard">',
-    '    <button type="button" class="btn-clipboard">',
+    '<div class="bd-code-snippet position-relative">',
+    '  <div class="bd-clipboard position-absolute top-0 end-0">',
+    '    <button type="button" class="btn-clipboard position-absolute end-0">',
```
vprothais, line 25: Don't we need a visually hidden label for this button ? It only contains an SVG without any text.

vprothais: There is a aria-label on the svg, I missed it at first :)

julien-deramond:
In https://deploy-preview-2699--boosted.netlify.app/docs/5.3/docsref/, only the first example is handled this way. It's not possible to modify the focus order for the other ones too?

louismaximepiton:
I understand your point here, but the visual rendering is different between the examples and I think it could be hard to make the same focus order for each example you mention.

louismaximepiton:
Changed the focus order for each example, tell me if you think it's not too weird to manipulate and in the code as well.

julien-deramond (approved):
The new focus order is great! I haven't particularly checked the modifications, as they are only for the documentation and are gathered in some specific recognizable files. Just added some Boosted mod comments for maintenance and sync with upstream project.

PR #2706, opened 2024-08-23
Fix a11y audit 2-4: Adding title to links to avoid same labels

hannahiss:

### Related issues

#2667 issue 2-4

### Description

Adding title to links to avoid same labels in different links

### Motivation & Context

Accessibility audit from EASE

### Types of change

### Live previews

- <https://deploy-preview-2706--boosted.netlify.app/>


louismaximepiton (approved):

site/layouts/partials/footer.html
```diff
@@ -78,7 +78,7 @@ <h3 class="accordion-header footer-heading" id="headingCommunity">
...
             </p>
             <p class="mb-lg-1">
-              Orange modified code licensed <a href="{{ .Site.Params.repo }}/blob/main/LICENSE" target="_blank" rel="license noopener">MIT</a>-&nbsp;just like <a href="https://github.com/twbs/bootstrap/blob/main/LICENSE" target="_blank" rel="license noopener">Bootstrap</a>, docs <a href="https://creativecommons.org/licenses/by/3.0/" target="_blank" rel="license noopener">CC BY 3.0</a>.
+              Orange modified code licensed <a href="{{ .Site.Params.repo }}/blob/main/LICENSE" target="_blank" rel="license noopener">MIT</a>-&nbsp;just like <a href="https://github.com/twbs/bootstrap/blob/main/LICENSE" target="_blank" rel="license noopener" title="Bootstrap's license" aria-label="Bootstrap's license">Bootstrap</a>, docs <a href="https://creativecommons.org/licenses/by/3.0/" target="_blank" rel="license noopener">CC BY 3.0</a>.
```
julien-deramond, line 81: Noob question, but is the `aria-label` not enough here?

hannahiss: Well, the answer is again a debate of experts about what is better not... 
Of course `aria-label` is ok and even better supported than `title`, but experts find that the native "tooltip" generated by the `title` is a good indication for sighted users, when the accessible name is different from the link's text itself (which is the case here)
See for example: https://www.accede-web.com/en/guidelines/html-css/links-buttons/complete-non-explicit-links-and-buttons-using-hidden-text-aria-label-or-title/

julien-deramond: TBH, I find the `title` unnecessary in this case, as it's fairly obvious for sighted users. That said, I don't want to get into time-consuming discussions overt this neither 😄, so let's go ahead and merge it as is.

julien-deramond (approved):

PR #2709, opened 2024-08-23
Fix audit a11y: Change wrong images labels

hannahiss:

### Related issues

#2667 

### Description

Change labels:
- for Orange logos
- for tooltip example with an SVG

### Types of change

### Live previews

- <https://deploy-preview-2709--boosted.netlify.app/docs/about/brand/>
- <https://deploy-preview-2709--boosted.netlify.app/docs/components/tooltips/>


site/content/docs/5.3/about/brand.md
```diff
@@ -20,15 +20,15 @@ Have a need for Boosted's brand resources? Great! We have only a few guidelines
   <div class="col">
     <div class="ratio ratio-1x1">
     <figure class="d-flex bg-black">
-      <img src="/docs/{{< param docs_version >}}/assets/brand/orange-logo.svg" class="figure-img img-fluid m-auto" width="220" height="220" role="img" alt="Orange" loading="lazy">
+      <img src="/docs/{{< param docs_version >}}/assets/brand/orange-logo.svg" class="figure-img img-fluid m-auto" width="220" height="220" role="img" alt="Orange full logo" loading="lazy">
```
louismaximepiton, line 23:
```suggestion
      <img src="/docs/{{< param docs_version >}}/assets/brand/orange-logo.svg" class="figure-img img-fluid m-auto" width="220" height="220" role="img" alt="Orange master logo" loading="lazy">
```
I think I'd prefer to ear this rather than full.

louismaximepiton (commented):
The question is still running for me, but the logo is mainly used in the header and often used as a part of the brand name. For instance I'm thinking about Orange Cyberdefense, Orange Foundation, etc... Which title would then say `Orange master logo Cyberdefense`. My opinion is balancing between:
- We are a documentation so everything isn't perfect but we ease our UX.
- If people want to use the logo, they need to adapt the label and so we can have whatever in here.

hannahiss:
> The question is still running for me, but the logo is mainly used in the header and often used as a part of the brand name. For instance I'm thinking about Orange Cyberdefense, Orange Foundation, etc... Which title would then say `Orange master logo Cyberdefense`. My opinion is balancing between:
> 
> * We are a documentation so everything isn't perfect but we ease our UX.
> * If people want to use the logo, they need to adapt the label and so we can have whatever in here.

We didn't talk about this: of course, in the context of a website, if the logo is in a links that leads to home page, as it often does, the label shouldn't be "XXX logo" but something like "XXX home". For this, I think that developers can be inspired by Boosted code itself, don't you think? In this example, we don't have any context, so it would be weird to use such a label...

louismaximepiton (approved):
LGTM

site/content/docs/5.3/components/tooltips.md
```diff
@@ -130,7 +130,7 @@ With an SVG:
 <!--Boosted mod: added an aria-label="Default tooltip" because the title is automatically removed by JS-->
 
 <div class="bd-example tooltip-demo">
-  <a href="#" class="d-inline-block" data-bs-toggle="tooltip" data-bs-title="Default tooltip" aria-label="Default tooltip">
+  <a href="#" class="d-inline-block" data-bs-toggle="tooltip" data-bs-title="Default tooltip" aria-label="Focus or hover to open default tooltip">
```
julien-deramond, line 133: Maybe we can remove this modification from this PR as it's not really the same modification. And that will come back to Boosted automatically as soon as https://github.com/twbs/bootstrap/pull/40756/files#r1750127676 is taken into account?

hannahiss: I moved the change to PR https://github.com/twbs/bootstrap/pull/40808

julien-deramond (approved):

PR #2713, opened 2024-09-02
[OUDS] Docs review after first OUDS version (including opacity tokens)

hannahiss:
_Note: Please transform `- [ ]` into `- (NA)` in the description when things are not applicable_

### Related issues

#2589 

### Description

Docs review after first PRs from 2589 issue: suggestions for improving documentation or reporting problems, review opacity tokens.
Changes:
- in <https://deploy-preview-2713--boosted.netlify.app/docs/getting-started/accessibility/> the link name should be WCAG 2.2 and not 2.1
- in <https://deploy-preview-2713--boosted.netlify.app/docs/getting-started/contents/> and <https://deploy-preview-2713--boosted.netlify.app/docs/customize/custom-libraries/>: additional information in the formulation
- in <https://deploy-preview-2713--boosted.netlify.app/docs/customize/options/>: is `$enable-cssgrid` mentionned in <https://deploy-preview-2713--boosted.netlify.app/docs/layout/css-grid/> not missing? (from Bootstrap)
- in <https://deploy-preview-2713--boosted.netlify.app/docs/utilities/api/>: do we need to keep RTL option and paragraph about RTL since RTL is not yet implemented/documented?
- in <https://deploy-preview-2713--boosted.netlify.app/docs/utilities/display/>: removed the word "semantic" as it may be confusing for the reader
- in <https://deploy-preview-2713--boosted.netlify.app/docs/utilities/flex/#align-content>, "heads up" is not in a callout as usual (from Bootstrap)
- in <https://deploy-preview-2713--boosted.netlify.app/docs/about/license/>: there were two different links with the same accessible name which is an accessibility issue
- in <https://deploy-preview-2713--boosted.netlify.app/docs/migration/#pre-compiled-versions> I proposed another wording but I'm not totally satisfied either (see question below)

Other questions:
- in <https://deploy-preview-2713--boosted.netlify.app/docs/extend/approach/>: I can't see why the paragraph Utilities has not been uncommented?
[...]

PR #2723, opened 2024-09-09
Docs: fix different links having same accessible name

hannahiss:

### Related issues

This issue has been seen while reading OUDS Web documentation (PR #2713 )

### Description

Change the labels of links to make them discriminating.

### Motivation & Context

The links to Boosted and Booststrap license have the same accessible name ("in the project repository") but not the same target, which is an accessibility issue. 

### Types of change

### Live previews

- <https://deploy-preview-2723--boosted.netlify.app/docs/about/license/>


louismaximepiton (approved):
I'm fine with this, just wondering if we should include `full` word inside the Boosted license. However, it already works like a charm.

julien-deramond (approved):

PR #2726, opened 2024-09-10
Fix a11y audit: enlarge touch targets of right side menu "On this page"

hannahiss:

### Related issues

#2667 issue 1-8

### Description

Increase vertical padding to make the clicking area at least 24px high

### Types of change

### Live previews

- <https://deploy-preview-2726--boosted.netlify.app/docs>


site/assets/scss/_toc.scss
```diff
@@ -26,7 +26,7 @@
 
     a {
       display: block;
-      padding: .125rem 0 .125rem .75rem;
+      padding: .313rem 0 .25rem .313rem;
```
julien-deramond, line 29: If we modify values from Bootstrap, we should add the "Boosted mod" comment.

```suggestion
      padding: .313rem 0 .25rem .313rem; // Boosted mod
```

julien-deramond: `.313rem` seems weird to me as a value as it's `5.008px`

julien-deramond (changes requested):
It's no more left-aligned with "On this page" correctly with these values (tablet and desktop breakpoints, mobile seems OK):

Screenshot 2024-09-11 at 08 35 58

Reference on the _main_ branch:

Screenshot 2024-09-11 at 08 36 57

site/assets/scss/_toc.scss
hannahiss: Sorry I think I was a bit tired last evening 😓

julien-deramond (approved):

PR #2760, opened 2024-10-14
fix(forms): avoid .is-required element to be alone on a new line

### LINKED ISSUE #2758: The asterisk positioned on a label of a required field returns to the line
### Prerequisites

- [x] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [x] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems

### Describe the issue

In some of our GUIs, the width of a field label may correspond exactly to the width of the label text. In this case, if the field is mandatory, the asterisk added by the required class is on the line, giving a negative visual effect.

### Reduced test cases

Here is an example : https://codepen.io/jacques-lebourgeois/pen/vYogBLW

with the screenshot :  
Image

Windows

Chrome

5.3

julien-deramond:
### Related issues

Closes #2758

### Description

This PR fixes the issue described at #2758 where the asterisk showing that a form control is required can be alone on a new line when the label length has exactly the same length as the corresponding input.

I tried different non-working solutions based on `word-wrap` and non breakable space without success. This PR relies on @jacques-lebourgeois's suggestion.

The only drawback I've found of this fix is that the asterisk, intead of going to the new line, stays on the same line but is displayed in the safe area; outside of the bounding box that we should use. However, I'd consider that as a minor issue.

See:

Screenshot 2024-10-14 at 08 08 06

### Types of change

### Live previews

- https://deploy-preview-2760--boosted.netlify.app/docs/5.3/forms/overview/#required-field

### DIFF
--- a/scss/forms/_labels.scss
+++ b/scss/forms/_labels.scss
@@ -17,6 +17,7 @@
 }
 
 .is-required::after {
+  position: absolute;
   margin-left: $form-label-required-margin-left;
   color: $form-label-required-color;
   content: "*";


louismaximepiton (approved):
I tried few things as well, but I also think that it's the best option in here. Well done 🎉

PR #2854, opened 2025-01-30
chore(merge main):patched commit → 5708adc

louismaximepiton:
- [x] https://github.com/twbs/bootstrap/commit/a5d430d526dcb330b3e0f12aa1321cf948d726f1
- [x] https://github.com/twbs/bootstrap/commit/541acb4aac6a237ccf36755b6435c4098cc8cc8d
- [x] https://github.com/twbs/bootstrap/commit/5210c9b24c44f8230c3ca4813d30211289f27b2e
- [x] https://github.com/twbs/bootstrap/commit/5708adc816b9a6b07fb2b4f07c3b562935bd0a54

Introduces a deprecation warning from stylelint. I don't know if we manage this in here or not.


scss/mixins/_banner.scss
```diff
@@ -1,12 +1,12 @@
 @mixin bsBanner($file) {
   /*!
    * Boosted #{$file} v5.3.3 (https://boosted.orange.com/)
-   * Copyright 2014-2024 The Boosted Authors
-   * Copyright 2014-2024 Orange SA
+   * Copyright 2014-2025 The Boosted Authors
+   * Copyright 2014-2025 Orange SA
```
vprothais, line 5: We have 2015 as the starting date everywhere except here. Not really a problem but isn't it a mistake ?

vprothais (approved):

scss/mixins/_banner.scss
louismaximepiton: Tbh, I don't know why is that, I'd rather keep it as-is 🤷

vprothais: yep I already approved because I don't find any reason to change that, it's just a bit unexpected :)

PR #2904, opened 2025-03-20
docs: [OUDS] Small adjustments on buttons doc

hannahiss:

### Description

- add more margin around buttons in doc to be compliant with design documentation, in particular between buttons side to side, only when needed thanks to class `bd-btn-example`
- add missing dot at the end of warning callout like in "Links"
- remove useless `fill="currentColor"` on `<svg>`s since it is already set into the sprite file, and mention it in the doc
- remove useless `width="1rem" height="1rem"` on `<svg>`s since it is already done by styling, and remove it in the doc
- remove CSS/Sass variables at the end
- fix "JavaScript" typo in the loading button example

### Motivation & Context

Small fixes to do regarding what has been done on other documentation pages, and forgotten elements. 

### Types of change

### Live previews

- <https://deploy-preview-2904--boosted.netlify.app/components/buttons/>
- <https://deploy-preview-2904--boosted.netlify.app/examples/loading-buttons/>


site/assets/scss/_component-examples.scss
```diff
@@ -81,9 +81,9 @@
   }
 
   // Buttons
-  > .btn,
-  > .btn-group {
-    margin: .25rem .125rem;
+  .btn,
+  .btn-group {
```
vprothais, line 85: I find this change dangerous, because we could have component containing button, like inline alert, and we don't want to change the margin in this case. The direct child selector should be kept I think (but it won't work for button in div or other container :/

Maybe the most reliable way to do it is to add a parent class for this style and to add this class in the button examples ?

vprothais (changes requested):

site/assets/scss/_component-examples.scss
hannahiss: I added a class `bd-btn-example`

vprothais (approved):

PR #2976, opened 2025-04-15
Docs: chore(release): v5.3.5

louismaximepiton:


storybook/assets/Checkbox.stories-B4nsyOu2.js
vprothais, line 1: This file is seen as new and I don't see the removal of the previous version in the diff. I think we should have the removal of the previous Checkbox file ?

vprothais (approved):

PR #3035, opened 2025-06-25
feat(example): New Boosted compatibility example page

vprothais:
This example page is added to see what Boosted component looks like in OUDS when they are not developped yet.

The result is not so smooth :)

### Live previews

- <https://deploy-preview-3035--boosted.netlify.app/docs/0.4/examples/boosted-compat/>


vprothais:
Was just used for a test, closing this PR.

PR #3062, opened 2025-07-09
feat(component): Add Tag component

hannahiss:

### Related issues

#2845

### Types of change

- Breaking change (fix or feature that would change existing functionality)

### Live previews


site/src/content/docs/components/buttons.mdx
```diff
@@ -77,16 +77,14 @@ The recommended way of using an icon in a button is through an [SVG sprite file]
 When icons are purely decorative, they should be hidden from assistive technologies using `aria-hidden="true"`, as we’ve done in our text and icon examples. For icon only, we chose to keep the icon hidden from assistive technologies and add a visually hidden label. See [icons accessibility]([[docsref:/extend/icons#icons-accessibility]]) for more info.
 </Callout>
 
-### Text and icon
+If really needed, you can use a font icon associated to the `.icon` class to set correct parameters for the `font-size` and `line-height`.
 
-The recommended way of using an icon in a button is an [embedded SVG]([[docsref:/extend/icons]]). You need to fill it using `currentColor` to respect button color scheme (which can be done inside the SVG sprite file).
+### Text and icon
```
louismaximepiton, line 91: Are the changes inside this file meant to be inside this PR ?

site/src/pages/index.astro
```diff
@@ -11,7 +11,7 @@ import { getConfig } from '@libs/config'
   <MastHead />
   <div class="container-fluid container-max-width pt-scaled-md pb-scaled-3xl bd-masthead-followup">
     <p class="mb-scaled-md d-inline-flex flex-column flex-md-row align-items-start align-items-md-center">
-      <a href={`/docs/${getConfig().docs_version}/migration`} class="tag me-md bd-tag">New in v0.5</a>
+      <a href={`/docs/${getConfig().docs_version}/migration`} class="tag bd-tag me-md">New in v0.5</a>
```
louismaximepiton: We probably need to ask a design review on this.

scss/_variables.scss
```diff
@@ -2244,44 +2244,6 @@ $footer-gap:                              $spacer * .75 !default;
 $footer-gap-xl:                           $spacer * 1.7 !default;
 // scss-docs-end footer
 
-
```
MaxLardenois, line 2247: No OUDS mod on the deletion?

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+Tag indicator can be used near some text (to show status or key details), or in a list (to indicate or filter by categories or keywords).
+
+<Callout type="info">
+  For accessibility and semantics reasons, tags, like any other informative elements, must be enclosed in an ancestor or parent semantic tag. In the following examples, tag are used either in `<li>`, `<h2>` or `<p>`: you can use any other semantic tag depending on the usage context.
+</Callout>
+
+OUDS Web has a basic `.tag` class that sets up default styles.
+
+<Example code={`<ul class="list-unstyled d-flex column-gap-xs">
```
MaxLardenois: I'm wondering if we should display all the examples at once like that, maybe it is acceptable for tags? idk

scss/_tag.scss
```diff
@@ -0,0 +1,189 @@
...
+  align-items: center;
+  justify-content: center;
+  min-width: var(--#{$prefix}tag-min-width);
+  min-height: var(--#{$prefix}tag-min-height);
+  padding: subtract(var(--#{$prefix}tag-padding-y), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-end), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-y), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-start), var(--#{$prefix}tag-border-width));
+  margin: 0;
+  font-weight: 700;
+  color: var(--#{$prefix}tag-color);
+  white-space: nowrap;
+  vertical-align: text-bottom; // TODO alignment to check with text in <p>
```
MaxLardenois: Looks like it's not aligned in the first example?

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+    <li><span class="tag">5G Ready</span></li>
+    <li><span class="tag">4G</span></li>
+  </ul>
+  <h2>Title <span class="tag">Coming soon</span></h2>
+  <p><span class="tag">New</span> Some text</p>`
+  } />
+
+### Color variants
+
+OUDS Web [color and background utilities]([[docsref:/helpers/color-background]]) can be applied to display tag colored variants, either emphasized or muted. Each color is serving its own semantic purpose. Please follow the [Tag design specifications](https://unified-design-system.orange.com/) to choose the right badge for the right category. {/* TODO: really? OR thit's just colors? if semantic, set the right url for design documentation, otherwise remove this part */}
```
MaxLardenois:
```suggestion
OUDS Web [color and background utilities]([[docsref:/helpers/color-background]]) can be applied to display colored tags variants, either emphasized or muted. Each color has its own semantic purpose. Please follow the [tag design specifications](https://unified-design-system.orange.com/) to choose the right badge for the right category. {/* TODO: really? OR thit's just colors? if semantic, set the right url for design documentation, otherwise remove this part */}
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+    <li><span class="tag text-bg-status-accent-muted">Accent</span></li>
+    <li><span class="tag text-bg-status-positive-muted">Positive</span></li>
+    <li><span class="tag text-bg-status-info-muted">Info</span></li>
+    <li><span class="tag text-bg-status-warning-muted">Warning</span></li>
+    <li><span class="tag text-bg-status-negative-muted">Negative</span></li>
+  </ul>`} />
+
+### Squared
+
+To display tags with square corners, add the `.rounded-none` utility class to the tag.
```
MaxLardenois:
```suggestion
To display tags with squared corners, add the `.rounded-none` utility class to the tag.
```
Or 
```suggestion
To display tags without rounded corners, add the `.rounded-none` utility class to the tag.
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+    </span>
+    Negative
+  </span>`} />
+
+#### Icon
+
+<Callout type="info">
+  The recommended way of using an icon in a tag is through an [SVG sprite file]([[docsref:/extend/icons#svg-sprite]]).
+
+  Since icons in tags are always next to some text, they are purely decorative and should be hidden from assistive technologies using `aria-hidden="true"`, as demonstrated in our tag examples.
```
MaxLardenois:
```suggestion
  Since icons in tags are always next to some text, they are purely decorative and should be hidden from assistive technologies using `aria-hidden="true"`, as demonstrated in our examples.
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+<Example showPreview={false} code={`<span class="tag">
+    <span class="tag-asset-container">
+      <svg aria-hidden="true">
+        <use xlink:href="/docs/${getConfig().docs_version}/assets/img/ouds-web-sprite.svg#heart-recommend" />
+      </svg>
+    </span>
+    Neutral
+  </span>`} />
+
+If really needed, you can use a font icon associated to the `.icon` class to set correct parameters for the `line-height`.
```
MaxLardenois:
```suggestion
If absolutely necessary, you can use a font icon associated to the `.icon` class to set correct parameters for the `line-height`.
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+        Settings
+      </span>
+    </li>
+  </ul>`} />
+
+#### Loader
+
+{/* TODO : do we have use cases for loader? Do we need visually hidden label? */}
+
+To display a loader, just add the provided svg and a `.visually-hidden` text aside inside the `.tag-asset-container`.
```
MaxLardenois:
```suggestion
To display a loader, just add the following SVG and a `.visually-hidden` text inside the `.tag-asset-container`.
```
Did you mean the `aside` element?

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+    </li>
+  </ul>`} />
+
+## Tag input
+
+<Callout type="warning">
+  Please note that all the variants above shouldn't be applied on this version of the component.
+</Callout>
+
+Standard tags are used in an informative way. Tag inputs are interactive and can take all the usual statuses: enabled, hover, pressed, disabled or focus. The Tag input should be used along with a `<button>`.
```
MaxLardenois:
```suggestion
Standard tags are used in an informative way. Tag inputs are interactive and can take all the usual states: enabled, hover, pressed, disabled or focussed. The tag input should be used along with a `<button>`.
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+They aren't auto removed at the moment, so you need to implement this behavior by yourself.
```
MaxLardenois:
```suggestion
They aren't removed on click at the moment, so you need to implement this behavior by yourself.
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+    <li><button type="button" class="tag tag-input">Label</button></li>
+    <li><button type="button" class="tag tag-input">Label</button></li>
+    <li><button type="button" class="tag tag-input">Label</button></li>
+    <li><button type="button" class="tag tag-input">Label</button></li>
+    <li><button type="button" class="tag tag-input">Label</button></li>
+  </ul>`} /> */}
+
+### Disabled
+
+To disable a Tag input, just add `disabled` attribute to the `.tag`
```
MaxLardenois:
```suggestion
To disable a tag input, just add `disabled` attribute to the `.tag`
```

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+Standard tags are used in an informative way. Tag inputs are interactive and can take all the usual statuses: enabled, hover, pressed, disabled or focus. The Tag input should be used along with a `<button>`.
+
+They aren't auto removed at the moment, so you need to implement this behavior by yourself.
+
+<Example code={`<ul class="list-unstyled d-flex flex-wrap column-gap-xs">
+    <li><button type="button" class="tag tag-input">Label</button></li>
+    <li><button type="button" class="tag tag-input">Label</button></li>
+  </ul>`} />
+
+{/* Here is the expected rendering when there is a bunch of them.
```
MaxLardenois: either we show it or we delete this imo

scss/_tag.scss
```diff
@@ -0,0 +1,189 @@
...
+    fill: none;
+    stroke: var(--#{$prefix}tag-asset-color);
+    stroke-dasharray: var(--#{$prefix}tag-loading-dasharray);
+    stroke-width: 8;
+    transform-origin: center;
+    animation: var(--#{$prefix}tag-loading-animation);
+  }
+}
+
+.tag-disabled { // TODO: [aria-disabled="true"] ?
```
MaxLardenois: Not sure, what would be the use case of a disabled (normal) tag?

scss/_variables.scss
louismaximepiton: It was a Boosted specific component, so no need here.

scss/_tag.scss
louismaximepiton: IMO, it should be handled by project directly, so we should be deleting that line, let's see with other people, what to do with it. Or maybe show examples with some utilities

site/src/content/docs/components/tag.mdx
louismaximepiton: Let's keep this comment open, but I think it was done as an overview of all the components shapes. I don't like it neither since it shows the alignment with some other tags without effort, and not like it should.

site/src/content/docs/components/tag.mdx
louismaximepiton: I don't like it since font icon could be used by the project if they have more than 20-30 icons in the project.

site/src/content/docs/components/tag.mdx
louismaximepiton: No I meant as a sibling, but there no need here for the comprehension indeed !

site/src/content/docs/components/tag.mdx
louismaximepiton: I'd like to show it tbh, I only have this:
image
And I'm not sure it's the final state of it. What do you think, we should be showing it or not ?

scss/_tag.scss
louismaximepiton: I don't know tbh, let's wait and see if @hannahiss has more details about this.

site/src/content/docs/components/tag.mdx
MaxLardenois: I feel "really needed" is a French bad translation, I can be wrong. We have some hits on Linguee but for example the first is a translation by a french guy apparently.

Alternatively: 
```suggestion
If it proves necessary, you can use a font icon associated to the `.icon` class to set correct parameters for the `line-height`.
```

site/src/content/docs/components/tag.mdx
MaxLardenois: I'd say show it and the design or PO review will decide if it's necessary or not

site/src/pages/index.astro
MaxLardenois: Franco has proposed to switch to a Blue or Green tag and remove the link. The various new elements should have a direct link to their respective doc page.

scss/_tag.scss
vprothais: I was wondering also what could be a valid use case. Maybe if a section of the page is disabled because you need to check a radio or something to activate it ?
Since this component should only be used when really disabled and not on any other use case, I would use the aria-disabled selector here, but maybe it would be acceptable to have this aria on a parent too ?
We can definitely check with @hannahiss and designers.

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+<Example code={`<ul class="list-unstyled d-flex flex-wrap column-gap-xs row-gap-md">
+    <li>
+      <span class="tag">
+        <span class="tag-asset-container">
+          <span class="icon si si-settings" aria-hidden="true"></span>
+        </span>
+        Settings
+      </span>
+    </li>
+  </ul>`} />
```
vprothais: In the previous example with an SVG Icon, we just displayed the `span.tag`. Is there any reason to have the list in this example ?

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+
+#### Loader
+
+{/* TODO : do we have use cases for loader? Do we need visually hidden label? */}
```
vprothais: Yeah good question. I think the visually hidden label is handy for the user to know something is loading, but it would just say "loading", should we add an aria-labelledBy or something to give more context ?
Shouldn't we explain that the content of this `status role` should be udpated when the loading is done ?
Maybe we need to ask Vincent for this point.

site/src/content/docs/components/tag.mdx
```diff
@@ -0,0 +1,602 @@
...
+        </span>
+        Small neutral
+      </span>
+    </li>
+  </ul>`} />
+
+## Tag input
+
+<Callout type="warning">
+  Please note that all the variants above shouldn't be applied on this version of the component.
```
vprothais: I would probably use `mustn't` or `can't` here because it is really forbidden to do so. I'm not completely sure about mustn't, it doesn't seems to be widely used in UK english.

site/src/pages/index.astro
vprothais: @louismaximepiton , We could do as discussed with Franco here, having the non clicable tag, and putting the link on the following text, even with different links for each components and the last one going to the migration guide.

scss/_tag.scss
vprothais: this value works not so bad for title, but it's not very good on simple text, we can see that in migration guide where vertical alignment is a bit odd. Maybe we should remove this line.

scss/_tag.scss
```diff
@@ -0,0 +1,189 @@
...
+    top: calc(-.5 * ($ouds-tag-size-min-height-interactive-area - var(--bs-tag-min-height)) - var(--#{$prefix}tag-border-width)); // stylelint-disable-line function-disallowed-list
+    right: calc(-1 * var(--#{$prefix}tag-border-width)); // stylelint-disable-line function-disallowed-list
+    bottom: calc(-.5 * ($ouds-tag-size-min-height-interactive-area - var(--bs-tag-min-height)) - var(--#{$prefix}tag-border-width)); // stylelint-disable-line function-disallowed-list
+    left: calc(-1 * var(--#{$prefix}tag-border-width)); // stylelint-disable-line function-disallowed-list
+    content: "";
+  }
+
+  &::after {
+    width: var(--#{$prefix}tag-asset-container-size);
+    height: var(--#{$prefix}tag-asset-container-size);
```
vprothais: I would probably make this element size in rem, for it to grow with the root font size. In this case, I find it convey some meaning, so it could be interesting, WDYT ?
If we decide to do it, we can't use the custom pro, because it can't be defined in rem, we should probably use directly the scss variables with the rem mixin.

scss/_tag.scss
```diff
@@ -0,0 +1,189 @@
...
+  &:has(.tag-loader) {
+    --#{$prefix}tag-asset-padding: #{$ouds-tag-space-inset-loader-default};
+  }
+}
+
+.tag-bullet {
+  width: 100%;
+  height: 100%;
+  background-color: var(--#{$prefix}tag-asset-color);
+  @include border-radius(#{$ouds-tag-border-radius});
```
vprothais: I would probably use the token `$ouds-border-radius-pill` here directly rather than the tag border radius. In Orange case, the values are the same, but it may be different in other brand.

scss/_tag.scss
```diff
@@ -0,0 +1,189 @@
...
+}
+
+.tag-loader { // TODO copied from button loader, should be factorized
```
vprothais: should we have a file with placeholders that could be extended ? or a mixin ? The style on img,svg,.icon is also often used is the code

site/src/content/docs/migrations/migration.mdx
```diff
@@ -12,19 +12,23 @@ toc: true
 
 ### Contents
 
-- <span class="badge text-bg-status-negative-emphasized">Breaking</span> The focus-visible polyfill has been removed. So you no longer need it on your side, and you should replace all the calls to `:focus[data-focus-visible-added]` by `:focus-visible`.
+- <span class="tag tag-sm rounded-none text-bg-status-negative-emphasized">Breaking</span> The focus-visible polyfill has been removed. So you no longer need it on your side, and you should replace all the calls to `:focus[data-focus-visible-added]` by `:focus-visible`.
```
vprothais, line 15: The default style for tag is rounded. Should we use the default style or the square one ?
The documentation says square tags could be used for more formal or technical feel and for important notices. Not sure which one we should use. Maybe we could ask designers what they think ?

scss/_tag.scss
louismaximepiton: We might have trouble using this since if there is a `.text-bg-*`, it would override the selector since it's using `!important`. So unless we use it in here or we act that we don't want to use `.text-bg-*` classes in here, I think it'll be hard to achieve..

site/src/content/docs/components/tag.mdx
louismaximepiton: I think we said that one use case would be an action performed that changes a state. In this case, maybe the best here is to avoid saying it to avoid overwhelming people ?

scss/_tag.scss
louismaximepiton: I made the change for all, since the loading state might be important as well and the icon might be meaningful in some contexts.

scss/_tag.scss
louismaximepiton: That is a good idea indeed, we might dig that a bit.

site/src/content/docs/migrations/migration.mdx
louismaximepiton: Yes, we should definitely, but I think the use case is correct here 🤔

site/src/content/docs/components/buttons.mdx
hannahiss: Changes moved to PR #3100

site/src/content/docs/components/tag.mdx
hannahiss: I think I prefer the first one (If really needed)

site/src/content/docs/components/tag.mdx
hannahiss: I think this is useful and could take place in a "Layout" section

scss/_tag.scss
hannahiss: I can't understand this: how a not interactoive element can be desibaled... For me, it just a tag with a grey color, and so, the color is not accessible, which is an issue...

scss/_tag.scss
hannahiss: Yes we should do that

site/src/content/docs/components/tag.mdx
MaxLardenois, line 1: Remember to rename tag to tags cf #3124

MaxLardenois (changes requested):

scss/_tags.scss
```diff
@@ -1,141 +1,224 @@
...
+  gap: var(--#{$prefix}tag-asset-gap);
   align-items: center;
-  padding: subtract(var(--#{$prefix}tag-padding-y), calc(var(--#{$prefix}tag-font-shift) * .5)) var(--#{$prefix}tag-padding-x) add(var(--#{$prefix}tag-padding-y), calc(var(--#{$prefix}tag-font-shift) * .5));
-  @include font-size(var(--#{$prefix}tag-font-size));
-  font-weight: var(--#{$prefix}tag-font-weight);
-  line-height: add(var(--#{$prefix}tag-font-size), .125rem);
+  justify-content: center;
+  min-width: var(--#{$prefix}tag-min-width);
+  min-height: var(--#{$prefix}tag-min-height);
+  padding: subtract(var(--#{$prefix}tag-padding-y), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-end), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-y), var(--#{$prefix}tag-border-width)) subtract(var(--#{$prefix}tag-padding-start), var(--#{$prefix}tag-border-width));
```
MaxLardenois: What would you think about using directly calc() as subtract will be removed in Bootstrap 6
```suggestion
  padding: calc(var(--#{$prefix}tag-padding-y) - var(--#{$prefix}tag-border-width)) calc(var(--#{$prefix}tag-padding-end) - var(--#{$prefix}tag-border-width)) calc(var(--#{$prefix}tag-padding-y) - var(--#{$prefix}tag-border-width)) calc(var(--#{$prefix}tag-padding-start) - var(--#{$prefix}tag-border-width)); // stylelint-disable-line function-disallowed-list
```

scss/_tags.scss
```diff
@@ -1,141 +1,224 @@
...
+    --#{$prefix}tag-asset-color: #{$ouds-color-surface-status-warning-emphasized};
+  }
+
+  &.text-bg-status-negative-muted {
+    --#{$prefix}tag-asset-color: #{$ouds-color-surface-status-negative-emphasized};
+  }
+
+  &:has(.tag-asset-container) {
+    --#{$prefix}tag-padding-start: var(--#{$prefix}tag-padding-asset);
+  }
```
MaxLardenois: You missed this one I guess

```suggestion
```

scss/_tags.scss
```diff
@@ -1,141 +1,224 @@
...
+  --#{$prefix}tag-padding-end: #{$ouds-tag-space-padding-inline-default};
+  --#{$prefix}tag-padding-asset: #{$ouds-tag-space-padding-inline-asset-default};
+  --#{$prefix}tag-padding-y: #{$ouds-tag-space-padding-block-default};
+  --#{$prefix}tag-size-asset: #{px-to-rem($ouds-tag-size-asset-default)};
+  --#{$prefix}tag-asset-gap: #{$ouds-tag-space-column-gap-default};
+  --#{$prefix}tag-asset-color: currentcolor;
+  --#{$prefix}tag-loading-dasharray: 92;
+  --#{$prefix}tag-loading-animation: 2.1875s infinite linear rotate1-indeterminate, 1.25s linear infinite rotate2-indeterminate;
+  --#{$prefix}tag-positive-icon: url("data:image/svg+xml,<svg viewBox='0 0 960 960' xmlns='http://www.w3.org/2000/svg'><path fill-rule='evenodd' d='M500,75C265.279,75,75,265.279,75,500S265.279,925,500,925,925,734.721,925,500,734.721,75,500,75ZM722.051,339.988l-254.642,320-0.007-.006a39.635,39.635,0,0,1-62.125,0l-0.006.005-127.321-160,0.006-.006a40.142,40.142,0,0,1,2.928-53.267l15.915-16a39.612,39.612,0,0,1,54.022-2.077l0.006-.008L428.382,515,663.77,285.818l0.006,0.008A39.734,39.734,0,0,1,730.77,315a39.952,39.952,0,0,1-8.726,24.981Z' transform='scale(.96)' fill='currentColor'/></svg>");
```
MaxLardenois, line 16: Did you run SVGO on the SVGs? I get some bytes back but i don't know if we still do that if we use directly Solaris (also, shouldn't Solaris icons get the SVGO treatment by default?)

site/src/components/shortcodes/Example.astro
MaxLardenois, line 1: Should this be in this PR? Also why is it commented out?

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
 ---
 title: Tags
-description: Use tags to convey information, apply filters or display a selection of items.
+description: A tag is a small element that shows short info like a label, keyword, or category. It helps users quickly find, group, or understand content.
 aliases:
   - "/docs/components/tags/"
```
MaxLardenois, line 5: Adding aliases for non plural
```suggestion
  - "/docs/components/tags/"
  - "/docs/components/tag/"
  - "/docs/[[config:docs_version]]/components/tag/"
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+      <div class="card-body border-default border-thin border-top">
+        <a class="stretched-link" href="#tag">Tag</a> is used in an informative way, to show short info like a label, keyword, or category.
+      </div>
+    </div>
+  </li>
+  <li class="col">
+    <div class="card position-relative h-100">
+      <div class="card-top ratio ratio-16x9 bg-emphasized" aria-hidden="true">
+        <div class="d-flex align-items-center justify-content-center" data-bs-theme="dark">
+          <button type="button" class="tag tag-input" tabindex="-1">Tag input</button>
```
MaxLardenois:
```suggestion
          <button type="button" class="tag tag-input" tabindex="-1">Input tag</button>
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+
+<Example showToolbar={false} code={`<button type="button" class="tag tag-input">Input tag<span class="visually-hidden">Click to remove tag</span></button>`
+} />
+
+## Tag
+
+### Overview
+
+<Callout type="warning">
+For accessibility and semantics reasons, a tag -like any other informative element- must either be semantic element itself or be wrapped in appropriate semantic HTML element. In the examples below, tags appear as `<li>` elements or inside of `<h2>` and `<p>` elements. You may use any other semantic element that best fits the context.
```
MaxLardenois: Just [checked here](https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use) these should be dashes without spaces

```suggestion
For accessibility and semantics reasons, a tag—like any other informative element—must either be a semantic element itself or be wrapped in an appropriate semantic HTML element. In the examples below, tags appear as `<li>` elements or inside of `<h2>` and `<p>` elements. You may use any other semantic element that best fits the context.
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+    <li class="tag text-bg-status-accent-emphasized">Accent</li>
+    <li class="tag text-bg-status-positive-emphasized">Positive</li>
+    <li class="tag text-bg-status-info-emphasized">Info</li>
+    <li class="tag text-bg-status-warning-emphasized">Warning</li>
+    <li class="tag text-bg-status-negative-emphasized">Negative</li>
+  </ul>`} />
+
+
+<Callout type="warning">
+  Keep in mind that color alone should not be the only way to convey information, so if the color as a strong meaning, it should be reflected in the tag's text, or you should add a `.visually-hidden` additional text.
```
MaxLardenois: Feels redundant to me
```suggestion
  Keep in mind that color should not be the only way to convey information, so if the color has a strong meaning, it should be reflected in the tag's text, or you should add a `.visually-hidden` additional text.
```

or
```suggestion
  Keep in mind that color alone can not convey information, so if the color has a strong meaning, it should be reflected in the tag's text, or you should add a `.visually-hidden` additional text.
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+      Not tested
+    </li>
+    <li class="tag text-bg-status-negative-emphasized">
+      <span class="tag-status-icon"></span>
+      <span class="visually-hidden">Error:</span>
+      Tested
+    </li>
+  </ul>`} />
+
+In case you need to use a different icon than the predefined ones, you can confirgure your own icons by overriding the custom properties (CSS varibales) `--#{$prefix}tag-positive-icon`, `--#{$prefix}tag-info-icon` `--#{$prefix}tag-warning-icon` and `--#{$prefix}tag-error-icon`.
```
MaxLardenois:
```suggestion
In case you need to use a different icon than the predefined ones, you can confirgure your own icons by overriding the custom properties (CSS variables) `--#{$prefix}tag-positive-icon`, `--#{$prefix}tag-info-icon` `--#{$prefix}tag-warning-icon` and `--#{$prefix}tag-error-icon`.
```
Also elsewhere we use the computed variables names  (--bs-tag...) we should be consistent but I do not know what is the best option?

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+  <ul class="list-unstyled d-flex column-gap-xs">
+    <li class="tag tag-sm">Roaming</li>
+    <li class="tag tag-sm">5G Ready</li>
+  </ul>
+  <h2>Title <span class="tag align-middle">Coming soon</span></h2>
+  <h2>Title <span class="tag tag-sm align-middle">Coming soon</span></h2>
+  <p><span class="tag">New</span> Some text</p>
+  <p><span class="tag tag-sm">New</span> Some text</p>`} />
+
+Tags can be added side by side. They require horizontal and vertical spacing between each other. You can use [flexbox utilities]([[docsref:/utilities/spacing#gap]]) to manage the horizontal space (default value is obtained thanks to `.column-gap-xs`) the and vertical space (default value is obtained thanks to `.row-gap-md`).
```
MaxLardenois:
```suggestion
Tags can be added side by side. They require horizontal and vertical spacing between each other. You can use [flexbox utilities]([[docsref:/utilities/spacing#gap]]) to manage the horizontal space (default value is obtained thanks to `.column-gap-xs`) and the vertical space (default value is obtained thanks to `.row-gap-md`).
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+    <li class="tag">Label</li>
+    <li class="tag">Label</li>
+  </ul>`} />
+
+## Input tag
+
+### Overview
+
+<Callout type="warning">
+  Please note that all the variants of tag above must not be applied on input tag.
```
MaxLardenois: Just a suggestion but it draws attention
```suggestion
  Please note that all the variants of tag above *must not* be applied on input tag.
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+</Callout>
+
+Input tags are interactive and may be associated to an input field to be added dynamically. The input tag should be used along with a `<button>`, and can take all the usual states: enabled, hover, pressed, disabled or focused.
```
MaxLardenois: The input tag is a button I would say?
```suggestion
Input tags are interactive and may be associated to an input field to be added dynamically. The input tag should be used as a `<button>`, and can take all the usual states: enabled, hover, pressed, disabled or focused.
```

or
```suggestion
Input tags are interactive and may be associated to an input field to be added dynamically. The input tag should be used on a `<button>`, and can take all the usual states: enabled, hover, pressed, disabled or focused.
```

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,449 @@
...
+
+It can be removed on click, but you have to implement this behavior by yourself, with some JavaScript specific to your project.
+
+<Callout type="warning">
+  For accessibility reasons, input tags should be wrapped in a container with an appropriate `aria-label` to describe the group of tags. Each input tag should contain a `.visually-hidden` text to indicate the action that will be performed when the tag is clicked.
```
MaxLardenois: I guess Vincent A suggestion for Chip would be the same here? Though I am currently talking about it with him because I am not convinced it's better than this

site/src/components/shortcodes/Example.astro
hannahiss: Okay, sorry, that shouldn't be commented on. It's meant to be used under "Component types" to have more compact examples, and since these are really basic examples, there's no need for Stackblitz or copy... But displaying the code is still interesting.
Do you think I can leave it in the PR?

site/src/content/docs/components/tags.mdx
hannahiss: Well, let's wait a bit to have a final opinion on this... And I'll make the changes

scss/_tag.scss
hannahiss: I don't understand because `$ouds-tag-border-radius` = `$ouds-border-radius-pill` and is a given token, so we have to use it, right?

scss/_tags.scss
hannahiss: They are already processed by SVGO by the Finder tool. (I'll give you the link to the conversation about it)
But the question about which icon we have to use still stands... I'll ask it in the retro (but I cannot attend so you will-maybe-have an answer for me)

Aniort:
hello,
in "Overview", in the § "For accessibility and semantics reasons,.."
we must add at the end: "When the purpose of the tags is not clear enough (in particular without the visually context), an` aria-label` must be added to the `<ul>` tag to explain the tags function.

vprothais:
> hello, in "Overview", in the § "For accessibility and semantics reasons,.." we must add at the end: "When the purpose of the tags is not clear enough (in particular without the visually context), an` aria-label` must be added to the `<ul>` tag to explain the tags function.

@Aniort , you told us yesterday that we should use a visually hidden content in the first `li` on the list instead of an `aria-label` on the `ul`. In the end you prefer the `aria-label` ?

site/src/components/shortcodes/Example.astro
MaxLardenois: I think we can keep yes, but we will need to remember to use it on Badges and Chips

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,452 @@
...
+} />
+
+## Tag
+
+### Overview
+
+<Callout type="warning">
+For accessibility reasons, a tag, like any other informative element, must either be a semantic element itself or be wrapped in an appropriate semantic HTML element. In the examples below, the tags appear as `<li>` elements or inside of `<h2>` and `<p>` elements. You may use any other semantic element that best fits the context.
+
+When inserting tags into a list, wrap the list in a semantic container (usually a `<ul>` or `<ol>`). If the purpose of the tags is not clear enough (especially without visually context), an `aria-label` must be added to the container to explain the function of the tags.
```
vprothais:
```suggestion
When inserting tags into a list, wrap the list in a semantic container (usually a `<ul>` or `<ol>`). If the purpose of the tags is not clear enough (especially without visual context), an `aria-label` must be added to the container to explain the function of the tags.
```

site/src/content/docs/migrations/migration.mdx
```diff
@@ -12,33 +12,37 @@ toc: true
...
-- <span class="badge text-bg-status-negative-emphasized">Breaking</span> Button CSS variables have been changed. `--bs-button-mono-color-bg-minimal-loading` is no longer available.
--<span class="badge text-bg-status-positive-emphasized">New</span> `.btn-rounded` is now available.
--<span class="badge text-bg-status-positive-emphasized">New</span> `.btn-brand` is now available.
+- <span class="tag tag-sm rounded-none text-bg-status-negative-emphasized">Breaking</span> Button CSS variables have been changed. `--bs-button-mono-color-bg-minimal-loading` is no longer available.
+- <span class="tag tag-sm rounded-none text-bg-status-positive-emphasized">New</span> `.btn-rounded` is now available.
+- <span class="tag tag-sm rounded-none text-bg-status-positive-emphasized">New</span> `.btn-brand` is now available.
 
 #### Chips
 
 - <span class="badge text-bg-status-positive-emphasized">New</span> Chips components have been implemented. See more in our [Chips page]([[docsref:/components/chips]]).
```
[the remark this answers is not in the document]
MaxLardenois: Ok Copilot but I think we want to do that in another PR as it would add much noise to this one

site/src/content/docs/components/tags.mdx
```diff
@@ -1,9 +1,452 @@
...
+    <li class="tag">Roaming</li>
+    <li class="tag">5G Ready</li>
+    <li class="tag">4G</li>
+  </ul>`
+} />
+
+OUDS Web [color and background utilities]([[docsref:/helpers/color-background]]) can be applied to display colored tag variants. Colors are classified in two hierarchies, emphasized (for key tags) or muted (for secondary tags), and each color has its own semantic purpose. Please follow the tag design guidelines to choose the right color for the right usage.
+
+<Callout type="warning">
+Keep in mind that color should not be the only way to convey information. If the color has strong meaning, reflect it in the tag's text or add additional text with the class `.visually-hidden`.
```
vprothais: The "or" at the end of this sentence could be misleading, adding visually accessible content is ok for screen readers, but won't be useful for people having trouble with color vision. We should do the first, and the second if needed.

site/src/content/docs/migrations/migration.mdx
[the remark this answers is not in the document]
vprothais: @MaxLardenois , In fact he is right, we already replaced all badges by tags in migrations docs, but the Chips entry has not been modified because it was merged after the replacement. We should fix it in this PR

scss/_tags.scss
```diff
@@ -1,141 +1,221 @@
...
-  .btn-check:checked + & {
-    color: var(--#{$prefix}tag-active-color);
-    background-color: var(--#{$prefix}tag-active-decoration-color);
-    border-color: var(--#{$prefix}tag-active-decoration-color);
+  &.loading:has(.tag-loader) {
+    color: #{$ouds-color-content-default} if($enable-important-utilities, !important, null);
+    background-color: #{$ouds-color-surface-status-neutral-muted} if($enable-important-utilities, !important, null);
+  }
+
+  .tag-loader { // TODO partially copied from button loader, should be factorized as much as possible
```
vprothais, line 125: Is this `Todo` still relevant, should we do that now or is it done and we should remove the comment ?

site/src/content/docs/migrations/migration.mdx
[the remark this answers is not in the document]
MaxLardenois: Oh yes my bad

scss/_tags.scss
MaxLardenois: Added a reminder here: #3152

MaxLardenois (commented):
Need to add a review to comment?

vprothais (approved):

PR #3093, opened 2025-08-12
feat(component): Add Text-input component

nilloq:
### Related issues

Closes #

### Context & Motivation

Add the Text-input component

### Description

New feature (non-breaking change which adds functionality)

### Checklists

- [ ] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/ouds/main/.github/CONTRIBUTING.md)

- [ ] The changes are well displayed in [Storybook](https://deploy-preview-{your_pr_number}--boosted.netlify.app/storybook) (be careful if example order has changed for DSM)

- [ ] Code review
- [ ] Design review
- [ ] A11y review

### Live previews

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -960,12 +960,12 @@ $btn-icon-padding-x:          subtract($ouds-button-space-padding-inline-icon-st
 // Forms
 
 // scss-docs-start form-text-variables
-$form-text-margin-top:                  .4375rem !default; // OUDS mod
-$form-text-font-size:                   null !default; // OUDS mod: instead of `$small-font-size`
-$form-text-font-style:                  null !default;
-$form-text-font-weight:                 $font-weight-bold !default; // OUDS mod: instead of `null`
-$form-text-line-height:                 $line-height-sm !default; // OUDS mod
-$form-text-color:                       var(--#{$prefix}secondary-color) !default;
+$form-text-margin-top:                  $ouds-text-input-space-padding-block-top-helper-text !default; // OUDS mod
+// OUDS mod: no $form-text-font-size
+// OUDS mod: no $form-text-font-style
+// OUDS mod: no $form-text-font-weight
+// OUDS mod: no $form-text-line-height
+$form-text-color:                       var(--#{$prefix}color-content-muted) !default; // OUDS mod
 // scss-docs-end form-text-variables
 
 // scss-docs-start form-label-variables
@@ -988,8 +988,8 @@ $form-helper-label-margin-bottom:       $form-label-margin-bottom - divide(($for
 // scss-docs-end form-helper-variables
 
 // scss-docs-start form-input-variables
-$input-padding-y:                       .5rem !default; // OUDS mod: instead of $input-btn-padding-y (temporary value waiting for input component)
-$input-padding-x:                       $spacer * .5 !default;
+$input-padding-y:                       $ouds-text-input-space-padding-block-default !default; // OUDS mod
+$input-padding-x:                       $ouds-text-input-space-padding-inline-default !default; // OUDS mod
 $input-font-family:                     $input-btn-font-family !default;
 $input-font-size:                       $input-btn-font-size !default;
 $input-font-weight:                     $font-weight-bold !default;
@@ -1003,27 +1003,27 @@ $input-padding-y-lg:                    $input-btn-padding-y-lg !default;
 $input-padding-x-lg:                    $input-btn-padding-x-lg !default;
 $input-font-size-lg:                    $input-btn-font-size-lg !default;
 
-$input-bg:                              var(--#{$prefix}body-bg) !default;
-$input-disabled-color:                  var(--#{$prefix}secondary-color) !default; // OUDS mod
-$input-disabled-bg:                     var(--#{$prefix}secondary-bg) !default;
-$input-disabled-border-color:           null !default;
+$input-bg:                              var(--#{$prefix}color-action-support-enabled) !default; // OUDS mod
+$input-disabled-color:                  var(--#{$prefix}color-action-disabled) !default; // OUDS mod
+$input-disabled-bg:                     var(--#{$prefix}color-action-support-disabled) !default; // OUDS mod
+$input-disabled-border-color:           var(--#{$prefix}color-action-disabled) !default; // OUDS mod
 
-$input-color:                           var(--#{$prefix}body-color) !default;
-$input-border-color:                    var(--#{$prefix}color-border-default) !default; // OUDS mod: instead of var(--#{$prefix}border-color)
-$input-border-width:                    0 !default; // OUDS mod TODO to be modified for inputs
+$input-color:                           var(--#{$prefix}color-content-default) !default; // OUDS mod
+$input-border-color:                    $ouds-text-input-color-border-enabled !default; // OUDS mod
+$input-border-width:                    $ouds-text-input-border-width-default !default; // OUDS mod
 $input-box-shadow:                      none !default; // OUDS mod
 
-$input-border-radius:                   var(--#{$prefix}border-radius) !default;
+$input-border-radius:                   $ouds-text-input-border-radius-default !default; // OUDS mod
 $input-border-radius-sm:                var(--#{$prefix}border-radius-sm) !default;
 $input-border-radius-lg:                var(--#{$prefix}border-radius-lg) !default;
 
-$input-focus-bg:                        $input-bg !default;
-$input-focus-border-color:              currentcolor !default;
-$input-focus-color:                     $input-color !default;
+// OUDS mod: no $input-focus-bg
+// OUDS mod: no $input-focus-border-color
+// OUDS mod: no $input-focus-color
 $input-focus-width:                     $input-btn-focus-width !default;
 $input-focus-box-shadow:                none !default; // OUDS mod
 
-$input-placeholder-color:               var(--#{$prefix}secondary-color) !default;
+// OUDS mod: no $input-placeholder-color
 $input-plaintext-color:                 null !default; // OUDS mod: instead of `var(--#{$prefix}body-color)`
 
 // OUDS mod: no $input-height-border
@@ -1037,7 +1037,7 @@ $input-height-sm:                       1.875rem !default;
 $input-height-lg:                       3.125rem !default;
 $input-line-height-lg:                  var(--#{$prefix}font-line-height-body-large) !default; // OUDS mod
 
-$input-transition:                      border-color $transition-duration $transition-timing, $transition-focus !default;
+$input-transition:                      none !default; // OUDS mod
 
 $form-color-width:                      2.5rem !default; // OUDS mod: instead of `3rem`
 $form-color-border-color:               var(--#{$prefix}emphasis-color) !default; // OUDS mod
--- a/scss/forms/_form-text.scss
+++ b/scss/forms/_form-text.scss
@@ -4,9 +4,9 @@
 
 .form-text {
   margin-top: $form-text-margin-top;
-  @include font-size($form-text-font-size);
-  font-style: $form-text-font-style;
-  font-weight: $form-text-font-weight;
-  line-height: $form-text-line-height; // OUDS mod
+  @include get-font-size("label-medium"); // OUDS mod: instead of @include font-size($form-text-font-size);
+  // OUDS mod: no font-style
+  // OUDS mod: no font-weight
+  // OUDS mod: no line-height
   color: $form-text-color;
 }
--- a/site/src/content/docs/forms/text-input.mdx
+++ b/site/src/content/docs/forms/text-input.mdx
@@ -1,0 +1,141 @@
+---
+title: Text input
+description: A text input is a user interface component that allows users to enter, edit, or select single-line textual data. It provides a visual and interactive affordance for text entry while supporting labels, placeholders, icons, helper messages, and validation feedback.
+toc: true
+---
+
+<Callout type="info">
+{/* TODO: set the right url for design documentation */}
+You can find here the [OUDS Checkbox design guidelines](https://unified-design-system.orange.com/).
+</Callout>
+
+## Basic example
+
+<Example code={`<div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInput1" placeholder="">
+    <label for="exampleFormControlInput1">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInput0" placeholder="Firstname">
+    <label for="exampleFormControlInput0">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="email" class="form-control" id="exampleFormControlInput2" placeholder="name@example.com">
+    <label for="exampleFormControlInput2">Email address</label>
+  </div>`} />
+
+## Variants
+
+### Alternative
+
+A minimalist input with a transparent background and a visible stroke outlining the field. This style may be interesting for contexts other than form pages:
+* When inputs need to feel lightweight and unobtrusive
+* In a header (search field)
+* In a selection/filtering feature in a product catalog
+
+<Example code={`<div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative" id="exampleFormControlInputAlternative" placeholder="">
+    <label for="exampleFormControlInputAlternative">Firstname</label>
+  </div>`} />
+
+### Rounded
+
+For a finish with rounded corners.
+
+<Example code={`<div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-rounded" id="exampleFormControlInputRounded" placeholder="">
+    <label for="exampleFormControlInputRounded">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative form-control-rounded" id="exampleFormControlInputRoundedAlt" placeholder="">
+    <label for="exampleFormControlInputRoundedAlt">Firstname</label>
+  </div>`} />
+
+### Icon
+
+<Callout type="info" name="info-icons-svg-sprite" />
+
+// TODO
+
+### Helper text
+
+To display a helper text below inputs, add a `.form-text` as a sibling of a `.form-control`.
+
+<Callout type="warning">
+Form text should be explicitly associated with the form control it relates to using the `aria-describedby` attribute. This will ensure that assistive technologies—such as screen readers—will announce this form text when the user focuses or enters the control.
+</Callout>
+
+<Example code={`<div class="form-control-item">
+    <input type="password" id="inputPassword5" class="form-control" aria-describedby="passwordHelpBlock" placeholder="">
+    <label for="inputPassword5">Password</label>
+  </div>
+  <div id="passwordHelpBlock" class="form-text">
+    Your password must be 8-20 characters long, contain letters and numbers, and must not contain spaces, special characters, or emoji.
+  </div>`} />
+
+## States
+
+### Disabled
+
+Add the `disabled` boolean attribute on an input to give it a grayed out appearance, remove pointer events, and prevent focusing.
+
+<Example code={`<div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInputDisabledEmpty" placeholder="" disabled>
+    <label for="exampleFormControlInputDisabledEmpty">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInputDisabled" value="Doe" disabled>
+    <label for="exampleFormControlInputDisabled">Lastname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative" id="exampleFormControlInputAlternativeDisabledEmpty" placeholder="" disabled>
+    <label for="exampleFormControlInputAlternativeDisabledEmpty">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative" id="exampleFormControlInputAlternativeDisabled" value="Doe" placeholder="" disabled>
+    <label for="exampleFormControlInputAlternativeDisabled">Lastname</label>
+  </div>`} />
+
+### Readonly
+
+Add the `readonly` boolean attribute on an input to prevent modification of the input’s value. `readonly` inputs can still be focused and selected, while `disabled` inputs cannot.
+
+<Example code={`<div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInputReadonlyEmpty" placeholder="" readonly>
+    <label for="exampleFormControlInputReadonlyEmpty">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control" id="exampleFormControlInputReadonly" placeholder="" value="Doe" readonly>
+    <label for="exampleFormControlInputReadonly">Lastname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative" id="exampleFormControlInputAlternativeReadonlyEmpty" placeholder="" readonly>
+    <label for="exampleFormControlInputAlternativeReadonlyEmpty">Firstname</label>
+  </div>
+  <div class="form-control-item mb-3">
+    <input type="text" class="form-control form-control-alternative" id="exampleFormControlInputAlternativeReadonly" placeholder="" value="Doe" readonly>
+    <label for="exampleFormControlInputAlternativeReadonly" class="form-label">Lastname</label>
+  </div>`} />
+
+### Invalid
+
+<Callout type="info" name="input-invalid" />
+
+To display an invalid input, add `.is-invalid` to a `.form-control`. Please take a look at our [Validation page to learn more]([[docsref:/forms/validation]]).
+
+{/* <Example code={`<div class="form-control-item mb-3">
+    <label for="exampleFormControlInputInvalidEmpty">Firstname</label>
+    <input type="text" class="form-control is-invalid" id="exampleFormControlInputInvalidEmpty" placeholder="" readonly>
+  </div>
+  <div class="form-control-item mb-3">
+    <label for="exampleFormControlInputInvalid">Lastname</label>
+    <input type="text" class="form-control is-invalid" id="exampleFormControlInputInvalid" value="Doe" placeholder="" readonly>
+  </div>
+  <div class="form-control-item mb-3">
+    <label for="exampleFormControlInputAlternativeInvalidEmpty">Firstname</label>
+    <input type="text" class="form-control form-control-alternative is-invalid" id="exampleFormControlInputAlternativeInvalidEmpty" placeholder="" readonly>
+  </div>
+  <div class="form-control-item mb-3">
+    <label for="exampleFormControlInputAlternativeInvalid" class="form-label">Lastname</label>
+    <input type="text" class="form-control form-control-alternative is-invalid" id="exampleFormControlInputAlternativeInvalid" value="Doe" placeholder="" readonly>
+  </div>`} /> */}
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -2,25 +2,142 @@
 // General form controls (plus a few specific high-level interventions)
 //
 
+%form-item {
+  // scss-docs-start form-control-css-vars
+  --#{$prefix}form-control-color: var(--#{$prefix}color-content-default);
+  --#{$prefix}form-control-background-color: var(--#{$prefix}color-action-support-enabled);
+  --#{$prefix}form-control-border-color: #{$ouds-text-input-color-border-enabled};
+  --#{$prefix}form-control-border-width-top: 0px;
+  --#{$prefix}form-control-border-width-right: 0px;
+  --#{$prefix}form-control-border-width-bottom: #{$ouds-text-input-border-width-default};
+  --#{$prefix}form-control-border-width-left: 0px;
+  --#{$prefix}form-control-border-width-focus: 0px;
+  --#{$prefix}form-control-border-radius: #{$ouds-text-input-border-radius-default};
+  --#{$prefix}form-control-padding-x: #{$ouds-text-input-space-padding-inline-default};
+  --#{$prefix}form-control-padding-y: #{$ouds-text-input-space-padding-block-default};
+  --#{$prefix}form-control-max-width: #{$ouds-text-input-size-max-width};
+  --#{$prefix}form-control-min-height: #{$ouds-text-input-size-min-height};
+  --#{$prefix}form-control-min-width: #{$ouds-text-input-size-min-width};
+  // scss-docs-end form-control-css-vars
+
+  // scss-docs-start form-label-css-vars
+  --#{$prefix}form-label-color: var(--#{$prefix}color-content-muted);
+  // scss-docs-end form-label-css-vars
+}
+
+.form-control-item {
+  @extend %form-item;
+  position: relative;
+
+
+  > .form-control {
+    min-width: var(--#{$prefix}form-control-min-width);
+    max-width: var(--#{$prefix}form-control-max-width);
+    min-height: var(--#{$prefix}form-control-min-height);
+
+
+    /* stylelint-disable-next-line declaration-no-important */
+    padding-top: calc(var(--#{$prefix}form-control-padding-y) * 4) !important; // stylelint-disable-line function-disallowed-list
+
+
+    &::placeholder {
+      color: transparent;
+    }
+  }
+
+  > label {
+    position: absolute;
+    top: 0;
+    left: 0;
+    z-index: 2;
+    display: flex;
+    align-items: center;
+    max-width: var(--#{$prefix}form-control-max-width);
+    height: 100%;
+    padding-top: subtract(var(--#{$prefix}form-control-padding-y), var(--#{$prefix}form-control-border-width-top));
+    padding-right: subtract(var(--#{$prefix}form-control-padding-x), var(--#{$prefix}form-control-border-width-right));
+    padding-bottom: subtract(var(--#{$prefix}form-control-padding-y), var(--#{$prefix}form-control-border-width-bottom));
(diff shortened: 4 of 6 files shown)


scss/_variables.scss
```diff
@@ -960,12 +960,12 @@ $btn-icon-padding-x:          subtract($ouds-button-space-padding-inline-icon-st
...
+$form-text-margin-top:                  $ouds-text-input-space-padding-block-top-helper-text !default; // OUDS mod
 $form-text-font-size:                   null !default; // OUDS mod: instead of `$small-font-size`
-$form-text-font-style:                  null !default;
-$form-text-font-weight:                 $font-weight-bold !default; // OUDS mod: instead of `null`
-$form-text-line-height:                 $line-height-sm !default; // OUDS mod
-$form-text-color:                       var(--#{$prefix}secondary-color) !default;
+// OUDS mod: no $form-text-font-style:                  null !default;
+// OUDS mod: no $form-text-font-weight:                 null !default;
+// OUDS mod: no $form-text-line-height:                 $line-height-sm !default; // OUDS mod
+$form-text-color:                       var(--#{$prefix}color-content-muted) !default; // OUDS mod
```
MaxLardenois: Not sure if really useful but shouldn't we keep the original bootsrap value in OUDS mod: instead of ***
(Several times in this files)

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
+---
+title: Text input
+description: A text input is a user interface component that allows users to enter, edit, or select single-line textual data. It provides a visual and interactive affordance for text entry while supporting labels, placeholders, icons, helper messages, and validation feedback.
```
MaxLardenois: Does it really allow to "select single-line textual data"? This make me think of a `select input`... I'm not sure this is necessary

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+      Enter your price in Pounds (£) without decimals.
+    </div>
+  </div>
+  <div class="text-input">
+    <div class="text-input-container mt-md">
+    <svg aria-hidden="true">
+      <use xlink:href="/docs/${getConfig().docs_version}/assets/img/ouds-web-sprite.svg#heart-recommend"/>
+    </svg>
+    <label for="inputTextWithPrefixSuffixLeadingTrailing">Email</label>
+    <div class="input-container" data-bs-prefix="🖳" data-bs-suffix="@orange.com">
```
MaxLardenois, line 242: Why this character in `data-bs-prefix`?

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+import { getConfig } from '@libs/config'
+
+<Callout type="info">
+{/* TODO: set the right url for design documentation */}
+You can find here the [OUDS Text Input design guidelines](https://unified-design-system.orange.com/).
+</Callout>
+
+## Overview
+
+Wrap a pair of `<input class="text-input-field">` and `<label>` elements in `.text-input-container` to style OUDS textual form fields. Note that the `<label>` must come before the `<input>`.
```
MaxLardenois: We are not consistent on th `input` style vs `<input>` elsewhere in the documentation but this might be better? I think we should be consistent and switch to this version everywhere if everyone is ok.

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+
+A `placeholder` attribute is required on each `<input>` as our CSS-only floating label implementation relies on the `:placeholder-shown` pseudo-element to detect when the input is empty.
+
+If the placeholder text is empty, it is not visible; only the `<label>` is shown to users.
```
MaxLardenois:
```suggestion
If the placeholder text is empty, it is not visible: only the `<label>` is shown to users.
```

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+
+<Example code={`<div class="text-input-container mb-md">
+    <label for="exampleTextInputWithEmptyPlaceholder">Email address</label>
+    <input type="email" class="text-input-field" id="exampleTextInputWithEmptyPlaceholder" placeholder="">
+  </div>`} />
+
+If the placeholder text is filled, `<label>`s will automatically adjust to their floated position.
```
MaxLardenois:
```suggestion
If a placeholder text is supplied, `<label>`s will automatically adjust to their floated position.
```

nilloq:
> As we discussed, this is the text to add in the accessibility : To be compliant with the WCAG [1.3.5: Identify Input Purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose) criteria, for each input field collecting information about the user, add the appropriate token to the autocomplete attribute (sufficient [technique H98](https://www.w3.org/WAI/WCAG21/Techniques/html/H98)).

Thank, took it into acccount in the last update

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+<Example code={`<div class="text-input-container mb-md">
+    <label for="exampleTextInputWithValue">Input with value</label>
+    <input type="email" class="text-input-field" id="exampleTextInputWithValue" placeholder="name@example.com" value="test@example.com">
+  </div>`} />
+
+## Variants
+
+### Outlined
+
+A minimalist input with a transparent background and a visible stroke outlining the field. This style may be interesting for contexts other than form pages:
```
MaxLardenois:
```suggestion
Add `.text-input-container-outlined` for a minimalist input with a transparent background and a visible stroke outlining the field. This style may be interesting for contexts other than form pages:
```

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,512 @@
...
+* In a selection/filtering feature in a product catalog
+
+<Example code={`<div class="text-input-container text-input-container-outlined mb-md">
+    <label for="exampleTextInputOutlined">Firstname</label>
+    <input type="text" class="text-input-field" id="exampleTextInputOutlined" placeholder="">
+  </div>`} />
+
+### Rounded
+
+For a finish with rounded corners.
```
MaxLardenois:
```suggestion
Add `.text-input-container-rounded` for a finish with rounded corners.
```

site/src/content/docs/forms/text-input.mdx
nilloq: You're right, I removed the select mention... I initially copied pasted from Figma design,

site/src/content/docs/forms/text-input.mdx
nilloq: It's a sample to display an icon as prefix ;-)

site/src/content/docs/forms/text-input.mdx
nilloq: Ok for me

site/src/content/docs/forms/text-input.mdx
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
MaxLardenois: It looks like an error, shouldn't we put a real icon?

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,365 @@
...
+  --#{$prefix}text-input-border-width-right: 0px;
+  --#{$prefix}text-input-border-width-bottom: #{$ouds-text-input-border-width-default};
+  --#{$prefix}text-input-border-width-left: 0px;
+  --#{$prefix}text-input-border-width-focus: 0px;
+  --#{$prefix}text-input-border-radius: #{$ouds-text-input-border-radius-default};
+  --#{$prefix}text-input-padding-x: #{$ouds-text-input-space-padding-inline-default};
+  --#{$prefix}text-input-padding-y: #{$ouds-text-input-space-padding-block-default};
+  --#{$prefix}text-input-max-width: #{$ouds-text-input-size-max-width};
+  --#{$prefix}text-input-min-height: #{$ouds-text-input-size-min-height};
+  --#{$prefix}text-input-icon-size: #{px-to-rem($ouds-text-input-size-leading-icon)};
```
MaxLardenois: These icons looks decorative to me, if so we should not have px-to-rem here

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,365 @@
...
+  --#{$prefix}text-input-helper-margin-top: #{$ouds-text-input-space-padding-block-top-helper-text};
+  --#{$prefix}text-input-helper-color: var(--#{$prefix}color-content-muted);
+
+  --#{$prefix}text-input-error-color: var(--#{$prefix}color-content-status-negative);
+  --#{$prefix}text-input-invalid-icon: url("data:image/svg+xml,<svg viewBox='0 0 960 960' xmlns='http://www.w3.org/2000/svg'><path d='M500.5,74.75C265.088,74.75,74.249,265.59,74.249,501S265.088,927.25,500.5,927.25,926.751,736.412,926.751,501,735.912,74.75,500.5,74.75ZM460.578,237.1c10.208-10.906,23.587-16.481,39.92-17.04,16.319,0.559,29.7,6.134,39.907,17.04q16.134,17.256,16.149,48.079c0,15.088-21.166,292.627-26.357,303.885s-14.487,16.893-27.855,16.893c-0.666,0-1.2-.194-1.844-0.229-0.649.035-1.2,0.229-1.858,0.229q-20.05,0-27.842-16.893c-5.2-11.258-26.358-288.8-26.358-303.885Q444.44,254.357,460.578,237.1Zm79.273,529.988a56.906,56.906,0,0,1-38.99,14.85,58.393,58.393,0,0,1-39.535-14.675q-16.885-14.66-16.886-41.01,0-23.025,16.138-39.179t39.548-16.147q23.376,0,39.913,16.147a52.768,52.768,0,0,1,16.521,39.179Q556.56,752.232,539.851,767.088Z'/></svg>");
```
MaxLardenois: Move this in /packages/[brand]/scss/_composite.scss for both brands (will differ in wireframe)

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,365 @@
...
+  @extend %text-input-item;
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  gap: var(--#{$prefix}text-input-column-row-gap);
+  justify-content: center;
+  min-width: var(--#{$prefix}text-input-min-width);
+  max-width: var(--#{$prefix}text-input-max-width);
+  min-height: var(--#{$prefix}text-input-min-height);
+  padding-top: subtract(var(--#{$prefix}text-input-padding-y), var(--#{$prefix}text-input-border-width-top));
```
MaxLardenois: Bootstrap will drop subtract in v6, I suggest we do not use it and switch to calc

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,520 @@
...
+* When inputs need to feel lightweight and unobtrusive
+* In a header (search field)
+* In a selection/filtering feature in a product catalog
+
+<Example code={`<div class="text-input-container text-input-container-outlined mb-md">
+    <label for="exampleTextInputOutlined">Firstname</label>
+    <input type="text" class="text-input-field" id="exampleTextInputOutlined" placeholder="">
+  </div>`} />
+
+### Rounded
```
MaxLardenois, line 70: This paragraph is probably brand specific (Orange for now, business tools in the future), Sosh and wireframe are rounded by default

Use `<BrandSpecific />`

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,365 @@
...
+
+    &.text-input-container-outlined {
+      --#{$prefix}text-input-background-color: var(--#{$prefix}color-action-support-disabled);
+      --#{$prefix}text-input-border-width-top: 0px;
+      --#{$prefix}text-input-border-width-right: 0px;
+      --#{$prefix}text-input-border-width-bottom: 0px;
+      --#{$prefix}text-input-border-width-left: 0px;
+    }
+
+    outline: 0;
```
MaxLardenois: should we put this before line 207 so we do not forget that it's in at the same level as the custom props?

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,365 @@
...
+  // Disabled inputs
+  &:has(.text-input-field:disabled) {
+    --#{$prefix}text-input-color: var(--#{$prefix}color-action-disabled);
+    --#{$prefix}text-input-border-color: var(--#{$prefix}color-action-disabled);
+    &:not(.text-input-container-outlined) {
+      --#{$prefix}text-input-background-color: var(--#{$prefix}color-action-support-disabled);
+    }
+
+    // iOS fix for unreadable disabled content; see https://github.com/twbs/bootstrap/issues/11655.
+    opacity: 1;
```
MaxLardenois: Same remark than the previous one, this should go before I think

scss/forms/_text-input.scss
nilloq: In Firefox, when zooming text only, I find the rendering better if the icon grows at the same time

scss/forms/_text-input.scss
nilloq: I changed to use the $danger-icon form the _composite.scss file, but currently it's not up-to-date with design

MaxLardenois: Is it the same? I think this is a specific icon for this component (and other invalid inputs)
You can add a variable in composite.scss for this component icon I think

scss/forms/_text-input.scss
MaxLardenois: Agreed but we decided to never make decorative icons grow with font size see #3160

scss/forms/_text-input.scss
nilloq: Ok I switched all subtract and add to calc

site/src/content/docs/forms/text-input.mdx
nilloq: Ok I enclosed that in a `<BrandSpecific brand="orange">` section

scss/forms/_text-input.scss
nilloq: Done

scss/forms/_text-input.scss
nilloq: Done

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+## Overview
+
+Wrap a pair of `<input class="text-input-field">` and `<label>` elements in `.text-input-container` to style OUDS textual form fields. Note that the `<label>` must come before the `<input>`.
+
+A `placeholder` attribute is required on each `<input>` as our CSS-only floating label implementation relies on the `:placeholder-shown` pseudo-element to detect when the input is empty.
+
+If the placeholder text is empty, it is not visible: only the `<label>` is shown to users.
+
+<Callout type="warning">
+To be compliant with the WCAG [1.3.5: Identify Input Purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose) criteria, for each input field collecting information about the user, add the appropriate token to the autocomplete attribute (sufficient [technique H98](https://www.w3.org/WAI/WCAG21/Techniques/html/H98)).
```
vprothais: could we use the word `value` instead of token in this sentence ? I find it a bit disturbing since `token` really have a specific meaning in a DS context.

site/src/content/docs/forms/text-input.mdx
vprothais: angled bracket are already used in other documentation to talk about the HTML element, we can stick to that notation.

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+</Callout>
+
+<Example code={`<div class="text-input-container mb-md">
+    <label for="exampleTextInputWithEmptyPlaceholder">Email address</label>
+    <input type="email" autocomplete="email" class="text-input-field" id="exampleTextInputWithEmptyPlaceholder" placeholder="">
+  </div>`} />
+
+If the placeholder text is supplied, `<label>`s will automatically adjust to their floated position.
```
vprothais: since placeholder is singular, I would probably use singular for labels too, since there is only one per input

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+  </div>`} />
+
+### Trailing action
+
+Trailing actions can be added to the right side of the input field by placing a minimal icon only button (`.btn`, `.btn-minimal` and `.btn-icon`) inside the `.text-input-container`. It can be used to provide actions related to the field: clear input, toggle password visibility, open a date picker, etc.
+
+<Example code={`<div class="text-input-container mb-md">
+    <label for="exampleTextInputEmptyWithAction">Label</label>
+    <input type="email" class="text-input-field" id="exampleTextInputEmptyWithAction" placeholder="">
+    <button class="btn btn-minimal btn-icon">
```
vprothais: I have a weird behavior on Firefox 144 on MacOS. The hover effect disappear when the pointer is over the icon.

Image

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+      More information
+    </a>
+  </div>`} />
+
+### Input prefix and suffix
+
+A textual prefix or suffix can be added to an input by wrapping the `<input>` in a `.input-container` and adding `data-bs-prefix` and/or `data-bs-suffix` attributes to it.
+
+<Callout type="warning">
+For accessibility purpose the prefix and suffix should be explicitly mentioned in the helper text related to the input to using the `aria-describedby` attribute. This will ensure that assistive technologies—such as screen readers—will announce these additional or already existing elements when the user focuses or enters the input.
```
vprothais:
```suggestion
For accessibility purpose the prefix and suffix should be explicitly mentioned in the helper text related to the input by the `aria-describedby` attribute. This will ensure that assistive technologies—such as screen readers—will announce these additional or already existing elements when the user focuses or enters the input.
```
I'm not sure about that one, but I think we have an error in the current sentence around the "to using"

site/src/content/docs/forms/text-input.mdx
vprothais: I think it is related to the bug you fixed @MaxLardenois about the font icon path when documentation is built.
Because for now, the font seems to not be loaded : 

Image

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+Trailing actions can be added to the right side of the input field by placing a minimal icon only button (`.btn`, `.btn-minimal` and `.btn-icon`) inside the `.text-input-container`. It can be used to provide actions related to the field: clear input, toggle password visibility, open a date picker, etc.
+
+<Example code={`<div class="text-input-container mb-md">
+    <label for="exampleTextInputEmptyWithAction">Label</label>
+    <input type="email" class="text-input-field" id="exampleTextInputEmptyWithAction" placeholder="">
+    <button class="btn btn-minimal btn-icon">
+      <svg aria-hidden="true">
+        <use xlink:href="${getVersionedDocsPath('/assets/img/ouds-web-sprite.svg#heart-empty')}"/>
+      </svg>
+      <span class="visually-hidden">Default</span>
```
vprothais: Maybe we should make more clear that this label should be updated with a good label corresponding to the button goal. Either by using a real case or a text clearly saying, you should update that.

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+    </svg>
+    <label for="inputTextWithPrefixSuffixLeadingTrailing">Email</label>
+    <div class="input-container" data-bs-prefix="🖳" data-bs-suffix="@orange.com">
+      <input type="email" id="inputTextWithPrefixSuffixLeadingTrailing" aria-describedby="globalTextHelpBlock" class="text-input-field" placeholder="">
+    </div>
+    <button class="btn btn-minimal btn-icon">
+      <svg aria-hidden="true">
+        <use xlink:href="${getVersionedDocsPath('/assets/img/ouds-web-sprite.svg#heart-empty')}"/>
+      </svg>
+      <span class="visually-hidden">Default</span>
```
vprothais: Same here than for the previous example about the hidden label.

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+  </div>
+  <div class="text-input">
+    <div class="text-input-container mt-md">
+      <label for="exampleTextInputInvalid">Email</label>
+      <input type="text" class="text-input-field is-invalid" id="exampleTextInputInvalid" value="john.doe@example" aria-labelledby="emailFeedback" placeholder="">
+       <button class="btn btn-minimal btn-icon">
+        <svg aria-hidden="true">
+          <use xlink:href="${getVersionedDocsPath('/assets/img/ouds-web-sprite.svg#heart-empty')}"/>
+        </svg>
+        <span class="visually-hidden">Default</span>
```
vprothais: Same here :)

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+      </svg>
+      <label for="exampleTextInputInvalidTrailingWithPrefixSuffix">Price</label>
+      <div class="input-container" data-bs-prefix="£" data-bs-suffix=".00">
+        <input type="text" class="text-input-field is-invalid" id="exampleTextInputInvalidTrailingWithPrefixSuffix" value="-2.2XX" aria-describedby="priceTextHelpBlock" aria-labelledby="invalidPrice" placeholder="">
+      </div>
+      <button class="btn btn-minimal btn-icon">
+        <svg aria-hidden="true">
+          <use xlink:href="${getVersionedDocsPath('/assets/img/ouds-web-sprite.svg#trash')}"/>
+        </svg>
+        <span class="visually-hidden">Default</span>
```
vprothais: Here the icon seems pretty clear, maybe we could use a hidden label explaining that the button will clear the field.

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,522 @@
...
+  <div class="mb-md">
+    <label for="exampleFormControlTextarea1" class="form-label">Example textarea</label>
+    <textarea class="form-control" id="exampleFormControlTextarea1" rows="3"></textarea>
+  </div>`} />
+
+## Sizing
+
+Set heights using classes like `.form-control-lg` and `.form-control-sm`.
+
+<Example code={`<input class="form-control form-control-lg" type="text" placeholder=".form-control-lg" aria-label=".form-control-lg example">
```
vprothais, line 434: I don't know why but large and small inputs are rounded. Could we fix that ?

scss/forms/_form-text.scss
```diff
@@ -4,9 +4,9 @@
...
 .form-text {
   margin-top: $form-text-margin-top;
-  @include font-size($form-text-font-size);
-  font-style: $form-text-font-style;
-  font-weight: $form-text-font-weight;
-  line-height: $form-text-line-height; // OUDS mod
+  @include get-font-size("label-medium"); // OUDS mod: instead of @include font-size($form-text-font-size);
+  // OUDS mod: no font-style
+  // OUDS mod: no font-weight
+  // OUDS mod: no line-height
```
vprothais: line-height was already added by OUDS/Boosted. There is no need for an OUDS mod comment for removing it

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,372 @@
...
+}
+
+.text-input-container {
+  @extend %text-input-item;
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  gap: var(--#{$prefix}text-input-column-row-gap);
+  justify-content: center;
+  min-width: var(--#{$prefix}text-input-min-width);
```
vprothais: I can't find the declaration of this variable, is it still existing somewhere ?

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,372 @@
...
+  --#{$prefix}text-input-border-color: #{$ouds-text-input-color-border-enabled};
+  --#{$prefix}text-input-border-width-top: 0px;
+  --#{$prefix}text-input-border-width-right: 0px;
+  --#{$prefix}text-input-border-width-bottom: #{$ouds-text-input-border-width-default};
+  --#{$prefix}text-input-border-width-left: 0px;
+  --#{$prefix}text-input-border-width-focus: 0px;
+  --#{$prefix}text-input-border-radius: #{$ouds-text-input-border-radius-default};
+  --#{$prefix}text-input-padding-x: #{$ouds-text-input-space-padding-inline-default};
+  --#{$prefix}text-input-padding-y: #{$ouds-text-input-space-padding-block-default};
+  --#{$prefix}text-input-max-width: #{$ouds-text-input-size-max-width};
```
vprothais: Since text input contains text and overflowing content is hidden, maybe we should use a max width in rem here to have field growing with font size ?

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,372 @@
...
+.text-input-container {
+  @extend %text-input-item;
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  gap: var(--#{$prefix}text-input-column-row-gap);
+  justify-content: center;
+  min-width: var(--#{$prefix}text-input-min-width);
+  max-width: var(--#{$prefix}text-input-max-width);
+  min-height: var(--#{$prefix}text-input-min-height);
```
vprothais: With this size in px and a bigger user font-size, we get a UI shifting when we click in the field, because it becomes bigger with the floating label. I don't know what is better, to have a very big field all the time, or the UI shifting. Maybe empty and filled input should still have the same height, even with bigger font size ?

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,372 @@
...
+    color: var(--#{$prefix}text-input-label-color);
+    @include transition(font-size .1s ease-out);
+
+    // Allow the whole text-input-container to be clickable to focus the input
+    &::before {
+      position: absolute;
+      top: 0;
+      left: 0;
+      width: 100%;
+      height: 100%;
```
vprothais: we could replace these 4 lines by an `inset: 0` that should do the same.

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,372 @@
...
+      cursor: text;
+      content: "";
+      background-color: transparent;
```
vprothais, line 94: why do we need to specify this ? Is there a variant where we have a color here ?

site/src/content/docs/forms/text-input.mdx
nilloq: Ok so we can solved this one?

site/src/content/docs/forms/text-input.mdx
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

site/src/content/docs/forms/text-input.mdx
nilloq: Fixed by increasing z-index from 2 to 3

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

scss/forms/_text-input.scss
nilloq: Yes, my bad, I added it

scss/forms/_text-input.scss
nilloq: Ok it seems better with rem when growing with font size... I push an update

scss/forms/_text-input.scss
nilloq: I push a version with rem, it found it better, but it's my opinion. Perhaps the design would prefer a fixed max height?

scss/forms/_text-input.scss
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
MaxLardenois: Yes, can you open an issue to apply. this everywhere it's not?

scss/_variables.scss
vprothais: yeah you're right, we though it would help when importing Bootstrap commits, to compare changes. Actually, I don't know if it will be so useful in the end regarding all the changes we make. But since we tried to do it, we can continue to do so, unless you all think it is useless.

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,370 @@
...
+// Text input error message
+.text-input-container ~ .text-input-error {
+  @extend %text-input-message;
+  display: none;
+  margin-top: var(--#{$prefix}text-input-helper-margin-top);
+  color: var(--#{$prefix}text-input-error-color);
+}
+
+.text-input-container:has(.text-input-field.is-invalid) ~ .text-input-error,
+.text-input-container:has(.text-input-field:invalid) ~ .text-input-error {
```
vprothais: I think this selector is risky, it will show every text-input-error that are sibling of a text-input container in error.
In the documentation we have some example without the `.text-input` parent, so `text-input-container` are all siblings, if the first is invalid, then the error message of all the following one will be shown.
We should probably use the `.text-input` parent in every example to avoid that.

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,370 @@
+//
+// Text input
+//
+
+%text-input-item {
```
vprothais: Related to this comment : https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/3093/files#r2478024091
If we always need to have a `.text-input` parent, then we could declare all those variables on this container, to avoid re declaring many times.

scss/forms/_text-input.scss
vprothais: +1, we decided that icons that are only decorative should be in px to avoid taking too much space when increasing font-size, since they should not be useful for the user.

scss/forms/_text-input.scss
```diff
@@ -0,0 +1,370 @@
...
+  margin-top: var(--#{$prefix}text-input-helper-margin-top);
+  color: var(--#{$prefix}text-input-helper-color);
+}
+
+.text-input-container ~ .link {
+  @extend %text-input-message;
+}
+
+.text-input-container:has(.text-input-field.is-invalid):has(~ .text-input-error) ~ .text-input-helper,
+.text-input-container:has(.text-input-field:invalid):has(~ .text-input-error) ~ .text-input-helper {
```
vprothais: Hum I was not expecting a tilde selector in a `has()`. How could it work ? `has()` is for children, but the `~` is for sibling. How can they be children and siblings at the same time ?

scss/_variables.scss
nilloq: Yes, but these value have been removed from _form-text.css file. So if we keep them, the build failed telling there are unused variables

scss/forms/_text-input.scss
nilloq: Ok fine, I changed icon size to px 😉

scss/_variables.scss
vprothais: The point is to keep the previous value in the comment. Instead of just saying :
```scss
$form-text-color: var(--#{$prefix}color-content-muted) !default; // OUDS mod
```
Adding the previous value in the comment :
```scss
$form-text-color: var(--#{$prefix}color-content-muted) !default; // OUDS mod instead of --#{$prefix}secondary-color
```

scss/forms/_text-input.scss
nilloq: You're right, I update the whole documentation and ccs to use a `.text-input` wrapper systematically

scss/forms/_text-input.scss
nilloq: Ok done, but I'll probably reuse a placeholder selector again in Text Area implementation

scss/forms/_text-input.scss
nilloq: Yes it works! You can combine :has with + or ~... (see https://tobiasahlin.com/blog/previous-sibling-css-has/). 
Here it means that in the case of the `.text-input-container` has a sibling `.text-input-error`, the sibling `.text-input-helper` is not displayed 😉

scss/forms/_form-text.scss
nilloq: Ok done

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

site/src/content/docs/forms/text-input.mdx
nilloq: Ok fixed

scss/_variables.scss
nilloq: Ok fixed

scss/forms/_text-input.scss
nilloq: You're right, it was unuseful, I removed it

site/src/content/docs/forms/text-input.mdx
```diff
@@ -0,0 +1,560 @@
...
+</Callout>
+
+<Example code={`<div class="text-input">
+    <div class="text-input-container">
+      <label for="exampleTextInputWithEmptyPlaceholder">Email address</label>
+      <input type="email" autocomplete="email" class="text-input-field" id="exampleTextInputWithEmptyPlaceholder" placeholder="">
+    </div>
+  </div>`} />
+
+If the placeholder text is supplied, `<label>` will automatically adjust to their floated position.
```
vprothais:
```suggestion
If the placeholder text is supplied, `<label>` will automatically adjust to its floated position.
```

PR #3143, opened 2025-10-02
feat(lib): add Sosh brand

louismaximepiton:
### Types of change

- [x] Non-breaking change
- [ ] Breaking change (fix or feature that would change existing functionality and usage)

### Related issues

NA

### Context & Motivation

Need to handle many brands at Orange, here is a way, but not the only one.

### Description

Homepage:
- [ ] Banner image
- [x] ~~Package name~~
- [x] ~~CDN links, associated hashes~~
- [x] ~~path must be updated~~

Transversal
- [x] ~~Orange reference to update to Sosh or ${brand}~~
- [x] ~~Favicon~~
- [x] ~~Update doc CSS to use tokens instead of hard coded values~~
- [x] ~~Use special set of icons ? Need to update docs examples respectively~~
- [ ] Use Sosh font
- [x] All colored bg examples must be remastered
- [x] Restore a good rendering for the tarteaucitron
- [x] Restore a good rendering for the search

~~Introduction:~~
- [x] ~~Different Codepen for different brands~~
- [x] ~~Package name~~
- [x] ~~CDN links, associated hashes~~

Download:
- [x] ~~Package name~~
- [x] ~~CDN links, associated hashes~~
- [x] ~~Careful about release urls and archives~~ -> Need to determine what will be the release process

~~Contents:~~
- [x] ~~ouds-web to be changed in a Highlighted~~
- [x] ~~Change the structure of the content~~
- [x] ~~Change the content of the page~~

~~Contribute:~~
- [x] ~~Change the wording about what needs to be done by user to contribute~~
- [x] ~~Update commands~~

~~Options:~~
- [x] ~~Values in the table to update~~

~~Color-palette:~~
- [x] ~~Change the whole palette.yml and probably other files in site/data~~
- [x] ~~If we keep the token reference, probably need to update the filepath~~
- [x] ~~Change import stack and probably refer to a file~~

~~Color modes:~~
- [x] ~~CDN links to update~~
- [x] ~~Include paths need to be updated~~
- [x] ~~If we keep the token reference, probably need to update the filepath~~

~~Breakpoints:~~
- [x] ~~Breakpoints could be different and also dimensions~~
- [x] ~~Customizable breakpoints to remove or replace in custom libraries~~
- [x] ~~Update comments with values inside~~

~~Containers:~~
- [x] ~~Values could be different and also dimensions~~
- [x] ~~Change import stack and probably refer to a file~~
- [x] ~~Remove Sass reference~~

~~Grid:~~
- [x] ~~Values could be different and also dimensions + calculations with scss tokens in doc~~
- [x] ~~Change import stack and probably refer to a file~~
- [x] ~~Customizable grid to remove or replace in custom libraries~~
- [x] ~~Remove Sass reference~~

~~Gutters:~~
- [x] ~~Values of gutters to remove or change~~
- [x] ~~Customizable gutters to remove or replace in custom libraries~~

Layout/utilities:
- [x] ~~Remove or handle margin and padding~~
- [x] Or remove page ?

~~CSS grid:~~
- [x] ~~Customizable part to remove or replace in custom libraries~~
- [x] ~~Remove Sass reference~~

~~Reboot:~~
- [x] ~~Rewrite the whole page, nothing to be left aside, the values, the sense of what's written.~~

~~Typography:~~
- [x] ~~Values in tables~~
- [x] ~~Remove Sass reference~~

~~Checkbox, radio, switch:~~
- [x] ~~Link to DSM to update ?~~
- [x] ~~Fix the rendering~~

~~Breadcrumb, links:~~
- [x] ~~Fix the rendering for the chevron in the middle of the component.~~

~~Button:~~
- [x] ~~Rounded section to remove for brand including title and toc~~
- [x] ~~Handle colored bg sections~~

~~Color&Background:~~
- [x] ~~Missing text-bg-brand-secondary and text-bg-brand-tertiary~~

~~Utilities: Remove Sass reference~~

~~Background:~~
- [x] ~~Missing bg-brand-secondary and bg-brand-tertiary~~
- [x] ~~How to handle data-bs-theme correctly and easily with tokens maybe ?~~

~~Border:~~
- [x] ~~Missing border-brand-secondary/tertiary and border-on-bg-brand-secondary/tertiary~~

~~Color:~~
- [x] ~~Missing text-brand-secondary/tertiary and text-on-bg-brand-secondary/tertiary~~
- [x] ~~Values in some paragraph~~
- [x] ~~Table to see back if this is needed~~

~~Flex:~~
- [x] ~~Remove Media object~~

~~Spacing:~~
- [x] ~~Reference to tokens to update maybe, and their final value or remove.~~
- [x] ~~Reference to tokens scaled to update maybe, and their final value or remove.~~

~~Text:~~
- [x] ~~Update the badge example~~
- [x] ~~Update font-size values or remove~~

~~Icons:~~
- [x] ~~Solaris icons ?~~
- [x] ~~Value to update or remove~~

~~Translations:~~
- [x] ~~Remove page~~

~~Migration from Boosted:~~
- [x] ~~Probably cut this page for other brands and update the side menu~~

~~Page with nothing to change:~~
~~Browsers&devices, accessibility (probably), columns, z-index, forms/overview, validation, Badge, Bullet list, Clearfix, Colored links, Divider, helpers/position, ratio, stacks, stretched link, text truncation, visually-hidden, API (probably), display, float, interaction, object fit, opacity, overflow, utilities/position, shadows, sizing, vertical align, visibility, utilities/z-index, approach, team, license, cookies, migration~~

Examples:
- [ ] Change thumbnails
- [x] ~~Check CSS~~

Handle this somehow:
- [x] Composer.json
- [x] ~~Licenses & copyrights to ORANGE SA anyway ?~~
- [x] ~~Notices~~
- [x] Package.js
- [x] ~~Package.json~~
  - [x] ~~scripts~~
  - [x] ~~description~~
  - [x] ~~files~~
- [x] Readmes
- [x] Contributing
- [x] ~~Support~~
- [x] ~~Manage Storybook brands somehow~~
- [x] Nuget
- [x] ~~Update all build files~~
- [x] ~~Update all hashes with https://www.srihash.org/ ~~
- [x] ~~Check sitemap~~
- [x] ~~Update migration guides~~
- [x] ~~Update data-bs-theme in the doc to be brand compatible~~
- [x] ~~Remove comments from tokenator in raw~~
- [x] ~~Clean tokenator~~
- [x] ~~define vars for css bg in docs~~
- [x] ~~See where to add the new file structure and explain how to use~~
- [x] ~~check component rendering in the doc~~
- [x] ~~check bundlewatch~~
- [x] Check all commits in merge
- [x] Check why doc scss variables aren't good
- [x] ~~Check all `.bg-` and add getMode when needed~~
- [x] ~~Check all the CI files~~
- [x] Reintroduce Customize > Sass
- [x] Check workflow files
- [x] Release scripts and processes

After PR is merged:
- [ ] Make redirects on the ouds-web-doc already existing
- [ ] Probably warn about storybook changing its urls for DSM
- [ ] Probably cut Custom libraries page for other brands and update the side menu
- [ ] Manage Storybook for multi branding
- [ ] Reintroduce values of tokens in the doc ? Spacing, Text, Icons, Reboot, Typography
- [ ] Reintroduce Solaris icons
- [x] Better the CSS for Prod in Head.astro
- [ ] Adapt the PR template to add the brand by default
- [ ] Check if sitemap from Sosh and Orange can be merged somehow

A11y review:
- [ ] Header brand selector

### Live previews

- <https://deploy-preview-3143--boosted.netlify.app/orange/>

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "release": "npm-run-all storybook-build release-zip*",
     "release-sri": "node build/generate-sri.mjs",
     "release-version": "node build/change-version.mjs",
-    "release-zip": "npm run release-zip -w packages/sosh && npm run release-zip -w packages/orange",
+    "release-zip": "npm-run-all release-zip -w packages/sosh release-zip -w packages/orange",
     "release-zip-examples": "node build/zip-examples.mjs",
     "dist": "npm-run-all --aggregate-output --parallel css js",
     "test": "npm-run-all lint dist js-test docs-build docs-lint",
--- a/scss/_chips.scss
+++ b/scss/_chips.scss
@@ -105,83 +105,85 @@
 }
 
 // Filter chip specific style that handles images, focus-visible and disabled states
-.chip-filter .chip-interactive:where(:has(svg, img, .icon)) {
-  --#{$prefix}chip-padding-end: #{$ouds-chip-space-padding-inline-icon};
-}
+.chip-filter {
+  .chip-interactive:where(:has(svg, img, .icon)) {
+    --#{$prefix}chip-padding-end: #{$ouds-chip-space-padding-inline-icon};
+  }
 
-.chip-filter input:focus-visible ~ .chip-interactive,
-.chip-filter .chip-interactive:has(input:focus-visible) {
-  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-unselected-focus};
-  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-unselected-focus};
-  --#{$prefix}chip-border-width: #{px-to-rem($ouds-chip-border-width-unselected-interaction)};
-  --#{$prefix}chip-color: #{$ouds-chip-color-content-unselected-focus};
+  input:focus-visible ~ .chip-interactive,
+  .chip-interactive:has(input:focus-visible) {
+    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-unselected-focus};
+    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-unselected-focus};
+    --#{$prefix}chip-border-width: #{px-to-rem($ouds-chip-border-width-unselected-interaction)};
+    --#{$prefix}chip-color: #{$ouds-chip-color-content-unselected-focus};
 
-  @include focus-visible();
-}
+    @include focus-visible();
+  }
 
-.chip-filter input:disabled ~ .chip-interactive,
-.chip-filter .chip-interactive:has(input:disabled) {
-  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-unselected-disabled};
-  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-unselected-disabled};
-  --#{$prefix}chip-color: #{$ouds-chip-color-content-unselected-disabled};
+  input:disabled ~ .chip-interactive,
+  .chip-interactive:has(input:disabled) {
+    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-unselected-disabled};
+    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-unselected-disabled};
+    --#{$prefix}chip-color: #{$ouds-chip-color-content-unselected-disabled};
 
-  pointer-events: none;
-}
+    pointer-events: none;
+  }
 
-// Filter chip specific style for selected state
-.chip-filter input:checked ~ .chip-interactive,
-.chip-filter .chip-interactive[aria-pressed="true"],
-.chip-filter .chip-interactive:has(input:checked) {
-  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-enabled};
-  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-enabled};
-  --#{$prefix}chip-border-width: #{px-to-rem($ouds-chip-border-width-selected)};
-  --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-enabled};
-  --#{$prefix}chip-padding-start: #{$ouds-chip-space-padding-inline-icon};
-  --#{$prefix}chip-tick-color: #{$ouds-chip-color-content-selected-tick-enabled};
+  // Filter chip specific style for selected state
+  input:checked ~ .chip-interactive,
+  .chip-interactive[aria-pressed="true"],
+  .chip-interactive:has(input:checked) {
+    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-enabled};
+    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-enabled};
+    --#{$prefix}chip-border-width: #{px-to-rem($ouds-chip-border-width-selected)};
+    --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-enabled};
+    --#{$prefix}chip-padding-start: #{$ouds-chip-space-padding-inline-icon};
+    --#{$prefix}chip-tick-color: #{$ouds-chip-color-content-selected-tick-enabled};
 
-  &:hover {
-    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-hover};
-    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-hover};
-    --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-hover};
+    &:hover {
+      --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-hover};
+      --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-hover};
+      --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-hover};
+      --#{$prefix}chip-tick-color: currentcolor;
+    }
+
+    &:active {
+      --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-pressed};
+      --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-pressed};
+      --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-pressed};
+      --#{$prefix}chip-tick-color: currentcolor;
+    }
+
+    &::before {
+      width: 1em;
+      height: 1em;
+      overflow: hidden;
+      font-size: px-to-rem($ouds-chip-size-icon);
+      line-height: 1;
+      color: var(--#{$prefix}chip-tick-color);
+      content: "";
+      background-color: currentcolor;
+      mask: $chip-tick-icon no-repeat 50% / px-to-rem($ouds-chip-size-icon);
+    }
+  }
+
+  input:checked:focus-visible ~ .chip-interactive,
+  .chip-interactive[aria-pressed="true"]:focus-visible,
+  .chip-interactive:has(input:checked:focus-visible) {
+    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-focus};
+    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-focus};
+    --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-focus};
     --#{$prefix}chip-tick-color: currentcolor;
   }
 
-  &:active {
-    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-pressed};
-    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-pressed};
-    --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-pressed};
+  input:checked:disabled ~ .chip-interactive,
+  .chip-interactive[aria-pressed="true"]:disabled,
+  .chip-interactive:has(input:checked:disabled) {
+    --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-disabled};
+    --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-disabled};
+    --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-disabled};
     --#{$prefix}chip-tick-color: currentcolor;
   }
-
-  &::before {
-    width: 1em;
-    height: 1em;
-    overflow: hidden;
-    font-size: px-to-rem($ouds-chip-size-icon);
-    line-height: 1;
-    color: var(--#{$prefix}chip-tick-color);
-    content: "";
-    background-color: currentcolor;
-    mask: $chip-tick-icon no-repeat 50% / px-to-rem($ouds-chip-size-icon);
-  }
-}
-
-.chip-filter input:checked:focus-visible ~ .chip-interactive,
-.chip-filter .chip-interactive[aria-pressed="true"]:focus-visible,
-.chip-filter .chip-interactive:has(input:checked:focus-visible) {
-  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-focus};
-  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-focus};
-  --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-focus};
-  --#{$prefix}chip-tick-color: currentcolor;
-}
-
-.chip-filter input:checked:disabled ~ .chip-interactive,
-.chip-filter .chip-interactive[aria-pressed="true"]:disabled,
-.chip-filter .chip-interactive:has(input:checked:disabled) {
-  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-disabled};
-  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-disabled};
-  --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-disabled};
-  --#{$prefix}chip-tick-color: currentcolor;
 }
 
 // Icon chip specific style
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -331,11 +331,6 @@ sup { top: -.5em; }
     text-decoration: $link-hover-decoration;
   }
 
-  .visited-links &:visited:not(.link, .link-chevron, .icon-link, :hover, :active, :focus-visible),
-  &:visited.visited-links:not(.link, .link-chevron, .icon-link, :hover, :active, :focus-visible) {
-    color: var(--#{$prefix}link-visited-color);
-  }
-
   &:focus-visible:where(:not([aria-disabled="true"])) {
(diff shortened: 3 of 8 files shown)


build/change-version.mjs
vprothais, line 73: I'm not sure how it works with npm workspace but shouldn't we do that also for all packages in order to have the good version number in packages.json of each package ?

build/generate-sri.mjs
```diff
@@ -19,54 +19,58 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url))
 
 sh.config.fatal = true
 
-const configFile = path.join(__dirname, '../config.yml')
+const BRANDS = ['orange', 'sosh']
```
vprothais: I've not read all the changes yet, but is there a lot of occurrences of this array ? Should we loop on folder names in packages instead of this static array ?

packages/orange/nuget/ouds-web-orange.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <tags>CSS mobile-first responsive front-end framework web bootstrap</tags>
+    <contentFiles>
+      <files include="**/*" buildAction="Content" />
+    </contentFiles>
+  </metadata>
+  <files>
+    <file src="nuget\ouds-web.png" target="" />
+
+    <file src="dist\css\ouds-web*.*" target="content\Content" />
```
vprothais: these paths are related to which folder ?
Maybe it is linked to the nuget command but I find it weird to have dist here directly :)

packages/orange/composer.json
```diff
@@ -0,0 +1,31 @@
+{
```
vprothais: since composer work per repository and not per package. I'm not sure having a composer.json in each package would be useful.

packages/orange/config.yml
```diff
@@ -24,6 +26,7 @@ icons_usage:                      "https://design.orange.com/0c1af118d/p/92bb17-
 bootstrap:                        "https://getbootstrap.com"
 ouds:
   web:                            "https://unified-design-system.orange.com/" # TODO: replace this URL by the web one or even "https://oran.ge/ds-web" when it will changed
+debug_template:                   "https://codepen.io/louismaximepiton/pen/WbreEwY"
```
vprothais, line 29: We must not forget to test it when the package is published :)

packages/orange/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## What’s included
+
+Within the download you’ll find the following directories and files, logically grouping common assets and providing both compiled and minified variations.
+
+<details>
+  <summary>Download contents</summary>
+
+  ```text
+  ouds-web-orange/
```
vprothais: I think the directory in the package will be named `web-orange` inside a `@ouds` directory.

packages/orange/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+OUDS Web’s documentation, included in this repo in the root directory, is built with [Astro](https://astro.build/) and publicly hosted on GitHub Pages at <https://web.unified-design-system.orange.com/orange/>. The docs may also be run locally.
+
+Documentation search is powered by [Algolia’s DocSearch](https://docsearch.algolia.com/).
+
+### Running documentation locally
+
+1. Run `npm install` to install the Node.js dependencies, including Hugo (the site builder).
+2. Run `npm run dist` (or a specific npm script) to rebuild distributed CSS and JavaScript files, as well as our docs assets.
+3. From the root `/ouds-web` directory, run `npm run astro-dev` in the command line.
```
vprothais: I'm a bit lost with this part of the documentation. Should it be done from the root of the repository ? If so, commands and forlders are not corresponding with what we have.

packages/sosh/composer.json
```diff
@@ -0,0 +1,31 @@
+{
```
vprothais: Like for Orange, not sure it is used at all.

NOTICE.txt
```diff
@@ -9,23 +9,29 @@ and reproducing the content of the NOTICE and DOCUMENTATION files.
...
-site/static/docs/**/assets/img/favicons/mstile-150x150.png
-site/static/docs/**/assets/img/favicons/mstile-150x150.png
-site/static/docs/**/assets/img/favicons/mstile-310x150.png
-site/static/docs/**/assets/img/favicons/mstile-310x310.png
-site/static/docs/**/assets/img/favicons/mstile-70x70.png
-site/static/docs/**/assets/img/favicons/safari-pinned-tab.svg
-site/static/docs/**/assets/img/ouds-web-sprite.svg
+packages/**/nuget/ouds-web.png
+packages/**/scss/tokens/composite.scss
+site/static/[brand]/docs/[version]/assets/brand/orange-logo.svg
```
vprothais, line 16: Is it ok to use that as a placeholder ? I think this file is here for legal reason, we should be sure the format is ok. If we have a doubt maybe we should use the `**` wildcard ?

build/change-version.mjs
louismaximepiton: Not quite sure how it would behave neither, I'd like to wait for the first release we make to update it, just to be sure.

louismaximepiton: Because `npm i` updates and recovers all the dev dependencies from the packages for example

louismaximepiton: Just tested and good to go!

packages/orange/nuget/ouds-web-orange.nuspec
louismaximepiton: Changed it !

packages/orange/config.yml
louismaximepiton: Definitely

packages/orange/README.md
louismaximepiton: I'm unsure about that, I don't know what it refers to, Changed it in the meantime

packages/orange/README.md
louismaximepiton: Tried something if it's clearer

NOTICE.txt
louismaximepiton: I don't think there is any placeholder in here, the directory have a weird name because they are Astro placeholders, but these are the exact path

packages/orange/README.md
vprothais: I just saw that it is already named ouds-web currently in ouds/main. Maybe it was done before we got the @ouds scope ¯\_(ツ)_/¯

NOTICE.txt
vprothais: Oh sorry, you're right, it is the Astro convention.

scss/mixins/_color-mode.scss
```diff
@@ -23,6 +23,7 @@
     }
   } @else {
     [data-bs-theme="#{$mode}"],
+    #{$ouds-root-selector}[data-bs-theme="#{$mode}"],
```
vprothais, line 26: Was it done due to selector specificity if the root-selector have a bigger specificity than attribute selector ?

scss/_reboot.scss
```diff
@@ -369,6 +368,21 @@ a {
   @extend %link-properties;
 }
 
+@if $ouds-root-selector != ":root" {
```
vprothais: I'm not sure to understand why this is done only when different from `:root` ?

package.json
```diff
@@ -69,36 +61,28 @@
...
+    "docs-serve": "npm run start",
     "docs-serve-only": "sirv _site --no-clear --port 9001",
     "lockfile-lint": "lockfile-lint --allowed-hosts npm --allowed-schemes https: --empty-hostname false --type npm --path package-lock.json",
     "update-deps": "ncu -u -x eslint,eslint-config-xo,eslint-plugin-unicorn,karma-browserstack-launcher,karma-rollup-preprocessor,sass",
-    "release": "npm-run-all dist release-sri docs-build storybook-build release-zip*",
+    "release": "npm-run-all storybook-build release-zip*",
     "release-sri": "node build/generate-sri.mjs",
     "release-version": "node build/change-version.mjs",
-    "release-zip": "cross-env-shell \"rm -rf ouds-web-$npm_package_version-dist ouds-web-$npm_package_version-dist.zip && cp -r dist/ ouds-web-$npm_package_version-dist && zip -qr9 ouds-web-$npm_package_version-dist.zip ouds-web-$npm_package_version-dist && rm -rf ouds-web-$npm_package_version-dist\"",
+    "release-zip": "npm run release-zip -w packages/sosh && npm run release-zip -w packages/orange",
```
vprothais: We could use npm-run-all here since it is already used for other scripts

scss/_chips.scss
```diff
@@ -104,86 +104,84 @@
...
+  --#{$prefix}chip-tick-color: currentcolor;
+}
+
+.chip-filter input:checked:disabled ~ .chip-interactive,
+.chip-filter .chip-interactive[aria-pressed="true"]:disabled,
+.chip-filter .chip-interactive:has(input:checked:disabled) {
+  --#{$prefix}chip-bg: #{$ouds-chip-color-bg-selected-disabled};
+  --#{$prefix}chip-border-color: #{$ouds-chip-color-border-selected-disabled};
+  --#{$prefix}chip-color: #{$ouds-chip-color-content-selected-disabled};
+  --#{$prefix}chip-tick-color: currentcolor;
```
vprothais: All selectors from line 107 to this one starts with the `.chip-filter` class, should we use nested selectors with this one as a parent. The goal is to group all the chip-filter selectors.

site/src/components/head/Head.astro
```diff
@@ -2,9 +2,10 @@
 import { getConfig } from '@libs/config'
 import { getVersionedDocsPath } from '@libs/path'
 import type { Layout } from '@libs/layout'
-import Stylesheet from '@components/head/Stylesheet.astro'
+// import Scss from '@components/head/Scss.astro'
```
vprothais: Do we need to keep this commented line for something ?

site/src/components/header/Navigation.astro
```diff
@@ -104,6 +104,49 @@ const { addedIn, layout, title } = Astro.props
...
+            <button
+              class='nav-link nav-icon dropdown-toggle'
+              id="bd-brand"
+              type="button"
+              aria-expanded="false"
+              data-bs-toggle="dropdown"
+              data-bs-display="static"
+              aria-label={`Toggle brand (${getConfig().brand})`}
+            >
+              <img src={getVersionedDocsPath(`assets/brand/${getConfig().brand}-logo.svg`)} alt="Toggle brand" loading="lazy" />
```
vprothais: isn't the alt redundant with the content of the following span ? Since we have an aria-label, shouldn't we have an empty alt here ?

site/src/components/header/Navigation.astro
```diff
@@ -104,6 +104,49 @@ const { addedIn, layout, title } = Astro.props
...
+              <span class='d-lg-none ms-2' id="bd-brand-text">Toggle brand</span>
+            </button>
+            <ul
+              class=`dropdown-menu dropdown-menu-end mb-2`
+              aria-labelledby="bd-brand-text"
+            >
+              <li>
+                <a
+                  href="#"
```
vprothais: It seems that we can get the current url with `Astro.url`. Could we use this with a JS function to define the real href here instead of the empty hash. Would be better for accessibility as the link wouldn't be considered already visited. And we would not need the javascript part on this component anymore.

site/src/components/header/Navigation.astro
```diff
@@ -104,6 +104,49 @@ const { addedIn, layout, title } = Astro.props
...
+                  class:list={["dropdown-item d-flex align-items-center", {active: getConfig().brand === 'orange'}]}
+                  data-bs-brand-value="orange"
+                  aria-current={getConfig().brand === 'orange'}
+                >
+                  <img class="bi me-sm" src={getVersionedDocsPath('assets/brand/orange-logo.svg')} alt="Toggle brand" loading="lazy" />
```
vprothais: Is this alt useful for the user ? Since we already have a label, I think we could consider this as decorative only.

site/src/components/header/Navigation.astro
```diff
@@ -104,6 +104,49 @@ const { addedIn, layout, title } = Astro.props
...
+                </a>
+              </li>
+              <li>
+                <a
+                  href="#"
+                  class:list={["dropdown-item d-flex align-items-center", {active: getConfig().brand === 'sosh'}]}
+                  data-bs-brand-value="sosh"
+                  aria-current={getConfig().brand === 'sosh'}
+                >
+                  <img class="bi me-sm" src={getVersionedDocsPath('assets/brand/sosh-logo.svg')} alt="Toggle brand" loading="lazy" />
```
vprothais: Same question here for the alt attribute than for Orange.

site/src/libs/sass-variables.ts
```diff
@@ -0,0 +1,21 @@
+import { exporter } from 'sass-export'
+import { getConfig } from './config'
+
+const options = {
+  inputFiles: [`packages/${getConfig().brand}/scss/tokens/_raw.scss`, `packages/${getConfig().brand}/scss/tokens/_semantic.scss`/*, `scss/_variables.scss`*/],
```
vprothais: Do we need to keep variables file commented in this list ?

site/src/libs/sass-variables.ts
```diff
@@ -0,0 +1,21 @@
...
+  includePaths: []
+}
+
+export const allTokens = exporter(options).getArray()
+
+export const getMode = (regex: string) => {
```
vprothais, line 16: Maybe we could add a little JSdoc on this one to explain what is it for, not completely obvious with the name or with the content of the function. Since it is expected with only `modes-on` tokens we could add this as a prefix in this function and remove it from every calls. Seems more logical to me to have this functions already searching in the good tokens list rather than needing the full name.

scss/mixins/_color-mode.scss
louismaximepiton: Yes totally, for instance we use a class so I set up at this specificity, but if we have an id for example, the root selector alone will apply light mode everytime when the mode is applied at the root level

scss/_reboot.scss
louismaximepiton: Yeah it's probably an old method that I used to change the & inside a selector, I'm updating it !

package.json
louismaximepiton: tbh, not sure how it handles the options, so I kept it like this, trying

louismaximepiton: Yeah doesn't work, and I don't want to set infinite command in here for something that is launched once per release

site/src/components/header/Navigation.astro
louismaximepiton: We need the Js part for keeping the target of the url, nonetheless, I like using Astro.url here.

package.json
vprothais: The commited command use npm-run-all, so does it works like that or should we go back to the previous one ?

PR #3284, opened 2025-12-17
feat: add max-width modifier for textual inputs

### LINKED ISSUE #3258: [OUDS] Max-width of component and text references
### Prerequisites

### Proposal

Components with max-width tokens available should not have a max-width by default because they can't be undefined afterwards by Figma users of the component. 

So all max-width will be removed on components in Figma, and should also be removed in OUDS Web.
If a max-width token exists for the component, we should provide a class for this component to apply the max-width on it. The token will be named with a convention (probably an emoticon) to emphasize the fact that it is not used in the Figma component and should be used as documentation and for Web.
Each component having this kind of token should have is own max-width helper and a paragraph explaining this point.

All typographies references should have a max-width defined when used, even in component. This means that when a label is used in a component as a typography reference, we should define a max-width on it, using the existing token.

Be careful, on some component we defined the typography on the global container, but the max-width should be applied only on the text. For example, each level of the breadcrumb should have this max-width, not the breadcrumb as a component.

This last point probably means that we should not define max-width directly in get-font-size for label references. Or having a parameter to disable/enable that when needed.

### Motivation and context

This max-width management will be documented in Zero Height for designers. The idea is to have a way for developer to apply that easily.

MaxLardenois:
### Types of change

- [ ] Non-breaking change

### Related issues

Closes #3258

### Checklists

- [ ] The changes are well displayed in [Storybook](https://deploy-preview-3284--boosted.netlify.app/storybook) (be careful if example order has changed for DSM)

- [ ] Code review
- [ ] Design review
- [ ] A11y review

### Live previews

- <https://deploy-preview-3284--boosted.netlify.app/>

### DIFF
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -34,7 +34,7 @@
     },
     {
       "path": "./packages/orange/dist/css/ouds-web.css",
-      "maxSize": "62.75 kB"
+      "maxSize": "63.0 kB"
     },
     {
       "path": "./packages/orange/dist/css/ouds-web.min.css",


scss/forms/_select-input.scss
```diff
@@ -3,8 +3,11 @@
 .select-input {
   @extend %text-item-common;
   --#{$prefix}text-input-min-width: #{px-to-rem($ouds-select-input-size-min-width)};
-  --#{$prefix}text-input-max-width: #{px-to-rem($ouds-text-input-size-max-width)};
   --#{$prefix}text-input-min-height: #{px-to-rem($ouds-text-input-size-min-height)};
+
+  &.max-width {
```
vprothais: I'm wondering if the class name should contains the component name. Like that it seems to be some kind of universal max-width value but in fact it depends on the component it is used, and doesn't do anything outside of those components (pretty much like the already existing active class in fact).

site/src/content/docs/forms/select-input.mdx
```diff
@@ -200,6 +200,27 @@ If a helper link is used in conjunction with a helper text, the `aria-describedb
     </a>
   </div>`} />
 
+## Layout
+
+### Max width
+
+By default select inputs will span the whole width of their parent container, to limit the width of the select input on wider parent container, add a `.max-width` to the `.text-input` container.
```
vprothais: We should add a link to the responsiveness section of the DSM when it is ready. I don't know if it will be kept with new generated content and therefore be already added.

scss/forms/_select-input.scss
MaxLardenois: Yes we had the same question with @louismaximepiton and I opted for his position.

site/src/content/docs/forms/select-input.mdx
MaxLardenois: Yes, for now we have nothing for text inputs so I did not add it here. Also we need a shortened version and that means asking Benoit for it.

scss/forms/_select-input.scss
MaxLardenois: Changed to .component-max-width

louismaximepiton (commented):
Should we have a following PR for other components ? Do we add an example for Password input as well ?

Issue #3045, opened 2025-07-01
[OUDS] Create a component page template for the doc


hannahiss:
### Prerequisites

- [x] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed feature requests
- [x] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)

### Proposal

It could look like this: 

# Approach 
# Standard variant(s)
# Variants on colored background
# Layout 
# Other variants : With icon, with chevron, etc
# Sizes : sm, md, ...
# States : disabled, visited, loading, etc

Of course, we should keep only the relevant pargraphs for each component

Then other specific paragraphs could come after. 

### Motivation and context

To ease the writting of a new component page, we could have a mdx hidden file which would contain a template/pattern for the doc.

hannahiss:
Original menu for Button:

## Base class
## Standard variants
### Rounded
## Variants on colored background
## With icon
### Text and icon
### Icon only
## Disable text wrapping
## Button tags
## Outline buttons
## Disabled state
## Loading state
## Block buttons

Would become:

## Overview
(Including standard variants and button tags)
## Layout
### Block buttons
### Disable text wrapping
## Variants
### On colored background
### Rounded
### With icon
#### Text and icon
#### Icon only
### Outline (see if we keep it)
## States
### Disabled
### Loading

hannahiss:
Original menu for Badge:
```
## Base class
## Standard variants
## Sizes
## With count
## With icons
### Positioned
```
Would become:
```
## Overview
Includes Base class and standard variants with colors
### Variants
## With count
## With icons (is 's' needed ?)
### Positioned
## Sizes
```

hannahiss:
Original menu for Breadcrumb:
```
## Basic example
## Behavior on smaller viewports
```

Proposition:
```
## Overview
## Layout
```

hannahiss:
[Deploy preview](https://deploy-preview-3062--boosted.netlify.app/docs/0.5/components/tag/)

Original menu for Tag:
```
## Tag
### Color variants
### Squared
### With asset
#### Bullet
#### Icon
#### Loader
### Sizes
## Tag input
### Disabled
```
Proposition:
```
## Overview
## Tag
### Overview
[...]

PR #3288, opened 2025-12-18
[OUDS] Add a11y labels to copy/stackblitz buttons on forms components

### LINKED ISSUE #2886: Manage "Copy to clipboard" and "Edit on Stackblitz" buttons labels fro accessibility
### Prerequisites

### Proposal

- Add discriminating information to the label of the "Edit in Stackblitz" and "Copy to clipboard" buttons of examples, to know exactly what they correspond to, like proposed in the PR https://github.com/twbs/bootstrap/pull/40820
- Use example shortcode for basic examples instead of back-quotes, to be able to pass a discriminating label for the "Copy to clipboard" button, like proposed in the PR https://github.com/twbs/bootstrap/pull/40829

### Motivation and context

All examples have some "Copy to clipboard" and "Edit on Stackblitz" buttons with the same label, which is not accessible, all over the documentation.

hannahiss:
### Related issues

Partially fixes #2886 

### Description

- Add a11y labels to copy/stackblitz buttons on all forms component except select input (already done)
- rename "read only" to readonly in checkbox/radio-button/switch for harmonization (like in Bootstrap and in text-input)
- remove useless imports

### Checklists

- [ ] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/ouds/main/.github/CONTRIBUTING.md)
- [ ] My change follows the page structure defined in #3045
- [ ] Links have been updated (title change impacts links for example)

- [ ] Code review

### Live previews

- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/text-input/>
- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/password-input/>
- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/text-area/>
- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/checkbox/>
- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/radio-button/>
- <https://deploy-preview-3288--boosted.netlify.app/docs/forms/switch/>


site/src/content/docs/forms/checkbox.mdx
```diff
@@ -403,11 +402,11 @@ Add the `disabled` attribute to the input, then the checkbox and the associated
   </div>`} />
 </BootstrapCompatibility>
 
-### Read only
+### Readonly
```
MaxLardenois: A little confused by this renaming because it comes from the design-defined API. I checked the Android repo and they use `readOnly` which suggests the space too and they use "read only" in the documentation.
Also for text input at least it's also "read only" on Figma... We used "Invalid" instead of "Error" for the state because it is the HTML API so it makes sense. In this case I think we can keep Read only even if the HTML API for inputs is `readonly` (there is no camel-case in html so it makes sense)
So my point of view is not to change that at this point

hannahiss: I'm not against "read only" instead of "readonly" but we should have the same in all components. I chose "readonly" because of Bootstrap.
So if we choose "read only", we should update text input instead.

MaxLardenois: Agreed

vprothais: I just discovered that there is a `:read-only` pseudo class : https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:read-only
There is a dash in the pseudo class, implying there should be a space between the two words.
I would stick to how it is named in Design side

PR #3332, opened 2026-02-02
fix(a11y): better a11y on doc

louismaximepiton:
### Related issues

NA

### Description

Add some unique roles that might be an issue in the future to better convey information to ATs.

### Checklists

- [x] My change follows the page structure defined in #3045
- [x] Links have been updated (title change impacts links for example)

- [ ] Code review

### Live previews

- <https://deploy-preview-3332--boosted.netlify.app/>

### DIFF
--- a/site/src/assets/examples/download-app/index.astro
+++ b/site/src/assets/examples/download-app/index.astro
@@ -35,7 +35,7 @@ export const aliases = [
   </symbol>
 </svg>
 
-<header class="sticky-top" data-bs-theme="dark" role="banner">
+<header class="sticky-top" data-bs-theme="dark">
   <nav class="navbar navbar-expand-lg" aria-label="Global navigation of download app example page">
     <div class="container-fluid container-max-width">
       <div class="navbar-brand me-auto">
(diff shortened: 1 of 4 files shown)


site/src/assets/examples/download-app/index.astro
```diff
@@ -35,7 +35,7 @@ export const aliases = [
   </symbol>
 </svg>
 
-<header class="sticky-top" data-bs-theme="dark">
+<header class="sticky-top" data-bs-theme="dark" role="banner">
```
MaxLardenois: `<header>` element has a banner role by default (except if it is inside another semantic element, but it's not the case here)

site/src/assets/examples/download-app/index.astro
```diff
@@ -851,7 +851,7 @@ export const aliases = [
   </div>
 </main>
 
-<footer class="footer navbar" data-bs-theme="dark">
+<footer class="footer navbar" data-bs-theme="dark" role="contentinfo">
```
MaxLardenois: Same question as banner on header, footer has a contentinfo role by default

site/src/assets/examples/download-app/index.astro
louismaximepiton: It's mainly in prevention, for people using old ATs/browsers couples and don't support this by default (which happens a lot on the few figures I have, but they may be wrong by the time (https://access42.net/uploads/2017/06/presentation_etude-le-bv_05122017v2.pdf) (2020 figures haven't been published yet and I don't know why 🤷)).

It's only proposals, but we can remove the PRs since these changes should be only for the old couples.

vprothais: The document is very old, more than 8 years old, not sure if it is still relevant. I would probably remove it, if it is already the built-in role.

PR #3377, opened 2026-02-19
feat(docs): add an inline search field for component in list doc page

MaxLardenois:
### Description

Add a search input for components in the components page, just for fun

### Checklists

- [ ] My change follows the page structure defined in #3045
- [ ] Links have been updated (title change impacts links for example)

- [ ] Code review

### Live previews

- <https://deploy-preview-3377--boosted.netlify.app/orange/docs/components/>


site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,50 @@
+---
+
+---
+
+<ouds-component-finder>
+  <div class="text-input mb-medium">
+    <div class="text-input-container">
+      <svg aria-hidden="true">
+        <use xlink:href="/orange/docs/1.0/assets/img/ouds-web-sprite.svg#search"></use>
```
louismaximepiton: It will update the icon for Sosh

```suggestion
        <use xlink:href={getVersionedDocsPath('assets/img/ouds-web-sprite.svg#search')}></use>
```

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,50 @@
+---
+
+---
+
+<ouds-component-finder>
+  <div class="text-input mb-medium">
```
louismaximepiton: I think we can use the gridgap here to have the same spacing between cards and this input.
```suggestion
  <div class="text-input mb-gridgap">
```

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,50 @@
...
+    <div class="text-input-container">
+      <svg aria-hidden="true">
+        <use xlink:href="/orange/docs/1.0/assets/img/ouds-web-sprite.svg#search"></use>
+      </svg>
+      <label for="oudsComponentFinderInput">Find component</label>
+      <input
+        type="search"
+        class="text-input-field"
+        id="oudsComponentFinderInput"
+        pattern="[a-zA-Z]+"
```
louismaximepiton: We should probably allow space in here (even if often we stop before) and probably special char like '-', or maybe we can remove the pattern.

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,50 @@
...
+
+---
+
+<ouds-component-finder>
+  <div class="text-input mb-medium">
+    <div class="text-input-container">
+      <svg aria-hidden="true">
+        <use xlink:href="/orange/docs/1.0/assets/img/ouds-web-sprite.svg#search"></use>
+      </svg>
+      <label for="oudsComponentFinderInput">Find component</label>
```
louismaximepiton: Probably add:
```suggestion
      <label for="oudsComponentFinderInput">Find component <span class="visually-hidden">and update the component list below</span></label>
```
Or maybe add an aria-live section around the components. Anyway it needs to go through an a11y review.

site/src/components/shortcodes/ComponentFinder.astro
louismaximepiton: Should we encapsulate all of this inside a `<form>`? Or is it useless in this case ?

louismaximepiton (changes requested):
I could spot some potential a11y issues, to be seen with our experts.

site/src/components/shortcodes/ComponentFinder.astro
MaxLardenois: ~~No we cannot add space or - because we use it in a querySelector afterward~~, I feel alphabetical only is quite safe and do the trick mostly. The pattern permits to avoir potential nasty tricks.

EDIT: Actually, if you're not a dummy like me, you can add space and - if you quote the queryselector (and have a double escape on the - in the pattern). My bad.

site/src/components/shortcodes/ComponentFinder.astro
MaxLardenois: Ah yes, no we need it for the validation (will use the js from form-validation).

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,54 @@
...
+            this.querySelectorAll(
+              `ul[data-component-list] > :not([data-name*="${searchInput.value.toLowerCase()}"])`
+            ).forEach((elt) => elt.classList.add('d-none'))
+          }
+        }, 0)
+      })
+    }
+  }
+
+  customElements.define('ouds-component-finder', OudsComponentFinder)
```
louismaximepiton, line 98: I'm wondering if we need the web component here, the js should work just fine even without it right ?

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,54 @@
...
+  </ul>
+</ouds-component-finder>
+
+<script>
+  class OudsComponentFinder extends HTMLElement {
+    connectedCallback() {
+      const searchInput = document.getElementById('oudsComponentFinderInput') as HTMLInputElement
+      if (!searchInput) return
+      searchInput.addEventListener('input', () => {
+        setTimeout(() => {
```
louismaximepiton: Is that for performance purpose ?

site/src/components/shortcodes/ComponentFinder.astro
```diff
@@ -0,0 +1,54 @@
...
+          this.querySelectorAll('ul[data-component-list] > *').forEach((elt) => elt.classList.remove('d-none'))
```
louismaximepiton: Should this be in the if below ? to keep the already existing list instead of giving the whole list again ?

louismaximepiton (commented):
Few comments that came to my mind yesterday night

site/src/components/shortcodes/ComponentFinder.astro
MaxLardenois: No, I add an issue with the value of the input that was not up to date but it seems to be good now, will remove

site/src/components/shortcodes/ComponentFinder.astro
MaxLardenois: We need to reset the list when the input is emptied

site/src/components/shortcodes/ComponentFinder.astro
MaxLardenois: We do not per se, but it permits to constrain the JS to here. Also a thing proposed by Astro

Issue #3360, opened 2026-02-13
[OUDS] Add action color helper or usage documentation


vprothais:
Add a new Color page in Foundation with the structure of https://unified-design-system.orange.com/472794e18/p/217ac6-color/b/11c27d, link to applicable existing page such as Color palette, Utilities > Background, Utilities > Color, Color modes...

Reorganize color palette page with a structure to indicate where to use each color (#3374)

Add a page (or paragraph) for action color custom props usage

PR #3453, opened 2026-03-17
feat(docs): reorganize color documentation

### LINKED ISSUE #3374: [OUDS] Enhance the foundation/color-palette page to give more context on color usage
This page shows all the colors, including decorative, functional and raw colors.
This page should be updated to give more context to each of the colors and token and in which context they can be used. For example decorative colors should only be used in illustrations.

MaxLardenois:
### Related issues

Closes #3374 , #3360

### Description

Reorganize color documentation
- replaces Foundation/Color Palette with Foundation/Colors (list and doc for all semantic colors)
- rewrites Utilities/Background 
- removes Utilities/Color & background (redirect to Background)
- removes Utilities/Colored links 
- small touches on Foundation/Color modes

### Checklists

- [x] My change follows the page structure defined in #3045
- [x] Links have been updated (title change impacts links for example)

- [ ] Code review

### Live previews

- <https://deploy-preview-3453--boosted.netlify.app/orange/docs/1.1/foundation/colors/>
- <https://deploy-preview-3453--boosted.netlify.app/orange/docs/1.1/utilities/color/>
- <https://deploy-preview-3453--boosted.netlify.app/orange/docs/1.1/utilities/background/>

### DIFF
--- a/site/src/content/docs/foundation/colors.mdx
+++ b/site/src/content/docs/foundation/colors.mdx
@@ -212,7 +212,7 @@ These colors can be used with [our opacity utilities]([[docsref:/utilities/opaci
   Please refer to the [opacity colors design reference](https://unified-design-system.orange.com/472794e18/p/217ac6-color/t/b4eb059054) for more information on usage.
 </Callout>
 
-export const opacityBackgroundStyle = "repeating-conic-gradient(#fff 0% 25%, #000 0% 50%) 0 0/ 20px 20px";
+export const opacityBackgroundStyle = "repeating-conic-gradient(var(--bs-color-bg-primary) 0% 25%, var(--bs-color-bg-inverse-high) 0% 50%) 0 0/ 20px 20px";
 
 <ColorTable
   colorPrefix="ouds/opacity/"


site/data/sidebar-utilities.yml
```diff
@@ -6,8 +6,6 @@
     - title: Aspect ratio
     - title: Background
     - title: Border
-    - title: Color & background
-    - title: Colored link
```
louismaximepiton, line 10: We don't keep the pages ? Even for the control that they are well implmented on our side ?

site/data/sidebar-foundation.yml
```diff
@@ -6,7 +6,7 @@
     - title: Options
     - title: Tokens
       draft: true
-    - title: Color palette
+    - title: Colors
```
louismaximepiton, line 9: Thinking of this, should we introduce spacing, grid, ... pages on the same architecture.

site/src/content/callouts/warning-color-assistive-technologies.md
```diff
@@ -1 +1 @@
-**Accessibility tip:** Using color to add meaning only provides a visual indication, which will not be conveyed to users of assistive technologies like screen readers. Please ensure the meaning is obvious from the content itself (e.g., the visible text with a [_sufficient_ color contrast](/docs/[[config:docs_version]]/getting-started/accessibility/#color-contrast)) or is included through alternative means, such as additional text hidden with the `.visually-hidden` class.
+Relying solely on color to convey meaning provides only a visual cue, which may not be accessible to users of assistive technologies like screen readers. Make sure the meaning is clear from the content itself, such as visible text with [_sufficient_ color contrast](/docs/[[config:docs_version]]/getting-started/accessibility/#color-contrast), or include alternative indicators, such as additional text hidden with the `.visually-hidden` class.
```
louismaximepiton:
```suggestion
Relying solely on color to convey meaning provides only a visual cue, which is not accessible to users of assistive technologies like screen readers. Make sure the meaning is clear from the content itself, such as visible text with [_sufficient_ color contrast](/docs/[[config:docs_version]]/getting-started/accessibility/#color-contrast), or include alternative indicators, such as additional text hidden with the `.visually-hidden` class.
```

site/src/content/docs/components/icon.mdx
```diff
@@ -99,7 +99,7 @@ For SVGs (SVG sprites, inline SVGs or CSS mask image SVGs), the `fill="currentCo
 
 The following color utilities and CSS variables are intended for use only with functional icons. The warning icon is an exception as it has two different colors for its internal and external shapes, to ensure accessibility. For this icon, you can use the dedicated [accessible warning icon](#accessible-warning-icon).
 
-Here are [the normal contexts of use of these functional colors](https://r.orange.fr/r/S-ouds-doc-token-color-functional).
+Here are [the normal contexts of use of these functional colors](https://unified-design-system.orange.com/472794e18/p/217ac6-color/t/6d158c2daf).
```
louismaximepiton, line 102: Isn't it better since this link should always be up to date

```suggestion
Here are [the normal contexts of use of these functional colors](https://r.orange.fr/r/S-ouds-doc-token-color).
```

site/src/content/docs/getting-started/sass.mdx
```diff
@@ -245,7 +245,7 @@ $theme-colors: map-remove($theme-colors, "info", "light", "dark");
 
 ### Colors
 
-Next to the [Sass maps]([[docsref:/foundation/color-palette#sass-maps]]) we have, theme colors can also be used as standalone variables, like `$primary`.
+Next to the [Sass maps]([[docsref:/foundation/colors#sass-maps]]) we have, theme colors can also be used as standalone variables, like `$primary`.
```
louismaximepiton:
```suggestion
Next to the [Sass maps]([[docsref:/foundation/colors]]) we have, theme colors can also be used as standalone variables, like `$primary`.
```

Or remove the sentence maybe ?

site/src/content/docs/utilities/border.mdx
```diff
@@ -46,7 +46,7 @@ Or remove borders:
 
 ## Color
 
-Change the border color using utilities. The color utilities are generated from our `$ouds-border-colors` Sass map.
+Border color utilities apply a color [token]([[docsref:/foundation/tokens]]) as `border-color` for a [border]([[docsref:/foundation/colors#border]]) or ['always' color]([[docsref:/foundation/colors#always]]) dynamically depending on the [color mode]([[docsref:/foundation/color-modes]]).
```
louismaximepiton, line 49: Could almost be a shortcode

site/src/content/docs/utilities/color.mdx
```diff
@@ -10,44 +10,38 @@ toc: true
 import { getData } from '@libs/data'
 import { allTokens } from '@libs/sass-variables'
 
-<Callout type="info" name="warning-color-assistive-technologies" />
+<Callout title="Accessibility tip" type="info" name="warning-color-assistive-technologies" />
```
louismaximepiton: I'm wondering if we can set some information in the frontmatter of the shortcode

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,204 @@
...
+This page lists all the colors available in OUDS Web and their usage, organized by their category. All colors are displayed in both light and dark modes for reference. However, they should be applied using their respective utility classes or custom properties to ensure dynamic adaptation to the [color mode]([[docsref:/foundation/color-modes]]).
+
+## Background
+
+These colors are used for whole pages and structurally meaningful sections backgrounds.
+
+These colors can be used with [our background utilities]([[docsref:/utilities/background]]).
+
+<Callout type="info" title="Usage reference">
+  Please refer to the [background colors design reference](https://unified-design-system.orange.com/472794e18/p/217ac6-color/b/11c27d) for more information on usage.
```
louismaximepiton:
```suggestion
  Please refer to the [background colors design reference](https://r.orange.fr/r/S-ouds-doc-token-color/b/11c27d) for more information on usage.
```
In the whole page

site/src/content/docs/foundation/colors.mdx
louismaximepiton, line 1: Just saying for posterity but the colors that we are showing aren't always the good ones. Maybe consider having a dark bg on the dark mode column and a white bg on the light mode column to always display the right colors.

site/src/content/docs/utilities/background.mdx
louismaximepiton, line 1: Wondering if somehow we should display the bg colors the smae way as we display the colors in the colors page

site/src/content/docs/utilities/background.mdx
```diff
@@ -283,7 +311,7 @@ In OUDS Web, we have some backgrounds that are a bit specific because they can t
 
 ## Multiple backgrounds
 
-Sometimes it will be necessary to layer two background colors on top of each other. This can happen when designers opt to have two fill color on a container for example if the parent container have a specific background and we want to opacify a card on top of it. To achieve this with OUDS Web you will need two containers with the specified background color utility.
+Occasionally, you may need to layer two background colors on top of each other. This scenario arises when design requirements call for dual fill colors on containers—for example, when a parent container has a specific background and you need to place a card over it. To implement this in OUDS Web, you'll need to create two nested containers, each with its own specified background color utility class.
 
 <Example code={`<div class="p-large" style="background: repeating-linear-gradient(-45deg, #ccc, #ccc 40px, var(--bs-color-bg-secondary) 40px, var(--bs-color-bg-secondary) 80px)">
```
louismaximepiton: Just wondering why we have hard coded values in here

site/src/content/docs/utilities/color.mdx
```diff
@@ -10,44 +10,38 @@ toc: true
...
 
 ## Colors
 
-Colorize text with color utilities. To colorize icons, please refer to our [icon color section](#on-icons).
+Content color utility classes, in the form  `.text-*`, apply a color [token]([[docsref:/foundation/tokens]]) as `color` for a [content]([[docsref:/foundation/colors#background]]) or ['always' color]([[docsref:/foundation/colors#always]]) dynamically depending on the [color mode]([[docsref:/foundation/color-modes]]).
```
louismaximepiton:
```suggestion
Content color utility classes, in the form  `.text-*`, apply a color [token]([[docsref:/foundation/tokens]]) as `color` for a [content]([[docsref:/foundation/colors#content]]) or ['always' color]([[docsref:/foundation/colors#always]]) dynamically depending on the [color mode]([[docsref:/foundation/color-modes]]).
```

site/src/content/docs/utilities/color.mdx
```diff
@@ -10,44 +10,38 @@ toc: true
...
+
+To colorize icons, please refer to our [icon color section](#on-icons).
```
louismaximepiton:
```suggestion
To colorize icons, please refer to our [icon color section]([[docsref:/components/icon#icons-color]]).
```

site/src/content/docs/utilities/color.mdx
```diff
@@ -10,44 +10,38 @@ toc: true
...
 
 OUDS Web provides many `.text-*` and `.bg-*` utilities, but they should be used with care to meet our design specifications and [WCAG 2.2 accessibility standards for color contrast](https://www.w3.org/TR/WCAG/#distinguishable).
-To be sure to respect the specifications, it is necessary to define `color`, `background-color` and `font-size` altogether.
+To ensure compliance with the standard, define `color`, `background-color`, and `font-size` together.
 
 <BrandSpecific brand="orange">
```
louismaximepiton: This is also true for Orange Compact, it might be the case on several of them on the doc.

site/src/content/docs/getting-started/migration.mdx
louismaximepiton, line 1: Not quite sure that we should change the previous changes, but it means that we have broken links 🤔 Does it mean that we should change all the text that are deprecated ? I can see a `.text-on-status-emphasized-alt` which no longer exist.

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+  lightValue?: string
+  darkValue?: string
+  backdropLight?: string
+  backdropDark?: string
+}
+
+interface Props {
+  tokenRegexp: RegExp
+  colorPrefix: string
+  getUsage?: (tokenName: string) => string
```
louismaximepiton: The types aren't right, or maybe it's the `getCustomProp()` that is not well defined

site/src/content/docs/utilities/color.mdx
louismaximepiton:
```suggestion
title: Color
```

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+    ...(backdropColorName ? getLightDarkValuesFor(backdropColorName) : {})
+  }
+}
+---
+
+<div class="table-responsive mb-small">
+  <table class="table">
+    <thead>
+      <tr>
+        <th class="w-25">{colorPrefix}</th>
```
louismaximepiton: Why w-25 here ? and not w-50 just as the other cells ?
```suggestion
        <th class="w-50">{colorPrefix}</th>
```

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+              <td class="w-50">
+                <var>{token.colorName}</var>
+              </td>
+              <td>
+                <figure
+                  class="ratio-21x9 mb-none d-flex justify-content-center align-items-center"
+                  style={{
+                    backgroundColor: token.backdropLight ? token.backdropLight : token.lightValue
+                  }}
+                >
```
louismaximepiton, line 97: Wondering if we should add a `[title={token.lightValue}]` in here so everyone can access the same information, since the figcaption is hidden

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+                  {token.backdropLight && (
+                    <svg
+                      class="decorative-small-icon"
+                      x="15"
+                      y="15"
+                      width="20"
+                      height="20"
```
louismaximepiton: Do we need x, y, width and height in here ?

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+      </tr>
+    </thead>
+    <tbody>
+      {
+        allTokens
+          .filter((token) => tokenRegexp.test(token.name))
+          .map(matchAndGetValues)
+          .map((token: IColorToken) => (
+            <tr class="align-middle">
+              <td class="w-50">
```
louismaximepiton: Should it be a `th[scope=row]`? If yes, probably add `[scope=column]` in the `<thead>`

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+                    >
+                      <use xlink:href={getVersionedDocsPath('/assets/img/ouds-web-sprite.svg#accessibility-vision')} />
+                    </svg>
+                  )}
+                  <figcaption class="visually-hidden">
+                    <var>{token.darkValue}</var>
+                  </figcaption>
+                </figure>
+              </td>
+              <td class="w-25 user-select-all" style="max-width: 2rem;">
```
louismaximepiton:
```suggestion
              <td class="w-25 user-select-all">
```

louismaximepiton: I'm wondering if the text-wrap or white-space policy can be updated for this kind of cell. For instance remove the `:first-child` in `_content.scss` l.40.

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,204 @@
...
+These colors can be used with [our border utilities]([[docsref:/utilities/border#color]]).
+
+<Callout type="info" title="Usage reference">
+  Please refer to the [border colors design reference](https://unified-design-system.orange.com/472794e18/p/217ac6-color/b/11c27d) for more information on usage.
+</Callout>
+
+<ColorTable
+  colorPrefix="ouds/color/border/"
+  tokenRegexp={/\$ouds-color-border((?!-status|-on))-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.border-${colorName}`} />
```
louismaximepiton, line 150: In this table, `.border-focus-inset` and `.border-focus` don't exist, maybe split to have a dedicated table with var(--...)

site/src/content/docs/foundation/colors.mdx
louismaximepiton, line 1: It's missing the opacity category in here, maybe to add in the design page first ?

louismaximepiton, line 1: Not sure that we can do anything to it but the rendering on mobile looks broken (misaligned) compared to the one on desktop:

Image

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
+---
+import { getVersionedDocsPath } from '@libs/path'
+import { getTokenValue, type IDeclaration } from '@libs/sass-variables'
+import { allTokens } from '@libs/sass-variables'
+
+interface IColorToken {
+  colorName: string
+  lightValue?: string
+  darkValue?: string
```
louismaximepiton:
```suggestion
  lightValue: string
  darkValue: string
```
I think they are not optional ?

site/src/content/docs/utilities/color.mdx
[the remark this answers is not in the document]
MaxLardenois: What do you think human reviewers? I think it's not very useful

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+  - "/docs/customize/color-palette/"
+  - "/docs/foundation/color-palette/"
+  - "/docs/foundation/colors/"
+toc: true
+---
+
+import { getData } from '@libs/data'
+import { getConfig } from '@libs/config'
+import { allTokens, getTokenValue } from '@libs/sass-variables'
+import { getVersionedDocsPath } from '@libs/path'
```
louismaximepiton: None of the import is needed here

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+This page lists all the colors available in OUDS Web and their usage, organized by their category. All colors are displayed in both light and dark modes for reference. However, they should be applied using their respective utility classes or custom properties to ensure dynamic adaptation to the [color mode]([[docsref:/foundation/color-modes]]).
```
louismaximepiton:
```suggestion
This page lists all the colors available in OUDS Web and their usage, organized by their category. All colors are displayed in both light and dark modes for reference. However, they should be applied following the semantic by using their respective utility classes or custom properties to ensure dynamic adaptation to the [color mode]([[docsref:/foundation/color-modes]]).
```
Not sure of the wording but I'd like to make the semantic word appear;

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+<ColorTable
+  colorPrefix="ouds/color/bg/"
+  tokenRegexp={/\$ouds-color-bg-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.bg-${colorName}`} />
+
+## Surface
+
+These colors should be used as background for elements that are on another background.
+
+These colors can be used with [our background utilities]([[docsref:/utilities/background]]). Some of these surface colors have an opacity value and are designed to come on top of another color, we explain [how to achieve that here]([[docsref:/utilities/background#multiple-backgrounds]]).
```
louismaximepiton:
```suggestion
These colors can be used with [our background utilities]([[docsref:/utilities/background]]). These [surface colors are designed to come on top of another color]([[docsref:/utilities/background#multiple-backgrounds]]).
```
The link label isn't explcit enough to me

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+<Callout type="info" title="Usage reference">
+  Please refer to the [surface colors design reference](https://unified-design-system.orange.com/472794e18/p/217ac6-color/b/11c27d) for more information on usage.
+</Callout>
+
+<ColorTable
+  colorPrefix="ouds/color/surface/"
+  tokenRegexp={/\$ouds-color-surface-((?!status|brand))(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.bg-surface-${colorName}`} />
+
+<ColorTable
```
louismaximepiton, line 43: I don't understand why we separate brand here ? Maybe we should add title if we keep it like that ?

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+  colorPrefix="ouds/color/surface/brand/"
+  tokenRegexp={/\$ouds-color-surface-brand-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.bg-surface-brand-${colorName}`} />
+
+### Status
+
+These surface colors are used specifically to indicate a status.
```
louismaximepiton, line 50: Should we add the callout that specifies that the color isn't suffiicient enough to convey information ?

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+<Callout type="info" title="Usage reference">
+  Please refer to the [content colors design reference](https://unified-design-system.orange.com/472794e18/p/217ac6-color/b/11c27d) for more information on usage.
+</Callout>
+
+<ColorTable
+  colorPrefix="ouds/color/content/"
+  tokenRegexp={/\$ouds-color-content((?!-on|-status))-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.text-${colorName}`} />
+
+<ColorTable
```
louismaximepiton, line 78: No title for Status here ?

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+  These colors should only be used for custom components development, this is why no utilities are provided, please use directly the custom properties in your CSS.
+</Callout>
+
+<ColorTable
+  colorPrefix="ouds/color/overlay/"
+  tokenRegexp={/\$ouds-color-overlay-(?<colorName>.*)-light/} />
+
+## Always
+
+These colors are the sole invariant colors defined by OUDS Web, they will stay the same on every brand, modes or contexts.
```
louismaximepiton: They are not sole since decorative are invariant, but since we don't show them, I don't know what to do

site/src/content/docs/foundation/colors.mdx
louismaximepiton, line 1: I don't know if it is relevant but maybe we can always have the equivalent CSS var and Scss vars next to the classes ?

louismaximepiton, line 1: It looks a bit like the [color theme page](https://boosted.orange.com/docs/5.3/customize/color-theme/) we had back in the days

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,203 @@
...
+  ['positive', 'info', 'warning', 'negative', 'accent'].map((status) => (
+    <ColorTable
+      colorPrefix={`ouds/color/content/on/status/${status}/`}
+      tokenRegexp={new RegExp(`\\$ouds-color-content-on-\(\?<backdropColorName>status-${status}-\(\?<colorName>\.\*\)\)-light`)}
+      getUsage={(colorName) => `.text-on-status-${status}-${colorName}`} />
+  ))
+}
+
+<Callout type="info">
+  The following content colors are highly specific and should only be used with direction from a designer.
```
louismaximepiton:
```suggestion
  The following content colors are highly specific and should only be used with direction from a designer or within the context of a custom component.
```

site/src/content/docs/utilities/background.mdx
louismaximepiton: This import seems useless

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -9,51 +9,61 @@ toc: true
...
+
+<Callout title="A note on naming" type="info">
+  We talk about *color modes* instead of *themes* as this could refer to other concepts. The attribute is named `data-bs-theme` for Bootstrap compatibility reasons but it might change in future versions.
 </Callout>
 
-## Modes
+## `data-bs-theme` attribute
 
-Modes are a way to change the **color of the text and the components inside**. We provide 4 different modes that are `"light"`, `"dark"`, `"root"`, and `"root-inverted"`. The `"light"` and `"dark"` modes are static, while the `"root"` and `"root-inverted"` modes are dynamic.
+Use the `data-bs-theme` attribute to set the color mode on a container. It can be applied on the `<html>` element to support the global [color mode toggle](#javascript). By default, our CSS behaves as if `data-bs-theme="light"` were set on `<html>`, even when the attribute is not present.
```
louismaximepiton: It's missing the fact that only light and dark can be set on html

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -9,51 +9,61 @@ toc: true
...
+Dark mode will use light gray text and dark-mode components. To apply the dark mode independently from the cascade, you can set <span class="user-select-all">`data-bs-theme="dark"`</span> on any element.
 
-### Dynamic color modes
+## Dynamic color modes
 
-We have two more dynamic themes that are `"root"` and `"root-inverted"`. These themes are named dynamic because they depend on the main element mode, so by changing the main mode, you’re changing all those dynamic calls as well. The `"root"` theme will reset the color mode to the main theme of your page, while the `"root-inverted"` theme will reset the color mode to the opposite theme of your page.
+We offer two dynamic modes: `"root"` and `"root-inverted"`. These modes are called dynamic because they depend on the root element mode, so by changing the main mode, you’re changing all the dynamic areas using these modes as well. The `"root"` theme will reset the color mode to the root mode of your page, while the `"root-inverted"` mode will set the color mode to the opposite theme of your page.
```
louismaximepiton:
```suggestion
We offer two dynamic modes: `"root"` and `"root-inverted"`. These modes are called dynamic because they depend on the root element mode **only**, so by changing the main mode, you’re changing all the dynamic areas using these modes as well. The `"root"` theme will reset the color mode to the root mode of your page, while the `"root-inverted"` mode will set the color mode to the opposite theme of your page.
```

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -9,51 +9,61 @@ toc: true
...
 
 <Callout type="info">
-  Note that these themes **aren’t meant to be used on the root element**.
+  By definition these modes **should not be used on the root element**.
```
louismaximepiton:
```suggestion
  By definition these modes **can not be used on the root element**.
```

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -121,60 +114,68 @@ However, these 4 themes should be enough to deal with all of your use cases. If
...
-      <li><a class="dropdown-item active" href="#">Action</a></li>
-      <li><a class="dropdown-item" href="#">Action</a></li>
-      <li><a class="dropdown-item" href="#">Another action</a></li>
-      <li><a class="dropdown-item" href="#">Something else here</a></li>
-      <li><hr class="dropdown-divider"></li>
-      <li><a class="dropdown-item" href="#">Separated link</a></li>
-    </ul>
+<Example class="d-flex justify-content-between flex-wrap gap-2xsmall" code={`<div class="bg-always-white">
+    <div class="dropdown" data-bs-theme="light">
+      <button class="btn btn-dropdown dropdown-toggle" type="button" id="dropdownMenuButtonLight" data-bs-toggle="dropdown" aria-expanded="false">
```
louismaximepiton:
```suggestion
      <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenuButtonLight" data-bs-toggle="dropdown" aria-expanded="false">
```

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -121,60 +114,68 @@ However, these 4 themes should be enough to deal with all of your use cases. If
...
-    </button>
-    <ul class="dropdown-menu" aria-labelledby="dropdownMenuButtonLight">
-      <li><a class="dropdown-item active" href="#">Action</a></li>
-      <li><a class="dropdown-item" href="#">Action</a></li>
-      <li><a class="dropdown-item" href="#">Another action</a></li>
-      <li><a class="dropdown-item" href="#">Something else here</a></li>
-      <li><hr class="dropdown-divider"></li>
-      <li><a class="dropdown-item" href="#">Separated link</a></li>
-    </ul>
+<Example class="d-flex justify-content-between flex-wrap gap-2xsmall" code={`<div class="bg-always-white">
```
louismaximepiton, line 117: Not perfect but already way better

site/src/content/docs/utilities/background.mdx
```diff
@@ -10,27 +10,49 @@ import { getConfig } from '@libs/config'
...
-
-Utilities that set the text color: `.bg-surface-brand-primary`, `.bg-surface-brand-secondary`, `.bg-surface-brand-tertiary`, `.bg-surface-status-accent-emphasized`, `.bg-surface-status-accent-muted`, `.bg-surface-status-positive-emphasized`, `.bg-surface-status-positive-muted`, `.bg-surface-status-negative-emphasized`, `.bg-surface-status-negative-muted`, `.bg-surface-status-warning-emphasized`, `.bg-surface-status-warning-muted`, `.bg-surface-status-info-emphasized`, `.bg-surface-status-info-muted`, `.bg-always-black`, `.bg-always-white`.
+<Callout title="Accessibility tip" type="info" name="warning-color-assistive-technologies" />
+
+## Overview
+
+Background utility classes, in the form  `.bg-*`, apply a color [token]([[docsref:/foundation/tokens]]) as `background-color` for a [background]([[docsref:/foundation/colors#background]]), [surface]([[docsref:/foundation/colors#surface]]) or ['always' color]([[docsref:/foundation/colors#always]]) dynamically depending on the [color mode]([[docsref:/foundation/color-modes]]).
```
louismaximepiton: Or opacity token

site/src/content/docs/utilities/background.mdx
```diff
@@ -10,27 +10,49 @@ import { getConfig } from '@libs/config'
...
+
+*Depending on the context of the element with a `.bg-*` class and its resulting color, it may be necessary to add a single child element with a `[data-bs-theme]` [attribute to specify the color mode]([[docsref:/foundation/color-modes#how-to-use]]) for its descendants.*
+This ensures that all descendant nodes adopt the correct color mode for the specific background color. As a result:
+* Text nodes will automatically display in the appropriate [content color]([[docsref:/foundation/colors#content-on-other-colors]]).
+* Components will respond to the color mode and adapt accordingly.
+
+*Note:* It is not possible to add the attribute directly on the `.bg-*` element, as the background color would then adapt to the attribute instead of conforming to the inherited color mode set by its ancestors.
```
louismaximepiton: Not sure if we should keep it, or maybe in a `<Callout>` ?

site/src/content/docs/utilities/background.mdx
```diff
@@ -10,27 +10,49 @@ import { getConfig } from '@libs/config'
...
+
+*For some background colors, we have set the default content color directly* using a CSS `color` attribute. This allows descendants text nodes to automatically display in the correct color. However this approach is only suitable when the `.bg-*` element contains only text, if the element contains components, you must add a child container with the `[data-bs-theme]` attribute.
```
louismaximepiton:
```suggestion
*For some background colors, we have set the default content color directly* using a CSS `color` attribute. This allows descendants text nodes to automatically display in the correct color. However this approach is only suitable when the `.bg-*` element contains only **non-colorized** text, if the element contains components, you must add a child container with the `[data-bs-theme]` attribute.
```

site/src/content/docs/utilities/background.mdx
```diff
@@ -10,27 +10,49 @@ import { getConfig } from '@libs/config'
...
+  * `.bg-surface-status-positive-emphasized`
+  * `.bg-surface-status-positive-muted`
+  * `.bg-surface-status-negative-emphasized`
+  * `.bg-surface-status-negative-muted`
+  * `.bg-surface-status-warning-emphasized`
+  * `.bg-surface-status-warning-muted`
+  * `.bg-surface-status-info-emphasized`
+  * `.bg-surface-status-info-muted`
+  * `.bg-always-black`
+  * `.bg-always-white`
```
louismaximepiton, line 47: Thinking about this it's always hard to maintain and think to maintain this kind of list, maybe we should remove ?

site/src/content/docs/utilities/background.mdx
```diff
@@ -10,27 +10,49 @@ import { getConfig } from '@libs/config'
...
-* `[data-bs-theme]` [color mode attribute]([[docsref:/foundation/color-modes#how-to-use]]) is the best option most of the time and moreover when the element using a background utility contains complex sub-elements such as components that need to respond to color modes. It also sets the right color for texts inside
-* `.text-*` [color utilities]([[docsref:/utilities/color]]) when the background color and color couple are accessible together in light and dark mode, and there is no components inside
+In the following examples we always use a child element with `[data-bs-theme]` attribute as it is the recommended way to use the `.bg-*` classes and will work in every case even if we only have text in our elements.
 
 <Callout type="info">
-  You can find more information on the OUDS background colors in the [design guidelines](https://unified-design-system.orange.com/472794e18/p/217ac6-color/t/b3d025a22f)
+  You can find more information on the OUDS background and surface colors in the [design guidelines](https://unified-design-system.orange.com/472794e18/p/217ac6-color/t/b3d025a22f)
```
louismaximepiton, line 53: Can we use a redirect on this ?

site/src/content/docs/utilities/background.mdx
```diff
@@ -88,14 +110,20 @@ Here are the nominal cases you should be using, when there is no other theme int
...
-#### Inside a static theme
+## Inside a static color mode
 
-Here is a more complex example to understand how to use `[data-bs-theme]` in specific static contexts. In here you should only use the inverse static theme by replacing all `data-bs-theme="root-inverted"` by `data-bs-theme="light"`. So if you’re in a dark theme context, use only `data-bs-theme="light"` when and if needed and respectively.
+The following examples demonstrate the proper use of `[data-bs-theme]` in static color mode contexts. In this case, background colors will stay the same so it's safe to use a corresponding `[data-bs-theme]` to set the mode for the descendant nodes.
+
+For example in a `light` color mode context, dark backgrounds like `.bg-inverse-high` will stay dark so we must add a `data-bs-theme="dark"` child element to set the descendant in the right mode.
+
+<Callout type="info">
+  As you can see in the examples, it is not necessary to add the `[data-bs-theme]` child element when the color mode context already matches the background color. For example `.bg-primary` is a light background and therefore the parent `data-bs-theme="light"` already sets its descendant in the right mode.
```
louismaximepiton: Not sure what you meant here, the `.bg-primary` always adapt to the context, maybe the `.bg-brand-primary` as an example here ?

site/src/content/docs/utilities/background.mdx
```diff
@@ -131,7 +159,7 @@ Here is a more complex example to understand how to use `[data-bs-theme]` in spe
   `  </div>
   </div>`]} />
 
-We also provide some opacity backgrounds that should only be used in very specific context, to opacify the background and make some text readable. Prefer using `.bg-secondary` when possible.
+We also provide opacity-based background classes, which should be used only in specific contexts where you need to adjust background opacity to enhance text readability. In most cases, `.bg-secondary` is the preferred alternative.
```
louismaximepiton: Or `.bg-surface-secondary` to have opacity

site/data/sidebar-utilities.yml
MaxLardenois: The color & background content and examples are still present in the Background page (at the end), I kept the redirection from /color-background.
Colored link do not generate anything on our side even with bootstrap compat (we do not have colored-links.scss) so there is nothing to check.

However your question is good and maybe we should have a hidden page to test all bootstrap compat stuff that we do not want in the doc?

site/data/sidebar-foundation.yml
MaxLardenois: Maybe? It's a little easier to understand because there is no spacing categories but let's discuss this. #3501

site/src/content/callouts/warning-color-assistive-technologies.md
MaxLardenois: Tried something else, less convoluted.

site/src/content/docs/components/icon.mdx
MaxLardenois: It's not the same page, the one you suggest goes to "Overview" and we want "Usage > Content". We could ask for a direct link but I fear the modifications Maxime want to do in this page would change this... That's why I kept it as that, meaning to circle back to it when the page is modified and dry on the DSM.

site/src/content/docs/getting-started/sass.mdx
MaxLardenois: I think we said we need to rewrite this page too. For the moment I added a deprecated danger callout redirecting users to the colors page, wdyt?

site/src/content/docs/utilities/border.mdx
MaxLardenois: I feel we would not gain much clarity as the color list is also variable.

site/src/content/docs/utilities/color.mdx
MaxLardenois: Did that, now we can add a title: Something in a callout frontmatter and it will be used as the Callout title

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Not the same page. But I updated all links to the right paragraph

site/src/content/docs/utilities/background.mdx
MaxLardenois: Decided to not do it right now to prioritize other things, will see if users ask for more clarity. #3489

site/src/content/docs/getting-started/migration.mdx
MaxLardenois: I would say only when there is a dead link? The deprecation is stated in the version where it is removed.

site/src/components/shortcodes/ColorTable.astro
MaxLardenois: Yes ok, the information would be there but people would not have an easy way to copy

site/src/components/shortcodes/ColorTable.astro
MaxLardenois: Reckless copy-paste

site/src/components/shortcodes/ColorTable.astro
MaxLardenois: I tried but it breaks the alignment of all colors

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Yes, the problem comes from the namespace and I do not see how to prevent this

site/src/components/shortcodes/ColorTable.astro
MaxLardenois: Yes, older versions needed that but we can make it mandatory now

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Tried something. Not all surface have an opacity value and the "multiple background" tip is not necessary in all cases so I would prefer to keep the general phrasing. 
Added something on the link.

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Two reasons: if we don't we would display ouds/color/surface/brand-primary instead of ouds/color/surface/brand/primary. Also this is the way it is displayed on Figma (https://www.figma.com/design/Co2t6wHMf4GB9NJVGs2Hes/-OUDS-Lib--Design-tokens?node-id=14339-583)

site/src/content/docs/foundation/colors.mdx
MaxLardenois: As we refer to the Boackground utilties page for usage I do not think it is necessary

site/src/content/docs/foundation/colors.mdx
MaxLardenois: I tried to keep it in line with Figma (https://www.figma.com/design/Co2t6wHMf4GB9NJVGs2Hes/-OUDS-Lib--Design-tokens?node-id=14339-583)

site/src/content/docs/foundation/colors.mdx
MaxLardenois: tried something

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Actually I don't know where this comes from, I will try to stick to the Figma explanation like the others.

site/src/content/docs/foundation/colors.mdx
MaxLardenois: I think it would muddy the message (use the utilities when available) and make the page very busy but if we think it's necessary, we can.

site/src/content/docs/foundation/color-modes.mdx
MaxLardenois: I have an issue now with the btn-default, can you check if you have an idea...

site/src/content/docs/utilities/background.mdx
MaxLardenois: Without it there is no way to know which utility sets the color (apart from inspecting)...

site/src/content/docs/utilities/background.mdx
MaxLardenois: A redirect? There is no shortlink for now

site/src/content/docs/utilities/background.mdx
MaxLardenois: I tried "For example `.bg-primary` is a lighter background (in a light mode context) and therefore the parent `[data-bs-theme="light"]` already sets its descendant in the right mode."

On the contrary, you need the data-bs-theme on bg-surface-brand-primary because it's always a "lighter" background color.

Maybe I miss something but we need to explain why sometimes we have the attribute, sometimes not.

site/src/content/docs/utilities/background.mdx
MaxLardenois: Used bg-inverse-high instead

site/src/content/docs/foundation/colors.mdx
MaxLardenois: I added a Callout to explain

site/src/content/docs/utilities/color.mdx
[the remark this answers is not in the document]
louismaximepiton: It isn't very useful indeed, but it might be great if there were links to web.unified-design-system.orange.com/docs/utilities/colored-link, It might be great to keep it somewhere to avoid broken links

site/data/sidebar-utilities.yml
louismaximepiton: I like the hidden page idea.

site/src/content/docs/getting-started/sass.mdx
```diff
@@ -245,7 +245,11 @@ $theme-colors: map-remove($theme-colors, "info", "light", "dark");
 
 ### Colors
 
-Next to the [Sass maps]([[docsref:/foundation/color-palette#sass-maps]]) we have, theme colors can also be used as standalone variables, like `$primary`.
+<Callout title="Deprecated" type="danger">
+  Please read [the colors documentation]([[docsref:/foundation/colors]]) for information on colors
```
louismaximepiton:
```suggestion
  Please read [the colors documentation]([[docsref:/foundation/colors]]) for more information on colors.
```

site/src/content/docs/getting-started/migration.mdx
louismaximepiton: I think that aliases should be enough not to have dead links, but maybe your method should be better ?

site/src/components/shortcodes/ColorTable.astro
louismaximepiton: I'm fine with it, it's only for people to have access to the same information, thanks for the modfs :)

site/src/components/shortcodes/ColorTable.astro
louismaximepiton: Ok, it's a shame, but I see 😕

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,227 @@
...
+  tokenRegexp={/\$ouds-color-border((?!-status|-on|-focus))-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.border-${colorName}`} />
+
+<ColorTable
+  colorPrefix="ouds/color/border/status/"
+  tokenRegexp={/\$ouds-color-border-status-(?<colorName>.*)-light/}
+  getUsage={(colorName) => `.border-status-${colorName}`} />
+
+<Callout type="info">
+  The following border colors do not have utilities because they are used only in a Sass mixin to create focus rings
```
louismaximepiton:
```suggestion
  The following border colors do not have utilities because they are used only in a Sass mixin to create focus rings.
```

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,227 @@
...
+</Callout>
+
+<ColorTable
+  colorPrefix="ouds/opacity/"
+  tokenRegexp={/\$ouds-opacity-(?<colorName>.*)/}
+  getUsage={(colorName) => `.opacity-${colorName}`}
+  getStyle={(token, mode) =>
+    mode === 'light'
+      ? {opacity: token.lightValue, backgroundColor: 'var(--bs-color-surface-brand-primary)'}
+      : {opacity: token.darkValue, backgroundColor: 'var(--bs-color-surface-brand-primary)'}} />
```
louismaximepiton:
```suggestion
<ColorTable
  colorPrefix="ouds/color/opacity/"
  tokenRegexp={/\$ouds-color-opacity-(?<colorName>.*)/}
  getUsage={(colorName) => `.bg-opacity-${colorName}`} />
```

Not sure if the actual opacity tokens should or should not be in this page tbh.
Not sure if the suggestion opacity is background only or if it can be applied on borders and content as well, we might modify our doc accordingly

site/src/content/docs/foundation/colors.mdx
louismaximepiton: Me neither tbh 😕

site/src/content/docs/foundation/colors.mdx
louismaximepiton: For me every surface even if not opacity is meant to be palced on a background. The wording you proposed is better

site/src/content/docs/foundation/color-modes.mdx
louismaximepiton: It's maybe far from ideal but it feels fine to me, thanks

site/src/content/docs/utilities/background.mdx
```diff
@@ -3,34 +3,59 @@ title: Background
...
+This ensures that all descendant nodes adopt the correct color mode for the specific background color. As a result:
+* Text nodes will automatically display in the appropriate [content color]([[docsref:/foundation/colors#content-on-other-colors]]).
+* Components will respond to the color mode and adapt accordingly.
 
-Utilities that set the text color: `.bg-surface-brand-primary`, `.bg-surface-brand-secondary`, `.bg-surface-brand-tertiary`, `.bg-surface-status-accent-emphasized`, `.bg-surface-status-accent-muted`, `.bg-surface-status-positive-emphasized`, `.bg-surface-status-positive-muted`, `.bg-surface-status-negative-emphasized`, `.bg-surface-status-negative-muted`, `.bg-surface-status-warning-emphasized`, `.bg-surface-status-warning-muted`, `.bg-surface-status-info-emphasized`, `.bg-surface-status-info-muted`, `.bg-always-black`, `.bg-always-white`.
+<Callout type="info">
+  It is not possible to add the `[data-bs-theme]` attribute directly on the `.bg-*` element, as the background color would then adapt to the attribute instead of conforming to the inherited color mode set by its ancestors.
```
louismaximepiton, line 26: Actually it might be useful in some rare cases, but I don't know if we should promote that., let's keep it as-is.

site/src/content/docs/utilities/background.mdx
louismaximepiton: Yes, but maybe they are the only ones that need this behavior and this might be transparent for user, idk, let's keep it like that for now

site/src/content/docs/utilities/background.mdx
louismaximepiton: No sorry, I think I was kind of tired. The actual you proposed seems good 👍

site/src/components/shortcodes/ColorTable.astro
```diff
@@ -0,0 +1,132 @@
...
+        allTokens
+          .filter((token) => tokenRegexp.test(token.name))
+          .map(matchAndGetValues)
+          .map((token: IColorToken) => (
+            <tr class="align-middle">
+              <th class="w-50" scope="row">
+                <var>{token.colorName}</var>
+              </th>
+              <td>
+                <div class="bg-primary ratio-21x9" data-bs-theme="light">
```
louismaximepiton: Love it, thanks !

site/src/content/docs/foundation/colors.mdx
```diff
@@ -186,18 +189,39 @@ These colors are used for overlay elements like dropdown, modals and tooltips to
...
+
 ## Decorative
 
-<p>No tokens generated</p>
+<Callout type="info" title="Coming soon">
```
louismaximepiton, line 235: Are we sure that we'll implement them soon ? Or maybe never ?

site/src/content/docs/utilities/background.mdx
```diff
@@ -142,6 +171,8 @@ We also provide some opacity backgrounds that should only be used in very specif
 
 </details>
 
+With a `dark` color mode context, light backgrounds like `.bg-inverse-high` will stay light so we must add a `[data-bs-theme=="light"]` child element for descendants to set the descendant in the right mode.
```
louismaximepiton:
```suggestion
With a `dark` color mode context, light backgrounds like `.bg-inverse-high` will stay light so we must add a `[data-bs-theme="light"]` child element for descendants to set the descendant in the right mode.
```

site/src/content/docs/utilities/background.mdx
```diff
@@ -187,9 +216,9 @@ We also provide some opacity backgrounds that should only be used in very specif
 
 </details>
 
-#### Inside a dynamic theme
+## Inside a dynamic color mode
 
-Here is a more complex example to understand how to use `[data-bs-theme]` in specific dynamic contexts. In here you should only replace `data-bs-theme="root-inverted"` by `data-bs-theme="root"` and respectively.
+The following examples demonstrate the proper use of `[data-bs-theme]` in dynamic contexts. In this case you must take into account that some background colors will change when the color mode context change. For example in a (root) `light` color mode, a parent container with `[data-bs-theme=="root-inverted"]` will make the background color of a container with `.bg-inverse-high` light, therefore we must switch back the local color mode to `root` for its descendants.
```
louismaximepiton:
```suggestion
The following examples demonstrate the proper use of `[data-bs-theme]` in dynamic contexts. In this case you must take into account that some background colors will change when the color mode context change. For example in a (root) `light` color mode, a parent container with `[data-bs-theme="root-inverted"]` will make the background color of a container with `.bg-inverse-high` light, therefore we must switch back the local color mode to `root` for its descendants.
```

site/data/sidebar-utilities.yml
MaxLardenois: I will add an issue #3500

site/src/content/docs/getting-started/migration.mdx
MaxLardenois: I think a link to "Read more in the Colored links page" linking to another page is a little strange, not very understandable from a user POV (especially LLMs)
Redirections can be helpful to avoid dead links from the web but I'm not sure it's ok inside our documentation.

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Yes they are, but they will not always be used together on Figma like the "Multiple backgrounds" paragraph tries to help to implement.

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Yes, I don't know tbh wdyt @vprothais?
The Opacity tokens list on Figma is not in the Colors table so you have a point.
In any case we could move this paragraph elsewhere (and rename the ColorTable component)

site/src/content/docs/foundation/colors.mdx
MaxLardenois: Idk, but if we think not or in a very long time we should explain why. What is the reason we might not implement them or only in the far future? @vprothais ?

site/src/content/docs/utilities/color.mdx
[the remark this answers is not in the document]
MaxLardenois: Added an alias on components/links

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+These surface colors are used specifically to indicate a status.
+
+Like other surface colors, they can be used with [our background utilities]([[docsref:/utilities/background]]).
+
+{
+  ['positive', 'info', 'warning', 'negative', 'accent'].map((status) => (
+    <ColorTable
+      colorPrefix={`ouds/color/surface/status/${status}/`}
+      tokenRegexp={new RegExp(`\\$ouds-color-surface-status-${status}-\(\?<colorName>\.\*\)-light`)}
```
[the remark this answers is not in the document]
MaxLardenois: Works fine as is, would not work w/o escaping.
Also RegExp.escape does not seem to work

site/src/content/docs/utilities/background.mdx
```diff
@@ -3,34 +3,59 @@ title: Background
 description: Convey meaning through `background-color`.
 aliases:
   - "/docs/utilities/background/"
+  - "/docs/utilities/color-background/"
+  - "/docs/[[config:docs_version]]/color-background/"
```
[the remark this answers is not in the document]
MaxLardenois: Ok but once again I do not think there is inbound links to this page atm. This adds several pages for no real reason imo
But we can if humans want it @vprothais

site/src/components/shortcodes/Callout.astro
```diff
@@ -76,7 +65,21 @@ switch (type) {
     break
 }
```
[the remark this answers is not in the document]
MaxLardenois: Only found one occurence in the docs, changed it even if it is a custom component example

site/src/content/docs/foundation/colors.mdx
vprothais: Hum, good question. I was about to propose to use a checkerboard pattern as a background to avoid interpreting that this helpers provide some kind of orange declination but maybe we should don't have it here at all ? Since they are defined in color family, I think we can keep them here.
Here is an example of background found on stackoverflow (old school), maybre we can find a better one :
```css
    background-position: 0px 0px, 10px 10px;
    background-size: 20px 20px;
    background-image: linear-gradient(45deg, #eee 25%, transparent 25%, transparent 75%, #eee 75%, #eee 100%),linear-gradient(45deg, #eee 25%, white 25%, white 75%, #eee 75%, #eee 100%);
```
WDYT ?

site/src/content/docs/utilities/color.mdx
```diff
@@ -14,40 +14,34 @@ import { allTokens } from '@libs/sass-variables'
...
+<Callout type="warning">
+  Please note that aside from `.text-brand-primary`, `.text-default`, and `.text-muted`, most text-utility classes should be avoided as they are highly contextual. Instead, apply our [color modes]([[docsref:/foundation/color-modes]]) directly—as described in the [background utilities]([[docsref:/utilities/background]])—to get the right text color as well as ensure nested components render in the correct mode.
 </Callout>
 
 Here are some compliant combinations examples for texts:
 
 <Example code={[...allTokens.filter(token => token.name.match(/\$ouds-color-content-/) && !token.name.match(/-action-/) && token.name.match(/-light/) && token.compiledValue).map(token => {
     if (token.name.match(/-on-/)) {
-      return `<p class="bg-surface-${token.name.split('-').slice(4, -1).join('-')} text-${token.name.split('-').slice(3, -1).join('-')} p-small">Text ${token.name.split('-').slice(3, -1).join('-')}</p>`
+      return `<p class="bg-surface-${token.name.split('-').slice(4, -1).join('-')} text-${token.name.split('-').slice(3, -1).join('-')} p-small fs-hl">Text ${token.name.split('-').slice(3, -1).join('-')}</p>`
```
vprothais, line 37: did you increase the size regarding the accessibility requirement with brand-primary font size on white background ?

site/src/content/docs/utilities/background.mdx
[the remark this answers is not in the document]
vprothais: The files are added and almost never removed, so when we remove an alias from the list, it will not generate the page anymore, but the old page will still exists in the website, redirecting to an old version of the documentation (the version in which the documentation was generated). Saying that, the question is, will anyone arrive on this page, I would say no since links inside the documentation always use the real url with the version. It could only impact external url such as the one in RBO, but we can upgrade them. I think we can clean the alias list, or it will grow infinitely.

site/src/content/docs/utilities/background.mdx
vprothais: I think sharing the list is a good point. If we say that some helpers will also define the color, users will want to know which helpers are setting it or not. The point is still relevant, there is a really big risk to forget to update that list. I don't have any idea on how to handle that automatically but exporting scss maps to objects and we will have some issue doing that with the dependency not working so well in every cases.
Let's keep it like that for now.

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -9,51 +9,61 @@ toc: true
...
+Dark mode will use light gray text and dark-mode components. To apply the dark mode independently from the cascade, you can set <span class="user-select-all">`[data-bs-theme="dark"]`</span> on any element.
 
-### Dynamic color modes
+## Dynamic color modes
 
-We have two more dynamic themes that are `"root"` and `"root-inverted"`. These themes are named dynamic because they depend on the main element mode, so by changing the main mode, you’re changing all those dynamic calls as well. The `"root"` theme will reset the color mode to the main theme of your page, while the `"root-inverted"` theme will reset the color mode to the opposite theme of your page.
+We offer two dynamic modes: `"root"` and `"root-inverted"`. These modes are called dynamic because they depend **only** on the root element mode, so by changing the main mode, you’re changing all the dynamic areas using these modes as well. The `"root"` theme will reset the color mode to the root mode of your page, while the `"root-inverted"` mode will set the color mode to the opposite theme of your page.
```
vprothais:
```suggestion
We offer two dynamic modes: `"root"` and `"root-inverted"`. These modes are called dynamic because they depend **only** on the root element mode, so by changing the main mode, you’re changing all the dynamic areas using these modes as well. The `"root"` theme will reset the color mode to the root mode of your page, while the `"root-inverted"` mode will set the color mode to the opposite mode of your page.
```

site/src/content/docs/foundation/color-modes.mdx
```diff
@@ -67,41 +77,24 @@ We have two more dynamic themes that are `"root"` and `"root-inverted"`. These t
...
 
-#### `root-inverted` mode
-
-Here is an example of a block that inverts the main theme when it’s placed inside a dark block. To get the inverted main theme independently from the cascade, set via <span class="user-select-all">`data-bs-theme="root-inverted"`</span> on any element but the main one. Feel free to change the color mode of the page to see the effect.
-
-#### `root` mode
+### `root-inverted` mode
 
-Here is an example of a block that follows the main theme when it’s placed inside a dark block. To get the main theme independently from the cascade, set via <span class="user-select-all">`data-bs-theme="root"`</span> on any element but the main one. Feel free to change the color mode of the page to see the effect.
+This color mode inverts the main color mode. It will be useful for example inside a container with a darker background in root light mode like `.bg-inverse-high`. To apply this mode add <span class="user-select-all">`[data-bs-theme="root-inverted"]`</span> on any element but the main one.
```
vprothais: In this sentence, it feels like the "like `.bg-inverse-high`" goes with the "root light mode". Maybe we should move this part just after "darker background".

```suggestion
This color mode inverts the main color mode. It will be useful for example inside a container with a darker background, like `.bg-inverse-high`, in root light mode. To apply this mode add <span class="user-select-all">`[data-bs-theme="root-inverted"]`</span> on any element but the main one.
```

site/src/content/docs/foundation/color-modes.mdx
MaxLardenois: Tried to get rid of more of them

site/src/content/docs/foundation/color-modes.mdx
MaxLardenois: I omitted the second comma which I think make it less readable, tell me if it's ok

site/src/content/docs/utilities/color.mdx
MaxLardenois: Ah yes I did. Especially because we say "Here are some compliant combinations examples for texts:" after talking about the size.

site/src/content/docs/foundation/colors.mdx
vprothais: I have a weird rendering with a diagonal appearing in Firefox :

Image

In the stackoverflow there is a more recent way to do that : https://stackoverflow.com/questions/35361986/css-gradient-checkerboard-pattern

I did some test and we just need to tweak the background position a bit and it works well for us :
```css
background: repeating-conic-gradient(#fff 0% 25%, #000 0% 50%) 0 0/ 20px 20px;
```

site/src/content/docs/foundation/color-modes.mdx
vprothais: ok, good for me, I wasn't sure about it either

site/src/content/docs/utilities/color.mdx
vprothais: Yeah, it feels a bit big but we have to be consistent with what we say about font size with orange color, good idea

site/src/content/docs/foundation/colors.mdx
```diff
@@ -0,0 +1,237 @@
...
+        }
+      : {
+          opacity: token.darkValue,
+          background: opacityBackgroundStyle
+        }} />
+
+## Decorative
+
+<Callout type="info" title="Coming soon">
+  Decorative colors are not yet implemented in OUDS Web.
```
MaxLardenois, line 236:
```suggestion
<Callout type="info" title="For design only">
  Decorative colors are used for illustrations design purpose only, they should not be needed in OUDS Web. Please contact us if you need to use it in your project.
```

PR #3588, opened 2026-05-29
feat: update textarea to 1.2

### LINKED ISSUE #3534: [OUDS] [Component] Update Text area 1.2
## Release note:
- Scope: Trailing action=False / State=Loading → Replacement of the button component (loading state) with a “Trailing container” frame containing the circular progress indicator component
- The input text reference changes from a regular weight (font-weight-label-default) to a medium weight (font-weight-label-moderate).
- Removal of the component’s max-width to avoid blocking certain layouts and to provide greater flexibility of use. The max-width values remain available in the design documentation.
- Adding the resize handle boolean property
- For the “Empty” status and for all states (except the “Focus” state):
  - Update of the typographic reference for the “Label text” → From “Label-small-default” to “Label-large-default”
  - Modification of the padding block top value, with the addition of a new token → control-text-area-space-padding-block-top-empty
  - Removal of the padding block bottom to rebalance the component height across all statuses and all states.
- For the “Trailing container” frames:
  - The responsive behavior (auto-layout) of the height changes from “Fill container” to “Hug content.”
  - Removal of the max-height, token: ouds-control-text-area-size-max-height-assets-container
  - Removal of the min-height, token: ouds-action-button-size-min-height
  - Modification of the padding block, addition of 2 new tokens:
- For the “Empty” status → ouds-control-text-area-space-padding-block-empty-trailing-container
- For the “Empty (Placeholder)” and “Filled” status → ouds-control-text-area-space-padding-block-trailing-container
- Add a .fixed-height as an option to prevent autoresize
 
The Figma component is available here:
https://www.figma.com/design/5cF3f3KF2ppcj4m7LWqQTo/Control---Text-area?node-id=2708-1519

MaxLardenois:
### Types of change

- [ ] Non-breaking change

### Related issues

Closes #3534

### Checklists

- [ ] The changes are well displayed in [Storybook](https://deploy-preview-3588--boosted.netlify.app/storybook) (be careful if example order has changed for DSM)

- [ ] Code review
- [ ] Design review
- [ ] A11y review

### Live previews

- <https://deploy-preview-3588--boosted.netlify.app/orange/docs/1.3/components/text-area/>

### DIFF
--- a/scss/forms/_text-input.scss
+++ b/scss/forms/_text-input.scss
@@ -445,10 +445,10 @@
 
     .text-area-field {
       width: 100%;
-      height: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) + px-to-rem($ouds-text-area-size-min-height-input * 2) + var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
-      min-height: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) + px-to-rem($ouds-text-area-size-min-height-input) + var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
+      height: calc(px-to-rem($ouds-text-area-size-min-height-input * 2) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top) + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small));
+      min-height: calc(px-to-rem($ouds-text-area-size-min-height-input) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top) + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small));
       max-height: calc(px-to-rem($ouds-text-area-size-max-height-input) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
-      padding-top: calc($ouds-text-area-space-padding-block-top-empty + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) - var(--#{$prefix}text-input-border-width-top));
+      padding-top: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-top));
       padding-right: calc($ouds-text-input-space-column-gap-default + $ouds-button-size-min-width + $ouds-text-input-space-padding-inline-trailing-action - var(--#{$prefix}text-input-border-width-right));
       padding-bottom: calc($ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-bottom));
       padding-left: calc($ouds-text-input-space-padding-inline-default - var(--#{$prefix}text-input-border-width-left));
@@ -510,9 +510,17 @@
         content: "";
         background-color: transparent;
       }
+
+      // stylelint-disable selector-no-qualifying-type
+      &.is-required {
+        padding-right: calc(5px + $ouds-text-input-space-column-gap-label-small-asterisk);
+
+        &::after {
+          margin-left: $ouds-text-input-space-column-gap-label-small-asterisk;
+        }
+      }
     }
 
-    // stylelint-disable selector-no-qualifying-type
     &:has(label.visually-hidden) {
       .text-area-field {
         padding-top: calc($ouds-text-area-space-padding-block-top-empty - var(--#{$prefix}text-input-border-width-top));
@@ -520,37 +528,11 @@
     }
 
     &:has(.text-area-field:placeholder-shown:not([placeholder=" "]), .text-area-field[placeholder]:not(:placeholder-shown), .text-area-field:focus:not(:read-only)) {
-      .text-area-field {
-        height: calc(px-to-rem($ouds-text-area-size-min-height-input * 2) + $ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-bottom));
-        min-height: calc(px-to-rem($ouds-text-area-size-min-height-input) + $ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-bottom));
-
-        @supports (field-sizing: content) {
-          height: unset;
-          field-sizing: content;  // Not yet baseline, but supported in most browsers
-        }
-      }
-
       &:not(:has(label.visually-hidden)) {
-        .text-area-field {
-          padding-top: 0;
-          margin-top: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-top));
-        }
 
         > label {
           padding-top: calc($ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-top));
-          text-overflow: ellipsis;
-          white-space: nowrap;
           @include get-font-size("label-small");
-
-          &.is-required {
-            width: fit-content;
-            padding-right: calc(5px + $ouds-text-input-space-column-gap-label-small-asterisk);
-
-            &::after {
-              right: 0;
-              margin-left: $ouds-text-input-space-column-gap-label-small-asterisk;
-            }
-          }
         }
       }
     }
--- a/scss/forms/_text-input.scss
+++ b/scss/forms/_text-input.scss
@@ -85,7 +85,7 @@
       z-index: 2;
       overflow: hidden;
       color: var(--#{$prefix}text-input-label-color);
-      @include transition(font-size .1s ease-out);
+      @include transition(font-size .15s ease-out);
 
       // Allow the whole text-input-container to be clickable to focus the input
       &::before {
@@ -496,7 +496,7 @@
       color: var(--#{$prefix}text-input-label-color);
       text-overflow: ellipsis;
       white-space: nowrap;
-      @include transition(font-size .1s ease-out);
+      @include transition(font-size .15s ease-out);
       @include get-font-size("label-large");
 
       // Allow the whole text-area-container to be clickable to focus the input
--- a/scss/forms/_text-input.scss
+++ b/scss/forms/_text-input.scss
@@ -447,6 +447,7 @@
       width: 100%;
       height: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) + px-to-rem($ouds-text-area-size-min-height-input * 2) + var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
       min-height: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) + px-to-rem($ouds-text-area-size-min-height-input) + var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
+      max-height: calc(px-to-rem($ouds-text-area-size-max-height-input) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
       padding-top: calc($ouds-text-area-space-padding-block-top-empty + var(--#{$prefix}font-line-height-label-large) * var(--#{$prefix}font-size-label-large) - var(--#{$prefix}text-input-border-width-top));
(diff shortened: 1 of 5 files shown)


site/src/content/docs/components/text-area.mdx
```diff
@@ -112,13 +118,43 @@ To display a helper link below text areas, use a standard small link [with `.lin
...
   </div>`} />
 
+### Hidden label
+
+The label is optional from a visual perspective, but for accessibility reasons, it is mandatory to keep a `.visually-hidden` label. Hiding the label should be reserved when the purpose of the text area is obvious thanks to a placeholder or contextual icon.
+
+<Example buttonLabel="text area with hidden label" code={`<div class="text-area">
+    <div class="text-area-container">
+      <label for="textAreaWithoutLabel" class="visually-hidden">Hidden label</label>
+      <textarea id="textAreaWithoutLabel" class="text-area-field" placeholder="Please enter a comment"></textarea>
```
louismaximepiton: IMO, it's not good to show this kind of example since it can encourage people use it. Maybe add a details/summary around this example.

louismaximepiton: We should probably add a title attribute with the label content, but I let @Aniort decide how to correctly handle it.

site/src/content/docs/components/text-area.mdx
```diff
@@ -112,13 +118,43 @@ To display a helper link below text areas, use a standard small link [with `.lin
...
+      <textarea id="textAreaHelperLink" class="text-area-field" placeholder=" "></textarea>
     </div>
     <a href="#" id="feedbackTextHelpMoreLink" aria-labelledby="feedbackTextHelpMoreLink labelTextAreaHelperLink" class="link link-small">
       More information <span class="visually-hidden">on</span>
     </a>
   </div>`} />
 
+### Hidden label
+
+The label is optional from a visual perspective, but for accessibility reasons, it is mandatory to keep a `.visually-hidden` label. Hiding the label should be reserved when the purpose of the text area is obvious thanks to a placeholder or contextual icon.
```
louismaximepiton:
```suggestion
The label is optional from a visual perspective, but for accessibility reasons, it is mandatory to keep a `.visually-hidden` label. 

<Callout type="danger">
  Hiding the label should be reserved when the purpose of the text area is obvious thanks to the context
</Callout>
```

site/src/content/docs/getting-started/migration.mdx
```diff
@@ -45,15 +45,16 @@ toc: true
...
-### Utilities
+- <span class="tag tag-small tag-negative"><span class="tag-status-icon"></span>Breaking</span> Text area component has been updated to 1.2.0:
+  - the placeholder attribute is now mandatory
+  - a `.text-area-no-resize` class has been added to disable the resizing feature on the text area
+  Read more in our [text area page]([[docsref:/components/text-area]]) and in the [design changelog](https://r.orange.fr/r/S-ouds-doc-text-area_changelog).
 
-#### Shadow
+#### Text input
 
-- <span class="tag tag-small tag-warning"><span class="tag-status-icon"></span>Warning</span> `.shadow-default` has been deprecated in favor of `.shadow-elevated`. It will be removed in v1.5.0. Read more in our [shadow page]([[docsref:/utilities/shadow]]).
```
louismaximepiton, line 56: I think there was a merge conflict down here on Utilities > Shadow.

site/src/content/docs/getting-started/migration.mdx
```diff
@@ -45,15 +45,16 @@ toc: true
...
 
-#### Text input
-
-- <span class="tag tag-small tag-info"><span class="tag-status-icon"></span>Info</span> Text input component has been updated to 1.4.0 with no outward changes. Read more in the [design changelog](https://r.orange.fr/r/S-ouds-doc-text-input_changelog).
+#### Text area
 
-### Utilities
+- <span class="tag tag-small tag-negative"><span class="tag-status-icon"></span>Breaking</span> Text area component has been updated to 1.2.0:
+  - the placeholder attribute is now mandatory
+  - a `.text-area-no-resize` class has been added to disable the resizing feature on the text area
```
louismaximepiton: Previously, we add multiple tags when there are some breaking changes and some new, don't know which method is best

scss/forms/_text-input.scss
```diff
@@ -480,55 +508,51 @@
...
+        padding-top: calc($ouds-text-area-space-padding-block-top-empty - var(--#{$prefix}text-input-border-width-top));
       }
     }
 
-    .text-area-field {
-      z-index: 3;
-      width: 100%;
-      height: calc(px-to-rem($ouds-text-area-size-min-height-input) * 2 + $ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-bottom));
-      min-height: calc(px-to-rem($ouds-text-area-size-min-height-input) + $ouds-text-area-space-padding-block - var(--#{$prefix}text-input-border-width-bottom));
-      max-height: px-to-rem($ouds-text-area-size-max-height-input);
```
louismaximepiton, line 503: We don't have max-height anymore

scss/forms/_text-input.scss
```diff
@@ -460,14 +483,19 @@
...
-      padding-left: calc(#{$ouds-text-input-space-padding-inline-default} - var(--#{$prefix}text-input-border-width-left));
+      width: 100%;
+      padding-top: $ouds-text-area-space-padding-block-top-empty;
+      padding-left: calc($ouds-text-input-space-padding-inline-default - var(--#{$prefix}text-input-border-width-left));
       overflow: hidden;
       color: var(--#{$prefix}text-input-label-color);
       text-overflow: ellipsis;
       white-space: nowrap;
-      @include get-font-size("label-small");
+      @include transition(font-size .1s ease-out);
```
louismaximepiton: The transition duration on select is 0.15s, need to ask which one is good, maybe a token

scss/forms/_text-input.scss
```diff
@@ -460,14 +483,19 @@
     }
 
     > label {
+      position: absolute;
+      top: 0;
+      left: 0;
       z-index: 2;
-      padding-right: calc($ouds-text-input-space-padding-inline-trailing-action + var(--#{$prefix}text-input-trailing-action-width) + var(--#{$prefix}text-input-column-gap) - var(--#{$prefix}text-input-border-width-right));
```
louismaximepiton, line 464: I think the padding-right needs to stay as-is ?

Image

scss/forms/_text-input.scss
```diff
@@ -460,14 +483,19 @@
...
-      padding-left: calc(#{$ouds-text-input-space-padding-inline-default} - var(--#{$prefix}text-input-border-width-left));
+      width: 100%;
+      padding-top: $ouds-text-area-space-padding-block-top-empty;
```
louismaximepiton: Shouldn't we remove the border-size here ?

louismaximepiton (commented):
There are issues with the textarea:
- the textarea required field isn't visible anymore when the label is too long
  Image

- the textarea label isn't cropped anymore once focused
  Image
  maybe it should wrap now ? 
  Image

louismaximepiton (changes requested):

site/src/content/docs/components/text-area.mdx
MaxLardenois: I don't get why, this is present in the design specs... I can put the second sentence in a Callout though

MaxLardenois: Why a .visually-hidden label would not be enough?

site/src/content/docs/components/text-area.mdx
MaxLardenois: Ah yes thanks ;)

site/src/content/docs/getting-started/migration.mdx
MaxLardenois: tried something

MaxLardenois:
~Added an issue for the required marker issue as it could be a difficult pb to tackle~ This is fixed for text area but not select https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/3619

scss/forms/_text-input.scss
MaxLardenois: I will use .15s everywhere and open an issue in the dev&design synchro to ask for a token

scss/forms/_text-input.scss
```diff
@@ -439,19 +434,48 @@
...
 
+    .text-area-field {
+      width: 100%;
+      height: calc(px-to-rem($ouds-text-area-size-min-height-input * 2) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top) + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small));
+      min-height: calc(px-to-rem($ouds-text-area-size-min-height-input) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top) + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small));
+      max-height: calc(px-to-rem($ouds-text-area-size-max-height-input) + $ouds-text-area-space-padding-block * 2 - var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-bottom) - var(--#{$prefix}text-input-border-width-top));
+      padding-top: calc($ouds-text-area-space-padding-block + var(--#{$prefix}font-line-height-label-small) * var(--#{$prefix}font-size-label-small) - var(--#{$prefix}text-input-border-width-top));
```
louismaximepiton: I think it should be `margin-top` here and update the heights accordingly, otherwise, the text goes under the label

scss/forms/_text-input.scss
```diff
@@ -198,9 +195,7 @@
       }
     }
 
-    &:has(.text-input-field:placeholder-shown:not([placeholder=" "])),
-    &:has(.text-input-field[placeholder]:not(:placeholder-shown)),
-    &:has(.text-input-field:focus:not(:read-only)) {
+    &:has(.text-input-field:placeholder-shown:not([placeholder=" "]), .text-input-field[placeholder]:not(:placeholder-shown), .text-input-field:focus:not(:read-only)) {
       > label {
```
louismaximepiton, line 204: We should probably ask design to be more like us.
>the textarea label isn't cropped anymore once focused
Image
maybe it should wrap now ?
Image

louismaximepiton (commented):
2 small things and we're good to go I think

scss/forms/_text-input.scss
MaxLardenois: Put the padding-top on the container, looks like it works

scss/forms/_text-input.scss
MaxLardenois: This should be handled correctly

louismaximepiton (approved):

site/src/content/docs/components/text-area.mdx
vprothais: Sounds good to me, not sure why a title attribute should be provided but as you said let the expert talk :)

vprothais (approved):

site/src/content/docs/components/text-area.mdx
MewenLeHo: @Aniort is currently unavailable but I can give give you the opinion of an acessibility expert.

The rule is simple: **every input field must have a visible label**.

WCAG SC 3.3.2 A:

In order to be compliant, **at least one of the following conditions is met**:
- A label is present near each form element allowing them to be visually associated,
- **or the form element has a title attribute**,
- or the form element is accompanied by text that becomes visible when focus is set on it.

You must ensure one of these conditions is met for the affected input fields.

For information it as also the same in [RGAA 11.1.1](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#11.1.1)

vprothais: Thank you @MewenLeHo , does this mean that the placeholder cant be used as a label (I think of search field where we often have only a placeholder as visual label, and the "real" label is visually hidden).
We should add the title and the placeholder at the same time ?

MaxLardenois: Thank you for your input @MewenLeHo I will add the title attribute then.

MaxLardenois: @vprothais the placeholder is not visible when there already is an input in the field so it does not qualify as a label I think

MaxLardenois: @MewenLeHo do we keep the `visually-hidden` label with the `title` attribute?

MewenLeHo: **Placeholders are not reliable** because they disappear when you start typing, which can be a problem for people with cognitive disorders or short-term memory problems. That's why a placeholder alone is not sufficient to provide a visible label; it must be accompanied by persistent text near the form element.

However, there is an exception for form elements that do not have a visible label but whose role is obvious (example: a search field accompanied by a magnifying glass button, because **in this case the magnifying glass button acts as a visible label**). In such cases, this is an exception, and the criterion is considered compliant.

vprothais: > @vprothais the placeholder is not visible when there already is an input in the field so it does not qualify as a label I think

Yeah I had that in mind and it seems pretty logical. But I saw that the search fields of orange.com and orange.fr have a placeholder but no title attribute. The use case is pretty different though, a search field is almost always empty on page load, so the placeholder is probably always visible :)

vprothais: perfect, thank you @MewenLeHo !

MewenLeHo: @vprothais: you are right.
And if you have a look at their accessibility declaration statement, you will see that Success Criterion 3.3.2 is declared as non-compliant.

louismaximepiton: > @MewenLeHo do we keep the `visually-hidden` label with the `title` attribute?

Up :)

MewenLeHo: I would say 'yes', because some assistive technologies (such as older versions of screen readers) are not able to reliably read the content of the `title` attribute.

And do not forget that the accessible name (given by the `visually-hidden` label properly linked to the input) and the visible label (given by the `title` attribute) must match or else you would fail [WCAG SC 2.5.3](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html).

PR #3604, opened 2026-06-12
feat(docs): added links to github changelog to migration page headers

### LINKED ISSUE #3457: [OUDS] Add
### Prerequisites

### Proposal

Add a direct link to the Github changelog on each heading in Migration

### Motivation and context

It'd be helpful

duvalale:
### Related issues

Closes #3457 

### Description

Updated migration page headers to add link redirecting to matching changelog.

### Checklists

- [X] I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/ouds/main/.github/CONTRIBUTING.md)
- [X] My change follows the [developer guide](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/wiki/Developer-guide)
- [X] My change follows the page structure defined in #3045
- [X] Title and DOM structure is correct
- [X] Links have been updated (title change impacts links for example)

- [ ] Code review

### Live previews

- <https://deploy-preview-3604--boosted.netlify.app/>


MaxLardenois (changes requested):
I think it'd be better if the changelog link was on a dedicated link or button below the title (or on its right for wider viewports)
Here the link is not easily visible and it lacks an accessible text.

duvalale:
I updated the headers to add a dedicated link.
I had to use a remark and rehype plugins to be able to add the link and position it correctly in the DOM for keyboard navigation, otherwise, the focus order would have been incorrect (anchor-link focused after changelog link).

site/src/libs/utils.ts
```diff
@@ -47,3 +48,12 @@ export function processMarkdownToHtml(markdown: string): string {
 export function getComponentSVG(className: string): string {
   return `<svg class="${className}" width="1rem" height="1rem" aria-label=" - component type - "><use xlink:href="${getVersionedDocsPath('assets/img/ouds-web-sprite.svg#component-atom')}" /></svg>`
 }
+
+export function getVersionLink(version: string): string {
+  return `<a class="link float-end pt-none align-content-center" href="https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/releases/tag/${version}-ouds-web">Full changelog</a>`
```
MaxLardenois: I would use `p-none` here to improve alignment.

Also I think we could use a Link with icon (https://web.unified-design-system.orange.com/orange/docs/1.3/components/links/#with-icon) with the "external link" icon from Solaris (https://unified-design-system.orange.com/472794e18/p/03a2da-solaris-icon-finder) to make it clear it's another website (with a `target=_blank` maybe)

I don't know what will be the result in the TOC though...

site/src/libs/remark.ts
```diff
@@ -14,6 +14,8 @@ const configRegExp = /\[\[config:(?<name>[\w\.]+)]]/g
 const docsrefRegExp = /\[\[docsref:(?<path>[\w\.\/#-]+)]]/g
 // [[comp]]
 const compRegExp = /\[\[comp\]\]\s*/
+// [[version-link]]
+const versionLinkRegExp = /\[\[version-link\]\]\s*/
```
MaxLardenois: I'm wondering if we shouldn't use `[version-changelog-link]` to be more precise

site/src/libs/rehype.ts
```diff
@@ -32,3 +33,40 @@ export const rehypeBsTable: Plugin<[], Root> = function () {
     })
   }
 }
+
+// A rehype plugin to reorder header anchor and version links
+export const rehypeHeaderLinksOrder: Plugin<[], Root> = function () {
```
MaxLardenois, line 38: Maybe we could simplify this by using the option `before` on `rehypeAutolinkHeadings` 
This way the anchor-link would always be first (before the actual heading text), this would be that less to do. 
But it works as is, just an idea.

MaxLardenois (approved):
Nice work!
I have some suggestion but nothing blocking so I will approve

site/src/libs/rehype.ts
duvalale: I tried to keep the existing headers as is on purpose, but it is indeed simpler to move the anchor-link before the actual heading.
I removed the rehypeHeaderLinksOrder  and updated the autolink conf.

site/src/libs/remark.ts
duvalale: Done

site/src/libs/utils.ts
duvalale: I noticed that version link is not vertically aligned to the heading when using `p-none`. Only `pt-none` allows it to be centered.

I added the external link icon.

site/src/libs/remark.ts
duvalale: Updated to suggested value

site/src/libs/rehype.ts
duvalale: I kept the existing header as is on purpose, but if I am allowed to update it, it is indeed better to prepend the anchor link.
It's done, and the rehypeHeaderLinksOrder is removed.

site/src/libs/utils.ts
MaxLardenois: I feel it's better if it's somewhat bottom aligned as it's inly text but I will let others give their opinion

site/src/libs/rehype.ts
MaxLardenois: Sorry I was not clear: we should keep the end result as is (anchor-link after the heading text). But for this script I think it might be helpful if the anchor-link were always first, so that you only need to spilt the heading elements and reorder everytime, no need to find the anchor-link

site/src/libs/utils.ts
duvalale: Padding updated.

site/src/libs/rehype.ts
duvalale: Done as you suggested.
However, this does not simplify as much as I thought the function since we also have to manage the component icon displayed on some headings.

MaxLardenois:
There is two issues in the result : 
- a single `"` is present on the changelog link
- the icon is not visible

site/src/libs/utils.ts
```diff
@@ -47,3 +52,17 @@ export function processMarkdownToHtml(markdown: string): string {
 export function getComponentSVG(className: string): string {
   return `<svg class="${className}" width="1rem" height="1rem" aria-label=" - component type - "><use xlink:href="${getVersionedDocsPath('assets/img/ouds-web-sprite.svg#component-atom')}" /></svg>`
 }
+
+export function getVersionLink(version: string): string {
+    return `<a class="link icon-link float-end p-none" href="https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/releases/tag/${version}-ouds-web" target="_blank" rel="noopener">
```
louismaximepiton, line 57: Not a big fan of using float-end, but can't figure out how to achieve the smae thing in a different way.

site/src/libs/toc.ts
```diff
@@ -11,6 +12,12 @@ export function generateToc(allHeadings: MarkdownHeading[], types?: string[]) {
   const toc: TocEntry[] = []
 
   for (const heading of headings) {
+    if (heading.text.includes('Full changelog')) {
+      heading.text = heading.text.replace(' Full changelog', '')
```
louismaximepiton: I think that we don't need the condition above to run the line.

site/src/libs/utils.ts
```diff
@@ -30,6 +31,10 @@ export function getSlug(str: string) {
   return slug(str).replace(/--+/g, '-')
 }
 
+export function getHeadingSlug(str: string) {
+  return str.replace(/-------full-changelog---------------------/g, '')
```
louismaximepiton, line 35: I can't figure out why we have this specific string ?

site/src/libs/rehype.ts
```diff
@@ -32,3 +33,47 @@ export const rehypeBsTable: Plugin<[], Root> = function () {
     })
   }
 }
+
+// A rehype plugin to reorder header anchor link and header text
+export const rehypeHeaderLinksOrder: Plugin<[], Root> = function () {
```
louismaximepiton, line 38: I quite don't unserstand why we do want them in this exact order, I think title -> Full changelog -> anchor is fine no ?

louismaximepiton (approved):
I'm fine with the changes, tbh I think we could go through the frontmatter instead of a remark thing since all the title h2 are changed on the page. However, it works like a charm so I'm fine with the changes, if you wanna merge as-is.

site/src/libs/rehype.ts
duvalale: If I don't reorder them, the DOM will not reflect the reading/display order, and keyboard navigation won't match what the user sees : We will see title -> anchor -> full changelog, and keyboard navigation order will be title -> full changelog -> anchor.

site/src/libs/utils.ts
duvalale: From what I understand, rehypeHeadingIds generates slug from the content of the whole node, which includes the "full changelog" link, and generates this string.
If I don't remove it, the generated slug when clicking on anchor or TOC is `#vXYZ-------full-changelog---------------------`, which is not pretty in the URL.

site/src/libs/toc.ts
duvalale: Indeed, updated

PR #3712, opened 2026-07-22
fix(lib): rename pointer-events utilities

### LINKED ISSUE #3710: Duplicated class for padding-end and pointer-events
### Prerequisites

- [x] I have [searched](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [x] I have [validated](https://html5.validator.nu/) any HTML to avoid common problems

### Describe the issue

When adding the `pe-none` class, padding end is set to none, AND pointer event is also set to none.
It's not possible to set only padding end or pointer events to none.

### Reduced test cases

```html
<div class="col-4 pe-none">
  <a href="#">Link</a>
</div>
``` 

Windows

Firefox

### What version of OUDS Web are you using?

v1.4.0

MaxLardenois:
### Types of change

- [ ] Non-breaking change

### Related issues

Closes #3710

### Context & Motivation

See issue

### Description

Rename both pointer-events utilities pe-none and pe-auto to pointer-events-none and pointer-events-auto

### Checklists

- [ ] The changes are well displayed in [Storybook](https://deploy-preview-3712--boosted.netlify.app/storybook) (be careful if example order has changed for DSM)

- [ ] Code review
- [ ] Design review
- [ ] A11y review
- [ ] Commit on `main` following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)

### Live previews

- <https://deploy-preview-3712--boosted.netlify.app/>


packages/migrate/src/replacements/common.mjs
```diff
@@ -102,6 +102,10 @@ export const commonReplacements = [
   ['text-white-50', 'text-muted'],
   [...warnForClass('text-opacity-[0-9]+', { remove: true, link: 'https://web.unified-design-system.orange.com/docs/utilities/text' })],
 
+  // Interactions
+  ['pe-none', 'pointer-events-none'],
```
louismaximepiton, line 106: Maybe this change needs to be in the boosted file only since there might be issue changing it from ouds
Nonetheless, the pe-auto can stay here
Maybe add a warning about it somehow ?

scss/_utilities.scss
```diff
@@ -1058,7 +1058,7 @@ $utilities: map-merge(
     ),
     "pointer-events": (
       property: pointer-events,
-      class: pe,
+      class: pointer-events,
```
louismaximepiton, line 1061: Wondering if we should use a smaller class like `pev` or anything else

louismaximepiton (changes requested):

packages/migrate/src/replacements/common.mjs
MaxLardenois: common.mjs means boosted + ob1
I guess we could change pe-auto to pointer-events-auto in ouds
I do not think the warning would be a good think because for the time being we do not know if it was the pointer events or the padding one so the warning would come up indefinitely even for paddings usage

scss/_utilities.scss
MaxLardenois: I thought about it but I think it’s not necessary as we would gain a handful of bytes for a less readable thing

louismaximepiton (approved):

vprothais (approved):

PR #3749, opened 2026-08-06
chore(deps): Bump docsearch dependency to v4.7.0

louismaximepiton:
Bump de docsearch deps and update all the needed things inside to look like the previous one.


vprothais (commented):
I have a white square juste before the icon of some item in the list and I'm not sure what it should be : 

Image

vprothais:
We have an issue with the descender of some letters like "g", "i", "p", ... May be it is linked to the line height but they are hidden :

image

vprothais:
I just saw that we have an issue with the focus ring on the search input trailing elements like "clear" and "close". When navigating with the keyboard we don't have a focus state visible on those elements.

vprothais (approved):

PR #3819, opened 2026-09-15
feat: update Alert message to v1.2.0

### LINKED ISSUE #3769: Update component Alert Message to version 1.2.0
Inline Alert component has been upgraded to version 1.2.0 and should be updated in the library.

**This updates depends on Design Tokens 2.7.0**

Updated the min-height-bottom-action token value from core-ouds/dimension/1250 to core-ouds/dimension/1000
Renamed the dialog/alert/size/min-height-bottom-action-placement token to dialog/alert/size/min-height-bottom-action.
Added the ouds/dialog/alert/space/padding-block-bottom-content token to apply bottom padding to the description and bullet list.
Removed the row-gap-action token from the component
Added bottom padding to the icon Asset container using the ouds/dialog/alert/space/padding-block token, and changed its height to Fill container
Moved the border-width and row-gap-bullet tokens from the alert family to the alert-message, as they are used exclusively by the Alert message component
Renamed the ouds/dialog/alert/size/icon token to ouds/dialog/alert/size/asset, as this size is shared by both the icon and the progress indicator used in the component
Renamed the “Button” icon only to “Close button”
Added the new ouds/size/max-width/boxed-text token to define the maximum width of the text, description, and bullet list.

MaxLardenois:
### Types of change

- [ ] Non-breaking change

### Related issues

Closes #3769 

### Context & Motivation

Update to last version

### Description

See issue

### Checklists

- [x I have read the [contributing guidelines](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/main/.github/CONTRIBUTING.md)
- [x] My change pass all tests
- [x] My change is compatible with a responsive display
- [x] My change introduces changes to the documentation that I have updated accordingly
  - [x] Links have been updated (title changes impact links)

- [x] The changes need to be in the migration guide
- [ ] The changes are well displayed in [Storybook](https://deploy-preview-3819--boosted.netlify.app/storybook) (be careful if example order has changed for DSM)

- [ ] Code review
- [ ] Design review
- [ ] A11y review
- [ ] Commit on `main` following [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)

### Live previews

- <https://deploy-preview-3819--boosted.netlify.app/orange/docs/components/alerts/#alert-message>


scss/_alert.scss
```diff
@@ -43,6 +42,10 @@
     @include border-top-end-radius(var(--#{$prefix}alert-border-radius));
   }
 
+  .alert-icon {
+    margin-top: $ouds-alert-space-padding-block;
+  }
```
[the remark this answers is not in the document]
MaxLardenois: I guess the LLM inferred that the asset container should fill its container height based on the new token name. On the design side it's only the case when the text content is not wrapping so using margin-top still conforms to the design.

scss/_alert.scss
```diff
@@ -166,6 +177,10 @@
   margin-bottom: 0;
   @include get-font-size("label-large", true);
   font-weight: $ouds-font-weight-web-label-moderate;
+
+  .alert-message & {
```
louismaximepiton: I think this breaks some principles that we have, we should only use & before a complement 
This will result as .alert-message [data-bs-theme="orange"] .alert-label or something like this which won't select anything I think

site/src/content/docs/components/alerts.mdx
```diff
@@ -437,7 +437,8 @@ If the alert can be removed when its content is acknowledged, add a close button
 You can add an action [link]([[docsref:/components/links]]) (`<a>` or `<button` depending on the context of use) either after the `.alert-text-container` or inside the `.alert-action-container`.
 
 <Callout type="warning">
-  Please note that only the basic link should be used.
+  - Please note that only the basic link should be used.
```
louismaximepiton:
```suggestion
  - Please note that only the `Text only` link should be used.
```

site/src/content/docs/components/alerts.mdx
```diff
@@ -437,7 +437,8 @@ If the alert can be removed when its content is acknowledged, add a close button
...
+  - The [compact version of the link]([[docsref:/components/links#sizes]]) must be used when it is placed inside `.alert-container`.
```
louismaximepiton, line 441: Must ? I don't like to force people to be less accessible. For the project that don't have a designer, I think it's fine to have a large link.

site/src/content/docs/getting-started/migration.mdx
```diff
@@ -17,6 +17,18 @@ toc: true
...
 
+## v1.6.0 [[version-changelog-link]]
+
+<hr />
+
+### Components
+
+#### Alerts
+
+- <span class="tag tag-small tag-negative"><span class="tag-status-icon"></span>Breaking</span> Alert message action links must now use [their compact version]([[docsref:/components/links#sizes]]) when they are included at the bottom the an alert message. Read more in our [alerts page]([[docsref:/components/alerts#with-action]]).
```
louismaximepiton:
```suggestion
- <span class="tag tag-small tag-negative"><span class="tag-status-icon"></span>Breaking</span> Alert message action links must now use [their compact version]([[docsref:/components/links#sizes]]) when they are included at the bottom of an alert message. Read more in our [alerts page]([[docsref:/components/alerts#with-action]]).
```

skills/using-ouds-web/references/components/alerts.md
louismaximepiton, line 1: Did you update it manually or did you use a specific prompt to update it ?

site/src/content/docs/getting-started/migration.mdx
louismaximepiton: Do we need to add something in our migration tool ?

louismaximepiton (changes requested):

scss/_alert.scss
MaxLardenois: For some reason it works fine here, I do not have the [data-bs-theme="orange"]...

MaxLardenois: Still I moved it inside .alert-message

site/src/content/docs/components/alerts.mdx
MaxLardenois: Actually we do not have "Text only" in the link variant, tried something

site/src/content/docs/components/alerts.mdx
MaxLardenois: I would agree but in this case we would not follow the design... And we should not say that this is a breaking change if it's not mandatory...
Will keep for others opinions.

skills/using-ouds-web/references/components/alerts.md
MaxLardenois: Prompt: "Suggest an update to the using-ouds-web skill for alerts with the 1.2.0 update."

The LLM formatted it like that, also in migration.md (though I deleted a superfluous line)

site/src/content/docs/getting-started/migration.mdx
MaxLardenois: I do not think we can because we would need to detect where the link is placed in the DOM.  Even simply if it's inside an alert-message is not trivial

skills/using-ouds-web/references/components/alerts.md
louismaximepiton: Just asked to Vincent, but the Skill should be updated only on releases to avoid conflict between what can be pulled and the non-released things.

louismaximepiton (approved):

